PR #113, opened 2012-10-01
2192 edit reply title

swaldman3:
Fixed for S2 comment editing, with two new properties in case people want to change the titles.
Haven't fixed for the BML version, since there's a beta running to make that obselete :-)


kareila:
💩💩

PR #142, opened 2012-11-07
Bug4677/font sources

momijizukamori:
We may want to file a separate bug to add a small FAQ entry about this, as I'm not sure the note is clear enough.


nfagerlund:
sigh, robo, wyd.

PR #538, opened 2013-07-25
(bug 5162) We don't have friends. :(

rahaeli:
Change mention of 'friends-only' to 'locked', and while I'm at it,
strip the email subject/body. Tested that this doesn't break the
process-privacy worker, but for the life of me, cannot make the
success email get sent (either with or without these changes).
So, the actual email itself has not been tested; please test before
commit!


zorkian:
Landing untested to clear this out. ;)

Issue #759, opened 2014-06-22
Customize styles not in use


foxfirefey:
This will implement a way to customize a style without having it in use--letting people test out and adjust styles before applying them to their journal.

PR #760, opened 2014-06-22
Enables editing unused styles in the wizard.

foxfirefey:
- Adds a styleid variable to the wizard widgets, so it is not always
  assuming the current style.
- Gives an indication that the style being editing in the wizard is not the
  current style

Fixes Issue #759.


zorkian:
Hi! Do you think you will get back to this, or should we try to find a new owner?

zorkian:
(Form letter incoming.)

Issue #821, opened 2014-07-15
Should we limit what's returned from the /data/edges API?


Sophira:
I'm developing a suite of tools (for my own use only as of right now) to use the data APIs to get data. For some reason I decided it would be a good idea to try it on dw_news. I'm not entirely sure why, but there you go.

I don't have verbose output yet for exactly what stages took the most time, but as a whole, the fetching and processing took about a minute. I looked in my database afterwards to see how many edges had been entered:

```
mysql> select type, count(*) from account_edges where from_id = 5 or to_id = 5 group by type;
+--------+----------+
| type   | count(*) |
+--------+----------+
| watch  |    50000 |
| member |        4 |
| admin  |        2 |
+--------+----------+
3 rows in set (0.13 sec)
```

Yep, that's quite a lot.

Do we want to limit the amount of data that's sent from the APIs? I don't want to run the query again without permission, but I'm guessing that the biggest bottleneck would have been the downstream bandwidth. For my server that's about 1MB/s, meaning that if the bandwidth is indeed the limiting factor, then the implications are rather worrying.

May I make this query one more time and save the data to a file locally so that I can test things without having get the data over the network again?

zorkian:
cc @momijizukamori for your API stuff as a consideration. Since this hasn't cropped up as A Problem in 4 years I'm going to close it wrt our current APIs, but the new stuff should definitely make sure we have limits in them (at least planned, not saying we need them to begin with).

Sophira:
Out of curiosity, was there something wrong with my patch?

zorkian:
What patch? This is an issue, not a PR?

alierak:
Look up in the thread, there's a referencing commit in her repo, not a PR.

zorkian:
I don't review every random commit on every random fork. If someone wants a review, they need to submit a PR.

Sophira:
Yeah, I apologise, sorry. I thought I had made a PR.

Sophira:
Oh, I had (#856), and there were changes requested. Sorry.

PR #856, opened 2014-07-22
(Issue #821) Respect edge blacklist, lower limit

Sophira:
Data obtained by /data/edges can be quite large; limit the data to 5000 edges for most edge types. Also, respect the %LJ::FORCE_EMPTY_SUBSCRIPTIONS configuration option.

This patch also fixes a potential race condition and adds functionality to the member_userids and member_of_userids functions to request a limit to the number of userids returned.

---

This patch still needs testing and I'll be doing that after I make this PR. I would appreciate eyes on this pull request, however - in particular, I'm wondering if it's clear what I'm doing with using and assigning to the various lists inside the parameters to the load_userids calls, or if I should use a more verbose but more readily understandable form. (The reason I use lists in the first place is because there's the remote possibility of a race condition otherwise, if the user list changes between the first and second calls to the *_userids subs.)

I also had to add 'limit' support to member_userids and member_of_userids as they didn't support that before now. I haven't added this support to maintainer_userids and moderator_userids; if it's felt that they should have it, I'll add it in.

### DIFF
--- a/cgi-bin/DW/Controller/Edges.pm
+++ b/cgi-bin/DW/Controller/Edges.pm
@@ -9,6 +9,7 @@
 #      foxfirefey <skittisheclipse@gmail.com>
 #      Mark Smith <mark@dreamwidth.org>
 #      Andrea Nall <anall@andreanall.com>
+#      Sophie Hamilton <sophie-dw@theblob.org>
 #
 # Copyright (c) 2009 by Dreamwidth Studios, LLC.
 #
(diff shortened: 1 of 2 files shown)


zorkian:
[slowresponseit'smyfault]

Why would you want to include them? It feels like since they're not applicable, they shouldn't be included. That way you can do things like 'if has key maintainers' and then show a maintainer section?

zorkian:
(Form letter incoming.)

PR #931, opened 2014-08-11
Idea: move the test database to be separate from a hack's DB

afuna:
We have too many tests touching the database to be able to realistically
run a test without a database -- but we could at least separate them
from actual data in the 'hack.

The other really nice thing about this is that you can drop the test
databases and start from scratch at any time -- just in case you start
getting bad results because of old data (which was happening to me; I'd
deleted a bunch of users on my dev environment but still had their
entries lying around, so tests would intermittently fail for no reason
at all...)

Not ready for deploy. In particular, there will probably need to be
tweaks before this will be usable on an actual dreamhack server since
not everyone should have the same test db credentials.

Also I'm not happy with how I figure out whether I'm in testing or not
(eeeh), but what do you all think?


cgi-bin/ljlib.pl
```diff
@@ -39,6 +39,7 @@ BEGIN
 
 BEGIN {
     # mod_perl does this early too, make sure we do as well
+    $LJ::_T_RUNNING = 1 if $0 && $0 =~ m!^t/!;
```
zorkian: Do you prefer this over having something in each test file that says "I am a test, you are in the test environment"?

afuna: I think that's fine and is less likely to accidentally break in
the future.


On Wed, Jan 28, 2015, at 11:15 PM, Mark Smith wrote:
> In cgi-bin/ljlib.pl[1]:


> > @@ -39,6 +39,7 @@ BEGIN
>
>  BEGIN {
>      # mod_perl does this early too, make sure we do as well
> +    $LJ::_T_RUNNING = 1 if $0 && $0 =~ m!^t/!; Do you prefer this
>      over having something in each test file that says "I am a test,
>      you are in the test environment"?


> — Reply to this email directly or view it on GitHub[2].

>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/931#discussion_r23750212
  2. https://github.com/dreamwidth/dw-free/pull/931/files#r23750212

afuna:
Updated!

* renamed variable from `$LJ::_T_RUNNING` to `$LJ::_T_CONFIG`
* every test that pulls in ljlib.pl first sets `$LJ::_T_RUNNING = 1`
* also make sure texttool uses the test config when initializing test db

zorkian:
I don't think you actually pushed your updated changes.

afuna:
Oops. yes okay, now I have.

zorkian:
Thanks!

PR #1139, opened 2015-01-23
Remove PubSubHubbub

zorkian:
This system has not been maintained since it was implemented and doesn't
provide any real value. Remove it.


afuna:
Bye fred!

PR #1140, opened 2015-01-24
Remove blocking_report etc

zorkian:
This code is dead to us, we don't have anythign that can receive these
reports and computers are fast enough that we don't need to microprofile
our db/memcache code.


afuna:
goodbye code, I never knew ya

PR #1141, opened 2015-01-24
Add barebones Amazon SES SMTP sender

zorkian:
This uses Net::SMTPS to send email over TLS encrypted SMTP via Amazon's
SES service. This worker is a drop-in replacement, once configured you
can just spin this one up instead of the other send-email and it will
route email through SES instead of directly to the targets.


zorkian:
Tested in the Amazon SES sandbox and it sent email, so it seems to work.

afuna:
Oh interesting!

afuna:
So few spaces! Looks good otherwise, though.

PR #1144, opened 2015-01-25
Remove some more unused code

zorkian:
None of this is used in production Dreamwidth, hasn't been for years, so
let's remove it to simplify our codebase. We can always revive things we
end up deciding we want.


cgi-bin/ljlib.pl
```diff
@@ -1004,34 +1004,32 @@ sub start_request
             if LJ::is_enabled('esn_ajax');
 
         # contextual popup JS
-        if ( $LJ::CTX_POPUP ) {
```
afuna: pst I think this wasn't meant to be part of your changes

afuna: Hmm or maybe it is? I guess you're just assuming that this is enabeld

zorkian: Yes, I meant to do this. I'm removing things that are dead weight, particularly tunables that we don't really want to support. We don't really have many sites using DW code (if any?) so we shouldn't promulgate this arcane knowledge IMO.

rahaeli:
Didn't we use the Tor exit node listing to prevent payments from Tor nodes (more likely to be fraud)?

zorkian:
Yeah, years ago. It hasn't been used since we moved to ServerBeach, it seems, so I decided to cull it. If it turns out we need it at some point, we can throw it back in really easily (git reverts are peachy).

rahaeli:
Got it! Well, we haven't seen any increase in fraud (knock wood) with it not running, so let us hope this does not change. :)

PR #1145, opened 2015-01-25
[#1134] Use protocol-relative URL for image objects in S2

### LINKED ISSUE #1134: Fix mixed content issues with URLs generated in S2
Problem: S2 caches IMGROOT, etc urls. But on http pages, these should be http://.. and on https pages these should be https.

We tried a quick fix using protocol-relative links, but this broke DW-hosted mood themes, icons in emails, etc.

So we need to do this a better way!

### LINKED ISSUE #1138: Remove blank space after tag and before comma in lists without tag arrows
When you look at tags, there's a blank space before the comma on journal pages without arrows, which didn't use to be there before. Any page, any style. See http://dw-dev.dreamwidth.org/ for example.

afuna:
This means that we cache it as "//" rather than "http://" or "https://"
so it works both pages.

A bit hacky, but makes sure we're limited in scope to website-only
(journals only), and not affecting stuff like email.

Fixes #1134.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2442,6 +2442,9 @@ sub Image_std
     my $name = shift;
     my $ctx = $LJ::S2::CURR_CTX or die "No S2 context available ";
 
+    my $imgprefix = $LJ::IMGPREFIX;
+    $imgprefix =~ s/^https?://;
+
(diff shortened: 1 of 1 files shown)


afuna:
@zorkian this was the IMGROOT caching code you were referring to before right? or was there something else?

zorkian:
Yup this was it!

PR #1146, opened 2015-01-25
Remove cmdbuffer

zorkian:
Wow, this was vestigial. It was replaced with TheSchwartz a long long
time ago. Nothing actually uses qbufferd/cmdbuffer anymore, so they
should really just go away!


afuna:
Nice! I had no idea what this was for -- apparently it's because it's actually for nothing ;)

afuna:
Forgot to say -- looks good to me. Please fix conflicts and merge.

rahaeli:
Holy crap, that's ancient!

PR #1151, opened 2015-01-25
Remove blockwatch

zorkian:
We don't use it. It's another profiling system that we haven't kept up
with and is rotted. If we ever get crazy huge, we can reinstate it and
bring it up to speed.


zorkian:
PS this also deletes `doc/raw/lj.book` and `doc/raw/ljp.book` which we don't expose anywhere and just sit in our source tree and IMO don't provide any value sitting in our repository. Anything in those files we care about should be moved to the wiki.

afuna:
I don't believe that anyone has looked at those documentation files in *ages* so yeah I'm approving this.

cgi-bin/LJ/Blockwatch.pm
```diff
@@ -1,361 +0,0 @@
...
-    LJ::Blockwatch->end("ddlock");
-}
-
-# Memcache Hooks
-
-sub setup_memcache_hooks {
-    my $class = shift;
-    my $memcache = shift;
-
-    # There are no memcache hooks anymore, but mark my words... someday they will come back.
```
rahaeli: This comment cracks me up.

Another one bites the dust!

Issue #1126, opened 2015-01-19
tweaks to tag nav forward/back arrows


rahaeli:
Unless anybody else can come up with a better solution that addresses all the problems people have with this, my proposed solution is:

a) Give the forward/back arrows a CSS class so they can be hit with custom CSS or a Stylish script;

b) Toggle the display/nondisplay of the forward/back arrows by tag with a turnable arrow at the start of the tag module, much like on the profile. Hitting that arrow will remove all the forward/back arrows in the tag block (and redraw the tag list, obvs), and the system should remember last used value.

afuna:
Let's brainstorm a bit. Discussed with Mark, came up with the idea of
having something (arrows?) by the "Tags:" label which turns on tag
navigation mode. Then you can select the tag and just go forward/back
through it.

This solves the issue of clutter, without causing issues as in the case
of hover, etc.

But how to flesh out the interface further? Any way to make this work
without clunkiness?

rahaeli:
I like that idea, too!  I don't really have much of a sense of what it would look like, though.

ninetyd:
Dunno if this should be a separate issue but there's a blank space before the comma on journal pages without arrows, which didn't use to be there before. Any page, any style. See http://dw-dev.dreamwidth.org/ for example.

afuna:
new issue please! thank you :)

afuna:
Okay couple proposals.

Currentl form, we have the tag navigation arrows on all tags when we're in tags view or entry/reply view:

![screenshot 2015-01-23 12 27 15](https://cloud.githubusercontent.com/assets/364169/5882130/909e554c-a2fb-11e4-8b66-55af495555e1.png)

    Tags: <-- foo -->, <-- bar -->

Alternative proposals:
1) Turn on tag edit mode, inline fashion

* in tags view and entry/reply view, start with one set of arrows for the tags section (not one set overall)

    Tags: < > foo, bar

* clicking an arrow turns on tag navigation mode (additional styling on tags to make it clear that clicking on them does something different -- that is, we're picking the tag to navigate through). Clicking the other arrow turns navigation on that way; clicking the current arrow turns off navigation mode entirely. Clicking the tag navigates in the way indicated by the selected arrow.

    Tags: **≤** > foo, bar
                   
* on subsequent pages, the selected tag is bolded or otherwise emphasized to show we're still in the previous mode

    Tags: **≤** > **foo**, bar



[...]

PR #1153, opened 2015-01-25
Additional CSS class "tagnav" to provide customization options for tag

woggy:
Partial fix to #1126

### DIFF
--- a/styles/core2.s2
+++ b/styles/core2.s2
@@ -5314,11 +5314,11 @@ function Entry::print_tags() [fixed] {
             "<li>";
             if ($show_tagnav_links) {
                 $this->print_link_tag_prev();
-                print " ";
+                """<span class="tagnav-space"> </span>""";
             }
             """<a rel="tag" href="$t.url">$t.name</a>""";
             if ($show_tagnav_links) {
-                print " ";
+                """<span class="tagnav-space"> </span>""";
                 $this->print_link_tag_next();
             }
             $tag_count++;
--- a/styles/core2.s2
+++ b/styles/core2.s2
@@ -5447,11 +5447,11 @@ function Entry::print_link_prev() {
 }
 function Entry::print_link_tag_next() {
     var Link link = $this->get_link("nav_tag_next");
-    """<a href="$link.url">&rarr;</a>""";
+    """<a class="tagnav" href="$link.url">&rarr;</a>""";
 }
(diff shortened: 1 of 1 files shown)


woggy:
And to address the Most Common Complaint:

.tagnav { display:none; } 
.tagnav-space { display:none; }

in the CSS will make the arrows Gone As Though They Never Were.

azurelunatic:
<3

afuna:
Hmmm. ponder ponder. For the space, how about we just do:

    """<a class="tagnav" href="$link.url">&rarr;</a> """;

(that is, have the space inside the print_tagnav_prev/next bits) so that it's easier to manipulate all around?

woggy:
There was some discussion about whether folks wanted there to be spaces or not - figured we'd give them a separate handle so as to allow either option.

afuna:
How about we do something like:

    <span class="tagnav"><a...></a> </span>

And that way there's only one class to target instead of having to look up two of them?

woggy:
Some users requested the ability to remove the space between the arrows and
the tag name. if that's not a feature we want to support, then yes, I agree
that lumping everything under a single class is cleaner.

afuna:
Gotcha. I suggest having just one class for now and then let's see how
much that changes people's experience of the tags!


On Tue, Jan 27, 2015, at 06:47 AM, woggy wrote:
> Some users requested the ability to remove the space between the
> arrows and
>
> the tag name. if that's not a feature we want to support, then
> yes, I agree
>
> that lumping everything under a single class is cleaner.
>
> On Jan 26, 2015 1:50 PM, "Afuna" <notifications@github.com> wrote:
>
>
> > How about we do something like:
>
> >
>
> > <span class="tagnav"><a...></a> </span>
>
> >
>
> > And that way there's only one class to target instead of having to
> > look up
>
> > two of them?
>
> >
>
> > —
>
> > Reply to this email directly or view it on GitHub
>
> > <https://github.com/dreamwidth/dw-free/pull/1153#issuecomment-71535420>.
>
> >

>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/1153#issuecomment-71554980

woggy:
done and tested. can smoosh the commits if need be, but I think it should be relatively clean as-is.

afuna:
Great, thank you!

PR #1158, opened 2015-01-26
Convert /export to tt/foundation

zorkian:
This converts the /export page to the Foundation/TT system.


zorkian:
Strong review requested!

cgi-bin/DW/Controller/Export.pm
```diff
@@ -0,0 +1,51 @@
...
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself. For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Controller::Export;
+
+use v5.10;
+use strict;
+use warnings;
```
afuna: We don't actually use warnings on a per-file level; it's on for dev servers and off for prod.

views/export/index.tt
```diff
@@ -0,0 +1,106 @@
...
+
+This program is free software; you may redistribute it and/or modify it under
+the same terms as Perl itself.  For a copy of the license, please reference
+'perldoc perlartistic' or 'perldoc perlgpl'.
+
+%][%- sections.title = title_ml | ml -%]
+
+[%- CALL dw.ml_scope( '/export.bml' ) -%]
+[%- CALL dw.active_resource_group( "foundation" ) -%]
+[% sections.title = '.title' | ml %]
```
afuna: sections.title is set twice

views/export/index.tt
```diff
@@ -0,0 +1,106 @@
...
+
+[%- authas_form -%]
+
+<p>[% '.description' | ml %]</p>
+
+<form action='export_do?authas=[% u.username %]' method='post'>
+
+<fieldset>
+
+<div class="row">
```
afuna: If you wanted to restrict the width of the entire form, might be easier to just do:

    <fieldset class="row"><div class="columns large-x"></div></fieldset>

Use the offset / end classes from http://foundation.zurb.com/docs/components/grid.html too as appropriate to remove the need for the divs.

views/export/index.tt
```diff
@@ -0,0 +1,106 @@
...
+    <div class="large-2 columns"></div>
+    <div class="large-2 columns"><label>[% '.label.what' | ml %]</label></div>
```
afuna: To align the label better, you can use `label.inline` from http://foundation.zurb.com/docs/components/forms.html

But -- unless there's a very strong reason to do it this way, I would strongly suggest keeping it consistent with the rest of the site by having the label stack vertically. Otherwise, we're right back where we started with each form on each page having a different look and feel.

I do see what you mean about not wanting the form to be too large though, but for that I would suggest two key fixes:

* remove the "export what" dropdown. We only have one option! We don't need to offer a choice.
* we could make the "Fields" checkboxes go into two columns, too

afuna: Specific to the coding of the label: we want the label to be mapped to its relevant form field, e.g.,

    <label for="foo">...</label>
    <input type="..." id="foo">

This helps a lot with accessibility because it makes it programmatically possible to know what the actual label for the form field is for.

It wasn't clear from the style-guide, but (assuming vertical alignment of labels), but you can use the `[% form.x %]`  defined in FormHTML.pm so you don't have to remember all the little details around form elements all the time.

htdocs/export.bml.text
```diff
@@ -29,7 +29,7 @@
 
 .label.format=Format:
 
-.label.header=Header:
+.label.header=Include header line
```
afuna: I like this text so much better... but I don't see it on the new version of the page!

afuna:
Strongly reviewed as requested!

zorkian:
Updated as per requested. New page:

http://puu.sh/fba3h/281a570f4c.png

rahaeli:
lol that page. (Not your conversion of it, the page itself!) I should have put that in my talk :P

afuna:
Looks great, I like the streamlining of the options in particular. (Also, yay foundation)

zorkian:
Thanks for showing this old dog some new tricks! That Foundation stuff is pretty fancy, and it was a lot easier to work with than BML/HTML. ;)

Issue #693, opened 2014-04-03
Create nice month view for siteviews


anall:
[ former bug 1501 ] 

The month view is the last remaining journal-styleable view that is still using a BML page for ?style=site by default.

Compare:

http://exor674.dreamwidth.org/2013/06/?style=site
http://exor674.dreamwidth.org/2013/06/?style=site&fallback=s2

afuna:
Poking my head in! How much more work is needed to get this done? :)

zorkian:
@momijizukamori : that code you have above, is that supposed to be a pull request?

momijizukamori:
Iiiii thought this was already merged? If it's not I'll go back and poke at it

afuna:
Let me look into what happened and get back to you!

afuna:
Ahh I do see what happened. We had the code here, but we were still using the BML page by default. Fixed that in #1160, and removing the s1shortcomings bml file! Hooray.

momijizukamori:
Ahhh okay. Excellent!

kareila:
I'm working on the code tour and just noticed this issue is still open.  Is there anything that still needs to be done?

afuna:
Nope, I think we just didn't close it!

kareila:
I hereby declare this issue closed!  Anything we missed can open a new issue.

PR #1161, opened 2015-01-27
Use S2-based monthpage for style=site

afuna:
This exposes the work done in #693 by default, and removes the
talkread_s1short.bml file.


afuna:
Old:
screen shot 2015-01-26 at 6 22 04 pm

New:
screen shot 2015-01-26 at 6 21 34 pm

PR #1162, opened 2015-01-27
Shrink authas select width, space apart the button a bit

afuna:


afuna:
Old:
screen shot 2015-01-26 at 6 22 04 pm

New:
screen shot 2015-01-26 at 6 21 34 pm

rahaeli:
Yay!

PR #1164, opened 2015-01-27
Remove the last remnants of BML for journal views

afuna:


anall:
This seems to be what I was thinking of. +1<3

afuna:
Great! Thanks for reminding me to take care of that :)

PR #1165, opened 2015-01-27
Remove CProd

### LINKED ISSUE #1031: Remove CProd from the codebase
(I believe that this used to be in bugzilla)

Don't think we've used this, or plan to -- and it's one less thing to convert over if not needed.

kareila:
Fixes #1031 but hasn't been tested.


afuna:
good bye cprod

PR #1168, opened 2015-01-28
Bug 3965 - better organization of LJ functions (last gasp)

kareila:
This is the last of the work I had done cutting functions out of ljlib.pl and putting them in other modules, in order to reduce ljlib.pl to a more manageable size.  It got sidelined waiting for some other changes to be merged in on one of the affected files.  I don't think I ever tested these changes in their final form, but I don't anticipate issues.


cgi-bin/LJ/Entry.pm
```diff
@@ -2045,6 +2045,70 @@ sub load_log_props2
 }
 
 # <LJFUNC>
+# name: LJ::load_talk_props2
```
zorkian: Since talkprops are all for comments, wouldn't these better go in `LJ::Comment`?

cgi-bin/LJ/Entry.pm
```diff
@@ -2395,6 +2459,92 @@ sub get_logtext2
 }
 
 # <LJFUNC>
+# name: LJ::get_talktext2
```
zorkian: Same with this one re: `LJ::Comment` module.

cgi-bin/LJ/Entry.pm
kareila: The problem I saw with using LJ::Comment is that it doesn't already include any functions in package LJ, whereas LJ::Entry already includes package LJ functions for both entries and comments (see LJ::delete_all_comments, LJ::delete_comments).  Also LJ::Entry is already included in all the places where the functions being moved are called.  It's much easier and more consistent to do it this way, in my view.

zorkian:
Thank you for the explanation! That makes sense.

PR #1171, opened 2015-01-29
Bug 5242 - update all events to use LJ::Event::raw_subscriptions

kareila:
This is a cleanup project I undertook following Dre's work to remove legacy notification code and have everything use ESN.  I never submitted a pull request for it because of all the various projects-in-progress I have cluttering up my dreamhack, preventing me from pulling down a recent develop branch to test with.

That said, it's straightforward refactoring and should be OK to merge.  The first commit updates the method call and the second commit refactors common code into a shared method.  The only thing I haven't done is to see if any new events have been added in the intervening time that would benefit from similar changes.

### DIFF
--- a/cgi-bin/LJ/Event.pm
+++ b/cgi-bin/LJ/Event.pm
@@ -431,6 +431,40 @@ sub raw_subscriptions {
     return @subs;
 }
 
+# helper method to be called when overriding parent's raw_subscriptions
+# method to always return a subscription object for the user
+sub _raw_always_subscribed {
+    my ( $class, $self, %args ) = @_;
+    my $cid = delete $args{'cluster'};
+    croak("Cluser id (cluster) must be provided") unless defined $cid;
+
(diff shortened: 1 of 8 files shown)


afuna:
The only event with subscriptions/raw_subscriptions that hasn't been overridden is `cgi-bin/LJ/Event/JournalNewComment/Reply.pm`. Deliberate?

kareila:
Yes, the raw_subscriptions method in `JournalNewComment/Reply.pm` is overridden for a different reason and the code is substantially different from the code in the modules I changed, which forces the user to always have a subscription to that event (hence the name `_raw_always_subscribed`).  It looks like what `JournalNewComment/Reply.pm` does is to provide a bridge between old-style opt_gettalkemail preference and new-style ESN subscription, so it's currently a unique snowflake in that regard.

afuna:
Sounds good to me!

PR #1175, opened 2015-01-30
Remove LJTalk

### LINKED ISSUE #1109: remove ljtalk references
While code-diving to pick out some examples of "stuff that is lurking in the bottom of the codebase" for a talk I'm giving, I found references to LJ Talk in User.pm (section 18) -- all the LJ Talk stuff can be pulled out, since we don't run a Jabber instance and don't plan on doing so.

Might be more than just in User.pm, so do some digging to see what else can be yanked. (NB, this does not include the Jabber field on the profile; that can be any Jabber service.)

kareila:
Fixes #1109.  Removes all code and settings related to running our own jabber server.

I was able to verify this as far as making sure the web server on my dreamhack would still start up with these lines removed.


zorkian:
Thank you for doing this!

kareila:
That was fast :)

I was surprised how much code we still had for this...

zorkian:
Haha. I'm sitting here with some Talisker 10 year waiting for an all hands meeting to start, seemed like the perfect time to look at some code removal.

PR #312, opened 2013-03-11
Bug2152 prime

annariel:
A rebased version of Bug2152.  Ready for (as yet unmade) changes requested from code review.

PR #1182, opened 2015-02-03
[#312] Allow multiple stickies

afuna:
* Allows multiple stickies to be set both in the post forms via a sticky
  module, and in the Display tab under Manage Account

* Allows for different sticky allocations depending upon account type
  and attempts to handle switching between account types elegantly

Much thanks to @annariel for her initial work on this.

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -2915,9 +2915,17 @@ setting.sitescheme.label=Site Skin
 
 setting.sitescheme.label=Site Scheme
 
-setting.stickyentry.error.invalid=Invalid Dreamwidth entry ID or URL entered.
+setting.stickyentry.details.label=Note: blank entries will be ignored.
+
+setting.stickyentry.error.invalid2=Invalid Dreamwidth entry ID or URL entered for Sticky [[stickyid]].
+
+setting.stickyentry.error.duplicate=Duplicate ID or URL entered for Sticky [[stickyid]].
 
 setting.stickyentry.label=ID or URL of entry to make sticky
+
+setting.stickyentry.label2=IDs or URLs of Sticky Entries
+
+setting.stickyentryi.label=Sticky [[stickyid]]
 
 setting.stylemine.label=Comment Pages
 
--- a/cgi-bin/DW/Setting/StickyEntry.pm
+++ b/cgi-bin/DW/Setting/StickyEntry.pm
@@ -14,14 +14,13 @@
 package DW::Setting::StickyEntry;
 use base 'LJ::Setting';
 use strict;
-use warnings;
 
 sub should_render {
     $_[1] ? 1 : 0;
 }
 
 sub label {
-    $_[0]->ml( 'setting.stickyentry.label' );
+    $_[0]->ml( 'setting.stickyentry.label2' );
 }
 
 sub option {
@@ -30,17 +29,40 @@ sub option {
     my $key = $class->pkgkey;
     my $ret;
 
-    $ret .= LJ::html_text({
-        name  => "${key}stickyid",
-        id    => "${key}stickyid",
-        class => "text",
-        value => $errs ? $class->get_arg( $args, "stickyid" ) : $u->sticky_entry,
-        size  => 30,
-        maxlength => 100,
-    });
+    my @stickies = $u->sticky_entries;
+    my $username = $u->user;
 
+    foreach my $i ( 1...$u->count_max_stickies ) {
+        my $url = "";
+        my $placeholder = "";
+        my $e = $stickies[$i - 1];
+        if ( $e ) {
+            $url = $e->url;
+        } else {
+            $placeholder = $u->journal_base . "/1234.html" if $i == 1;
+        }
+        my $textentry = $errs ? $class->get_arg( $args, "stickyid${i}" ) : $url;
+
+        $ret .= "<label for='${key}stickyid${i}'>" . $class->ml( 'setting.stickyentryi.label', { stickyid => $i } ) . " </label>";
+        $ret .= LJ::html_text({
+            name  => "${key}stickyid${i}",
+            id    => "${key}stickyid${i}",
+            class => "text",
+            value => $textentry,
+            placeholder => $textentry ? "" : $placeholder,
+            size  => 50,
+            maxlength => 100,
+        });
+        $ret .= q{ - <a href='} . $e->url . q{'>} . ( $e->subject_html || "(no subject)" ) . q{</a>} if $e;
+        $ret .= "<br>";
+
+    }
+
+    # returns an error if any of the stickies are incorrectly formatted.
     my $errdiv = $class->errdiv( $errs, "stickyid" );
-    $ret .= "<br />$errdiv" if $errdiv;
+    $ret .= "$errdiv<br>" if $errdiv;
+
+    $ret .= "<em>" . $class->ml( 'setting.stickyentry.details.label' ) . "</em>";
 
     return $ret;
 }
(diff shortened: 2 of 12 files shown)


bin/upgrading/en.dat
```diff
@@ -2915,10 +2915,18 @@ setting.sitescheme.label=Site Skin
 
 setting.sitescheme.label=Site Scheme
 
-setting.stickyentry.error.invalid=Invalid Dreamwidth entry ID or URL entered.
+setting.stickyentry.details.label=<i>Note: blank entries will be ignored.</i>
```
zorkian: I thought we wanted to use `em` these days? And don't we try to avoid HTML in these kinds of things?

cgi-bin/DW/Setting/StickyEntry.pm
```diff
@@ -30,17 +29,40 @@ sub option {
...
+    my $username = $u->user;
 
+    foreach my $i ( 1...$u->count_max_stickies ) {
+        my $url = "";
+        my $placeholder = "";
+        my $e = $stickies[$i - 1];
+        if ( $e ) {
+            $url = $e->url;
+        } else {
+            $placeholder = $u->journal_base . "/1234.html" if ( $i == 1 );
```
zorkian: No parenthesis are necessary around postfix conditionals.

cgi-bin/DW/Setting/StickyEntry.pm
```diff
@@ -30,17 +29,40 @@ sub option {
...
+        }
+        my $textentry = $errs ? $class->get_arg( $args, "stickyid${i}" ) : $url;
+
+        $ret .= "<label for='${key}stickyid${i}'>" . $class->ml( 'setting.stickyentryi.label', { stickyid => $i } ) . " </label>";
+        $ret .= LJ::html_text({
+            name  => "${key}stickyid${i}",
+            id    => "${key}stickyid${i}",
+            class => "text",
+            value => "$textentry",
```
zorkian: Why is this quoted?

cgi-bin/DW/Setting/StickyEntry.pm
```diff
@@ -30,17 +29,40 @@ sub option {
...
+        });
+        $ret .= q{ - <a href='} . $e->url . q{'>} . ( $e->subject_html || "(no subject)" ) . q{</a>} if $e;
+        $ret .= "<br>";
+
+    }
+
+    # returns an error if any of the stickies are incorrectly formatted.
     my $errdiv = $class->errdiv( $errs, "stickyid" );
-    $ret .= "<br />$errdiv" if $errdiv;
+    $ret .= "$errdiv<br>" if $errdiv;
```
zorkian: `<br />` should be self-closed.

cgi-bin/DW/Setting/StickyEntry.pm
```diff
@@ -48,11 +70,38 @@ sub option {
...
+    my $max_sticky_count = $u->count_max_stickies;
+    my @stickies;
+    # Create a hash that we will use to check for duplicate entries.
+    my %unique = ();
+    my $username = $u->user;
+
+    for my $i ( 1 ... $max_sticky_count ) {
+        my $stickyi = $class->get_arg( $args, "stickyid${i}" ) || '';
+        # Unless this is a blank form entry...
+        unless ( $stickyi eq '' ) {
```
zorkian: This `unless` should be de-dented...

```perl
next if $stickyi;
```

Or something?

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
+    my $max_sticky_count_from_zero = $max_sticky_count - 1;
+    my $entry_length = @entry_ids;
+
+    my @currently_unused_stickies = @entry_ids[$max_sticky_count..$entry_length];
 
-    if ( defined $input ) {
-        unless ( $input ) {
+    # Check we've been sent input and it isn't empty.  If so we need to alter the sticky entries stored.
+    if ( defined $input_ref ) {
+        my $remote = LJ::get_remote();
```
zorkian: This method is on the `LJ::User` object, which means it shouldn't be checking `$remote`. That should be done at the input -- i.e., the API/page/etc that is making the call.

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
 
-        my $ditemid;
-        my $slug;
+        # sanity check the elements of the input array of candidate stickies.
+        my $new_sticky_count = 0;
+        foreach my $sticky_input ( @input ) {
+            $new_sticky_count++;
 
-        # takes ditemid/ditemid from URL
-        if ( $input =~ m!/(\d+)\.html! ) {
```
zorkian: This might want to be tail (`$`) anchored?

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
-            $ditemid = $1;
-        } elsif ( $input =~ m!(\d+)! ) {
```
zorkian: Head and tail anchors?

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
         }
 
-        # takes slug from URL
-        $slug = $1
-            if $input =~ m!^(?:.+)/(?:\d\d\d\d/\d\d/\d\d)/([a-z0-9_-]+)\.html$!;
+        # The user may have reused a sticky from before their account was downgraded.  To keep
+        # stickies unique we should remove this from the list of unused stickies.
+        my @new_unused_stickies;
+        # We create a hash from the input for quick membership checking.
+        my %sticky_hash = map { $_ =>  1 }  @input;
```
zorkian: Random two spaces before `@input`.

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
+
+    # The user may have previously had an account type that allowed more stickes.
+    # we want to preserve a record of these additional stickes in case they once
+    # more upgrade their account.  This means we must first extract these
+    # if they exist.
+    my $sticky_entry_ids =  $u->prop( 'sticky_entry' );
+    my @entry_ids = split( /,/, $sticky_entry_ids );
+
+    my $max_sticky_count = $u->count_max_stickies;
+    my $max_sticky_count_from_zero = $max_sticky_count - 1;
```
zorkian: Is this variable used anywhere?

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
 
-        $u->set_prop( sticky_entry => $ditemid );
-        return 1;
-    }
-    return $u->prop( 'sticky_entry' );
+# Make a particular entry into a particular sticky.
+sub sticky_entry_new {
+    my ( $u, $ditemid ) = @_;
+
+    my $remote = LJ::get_remote();
```
zorkian: No checking `$remote` in `LJ::User` methods.

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
+# Remove a particular entry from the sticky list
+sub sticky_entry_remove {
+    my ( $u, $ditemid ) = @_;
 
-    if ( my $ditemid = $u->sticky_entry ) {
-        my $item = LJ::Entry->new( $u, ditemid => $ditemid );
-        return $item if $item->valid;
-    }
-    return undef;
+    my $remote = LJ::get_remote();
```
zorkian: No checking `$remote` in `LJ::User` methods.

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
+    return unless $remote && $remote->can_manage( $u );
+
+    my $sticky_entry_ids =  $u->prop( 'sticky_entry' );
+    my @stickies = split( /,/, $sticky_entry_ids );
```
zorkian: This particular logic is done in several places and should probably be abstracted, or you should just call `sticky_entries_lookup` here.

cgi-bin/LJ/User.pm
```diff
@@ -3195,61 +3198,121 @@ sub thread_expand_all {
...
+
+    my @new_stickies;
+
+    @new_stickies  = grep { $_ != $ditemid } @stickies;
```
zorkian: Combine this and the line above... no sense in defining it just to fill it.

cgi-bin/DW/Setting/StickyEntry.pm
afuna: That's true for XHTML, but it's optional on HTML5 (which is what we're aiming for now).

cgi-bin/DW/Setting/StickyEntry.pm
afuna: Hmmm I think "next unless $stickyi" because the original, I just realized, is a double-negative.

cgi-bin/DW/Setting/StickyEntry.pm
zorkian: Mind blown.

--
Mark Smith mark@qq.is


On Mon, Feb 2, 2015, at 10:23 PM, Afuna wrote:
> In cgi-bin/DW/Setting/StickyEntry.pm[1]:


> >      my $errdiv = $class->errdiv( $errs, "stickyid" );
> -    $ret .= "<br />$errdiv" if $errdiv;
> +    $ret .= "$errdiv<br>" if $errdiv; That's true for XHTML, but it's
>      optional on HTML5 (which is what we're aiming for now).


> — Reply to this email directly or view it on GitHub[2].

>


Links:

  1. https://github.com/dreamwidth/dw-free/pull/1182#discussion_r23985084
  2. https://github.com/dreamwidth/dw-free/pull/1182/files#r23985084

cgi-bin/LJ/User.pm
afuna: I don't think that sticky_entries_lookup will work here because that won't maintain the order, but I'm taking the hint and making a convenience method ;)

afuna:
Done

cgi-bin/DW/Setting/StickyEntry.pm
```diff
@@ -48,11 +70,39 @@ sub option {
...
+            push @stickies,  $ditemid;
+            $unique{$ditemid} = 1;
+        } else {
+            # As soon as we detect a problem with a sticky we break out of the subroutine.
+            $class->errors( "stickyid" => ( $class->ml( 'setting.stickyentry.error.invalid2', { stickyid => $i } ) ) ) ;
+            return 1;
+        }
     }
+
+    my $remote = LJ::get_remote();
```
zorkian: Authorization is done at the _page_ level, not in a setting.

So, this needs to be in the controller. I don't think authorization should go anywhere else.

afuna: Good point. settings.bml already does this for me, so I've removed these two lines and pushed up my changes again.

zorkian:
Awesome, thank you! :)

PR #1186, opened 2015-02-04
CSS tweaks for /admin/console

kareila:
This applies Foundation-friendly CSS to the affected area but doesn't do a full conversion.

### DIFF
--- a/views/admin/console/index.tt
+++ b/views/admin/console/index.tt
@@ -16,6 +16,9 @@ the same terms as Perl itself.  For a copy of the license, please reference
 [%- sections.title = ".admin.link" | ml -%]
 [%- CALL dw.active_resource_group( "foundation" ) -%]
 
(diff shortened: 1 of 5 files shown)


views/admin/console/index.tt
```diff
@@ -16,6 +16,9 @@ the same terms as Perl itself.  For a copy of the license, please reference
 [%- sections.title = ".admin.link" | ml -%]
 [%- CALL dw.active_resource_group( "foundation" ) -%]
 
+[%- dw.need_res( { group => "foundation" }
```
zorkian: Is this missing a trailing comma?

afuna: Nah, it's optional (because TT)

kareila:
Squashed and repushed.

zorkian:
TT is magic! Thank you!

PR #1191, opened 2015-02-04
Add sandbox attribute to our embed iframes

afuna:
The current set should allow/disallow the same things as before, with
the exception of full-page redirects (not allowed)


zorkian:
What does the `sandbox` attribute do? It seems a little scary to say to allow scripts etc. Without this attribute, are those items allowed?

afuna:
Good writeup of the `sandbox` attribute here:
http://www.html5rocks.com/en/tutorials/security/sandboxed-iframes/

The sandbox attribute allows us to restrict iframes to only those
capabilities we choose to whitelist. Without the sandbox attribute,
these things are all already allowed.

allow-scripts allows JS execution, but only within the iframe (it can't
break out into the top-level window)

allow-same-origin allows the iframe access to same-origin (that is,
youtube.com can access its own cookies -- but not those of
embedded.dreamwidth.net)

allow-forms is for form submission

I've found that in general these three are required to get popular
embeds working (e.g, youtube, vimeo).



On Wed, Feb 4, 2015, at 12:35 PM, Mark Smith wrote:
> What does the sandbox attribute do? It seems a little scary to say to
> allow scripts etc. Without this attribute, are those items allowed?


>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/1191#issuecomment-72933580

PR #1197, opened 2015-02-06
/entry/new changes round 6

afuna:
This is the final round of miscellaneous changes. Anything after this
can go in its own bug.


afuna:
wrong branch

Issue #1195, opened 2015-02-05
(Bug 177) implement /admin/sendmail


kareila:
The "send message" page was implemented in #799.  Still needed:

* Fix missing strings /admin/sendmail/index.tt.admin.link and .admin.text

* Better error display flow in /admin/sendmail/index

* Add /admin/sendmail/lookup for reviewing previously sent messages

* Add /admin/sendmail/forms for managing frequently used text

kareila:
All that's left for this project now is adding /admin/sendmail/forms and integrating it with /admin/sendmail/index.

afuna:
:+1:

PR #1201, opened 2015-02-09
[#1195] /admin/sendmail/lookup

kareila:
The lookup page shows the list of previously sent messages for the selected account.
Clicking on the subject link for a message in the list displays the contents.

### DIFF
--- a/cgi-bin/DW/Controller/Admin/SendMail.pm
+++ b/cgi-bin/DW/Controller/Admin/SendMail.pm
@@ -34,8 +34,48 @@ DW::Controller::Admin->register_admin_page( '/',
 
 DW::Routing->register_string( "/admin/sendmail/index", \&index_controller );
 DW::Routing->register_string( "/admin/sendmail/lookup", \&lookup_controller );
+DW::Routing->register_string( "/admin/sendmail/message", \&message_controller );
 # FIXME: add this later
 #DW::Routing->register_string( "/admin/sendmail/forms", \&form_controller );
+
+# helper function for lookup and message
+my $enhance_row = sub {
+    my ( $row ) = @_;
+
+    # link to view full message
+    $row->{msgurl} = LJ::create_url( "/admin/sendmail/message",
+                       args => { id => $row->{msgid} } );
+
+    # time_sent needs a display version
+    $row->{time_sent_view} = LJ::time_to_http( $row->{time_sent} );
+
+    # build a request URL
+    if ( $row->{request} ) {
+        $row->{request_url} = LJ::create_url( "/support/see_request",
+                                args => { id => $row->{request} } );
+    }
+
+    # sendto might be a uid (if it's an email, no work is needed)
+    if ( $row->{sendto} =~ /^\d+$/ ) {
+        if ( my $u = LJ::load_userid( $row->{sendto} ) ) {
+            $row->{sendto} = $u;
+        } else {
+            $row->{sendto} = '[unknown user] '. $row->{sendto};
+        }
+    }
+
+    # sentfrom is definitely a uid
+    if ( my $ru = LJ::load_userid( $row->{remoteid} ) ) {
+        $row->{remote} = $ru;
+    } else {
+        $row->{remote} = '[unknown user] '. $row->{remoteid};
+    }
+
+    # add domain name to display site address
+    $row->{account_view} = $row->{account} . "\@$LJ::DOMAIN";
+
+    return $row;
+};
 
 sub index_controller {
     my ( $ok, $rv ) = controller( privcheck => \@privs, form_auth => 1 );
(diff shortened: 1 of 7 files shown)


cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+    my ( $ok, $rv ) = controller( privcheck => \@privs, form_auth => 1 );
+    return $rv unless $ok;
+    my $remote = $rv->{remote};
+
+    my $scope = sub { return '/admin/sendmail/lookup.tt' . $_[0] };
+
+    my $enhance_row = sub {
+        my ( $row ) = @_;
+        # link to view full message
+        $row->{msgurl} = '?id=' . $row->{msgid};
```
afuna: Hrrm, are we always sure that the current page has no arguments? (I think if there are, this will lead to an invalid url)

I suggest using LJ::create_url( undef, ...) instead to build the url.

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+        # time_sent needs a display version
+        $row->{time_sent_view} = LJ::time_to_http( $row->{time_sent} );
+        # build a request URL
+        if ( $row->{request} ) {
+            $row->{request_url} = "/support/see_request?id=";
```
afuna: LJ::create_url here, too pls.

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+            $row->{request_url} .= $row->{request};
+        }
+        # sendto might be a uid
```
afuna: what happens if sendto is not a uid?

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+        if ( $row->{sendto} =~ /^\d+$/ ) {
+            if ( my $u = LJ::load_userid( $row->{sendto} ) ) {
+                $row->{sendto} = $u;
+            } else {
+                $row->{sendto} = '[unknown user] '. $row->{sendto};
+            }
+        }
+        # sentfrom is definitely a uid
```
afuna: nitpick: blank line above the comment please

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+    my $r = DW::Request->get;
+    my $errors = DW::FormErrors->new;
+
+    if ( $r->did_post ) {
+        my $args = $r->post_args;
+
+        my $account = LJ::trim( $args->{account} );
+
+        if ( $account ) {
+            if ( my $priv = $LJ::SENDMAIL_ACCOUNTS{ $account } ) {
```
afuna: style nitpick: no space needed within a hash lookup.

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+            die $dbr->errstr if $dbr->err;
+
+            my @data = map { $enhance_row->( $_ ) } values %$rows;
+
+            $rv->{rows} = [ sort { $b->{time_sent} <=> $a->{time_sent} } @data ];
+            $rv->{account}  = $account;
+        }
+
+    } else {
+        my $args = $r->get_args;
```
afuna: Oh hmmmm. I found this slightly confusing. Would it be easier to have a different URL for each: lookup/account and lookup/message maybe? There's not much shared logic between the two. what there is (enhance_row) can be made into  convenience method in the controller itself.

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+            my $dbr = LJ::get_db_reader();
+            my $row = $dbr->selectrow_hashref(
+                "SELECT * FROM siteadmin_email_history WHERE msgid=?",
+                undef, $msgid );
+            die $dbr->errstr if $dbr->err;
+
+            return error_ml( $scope->( '.error.nomsg' ) ) unless $row;
+            $row = $enhance_row->( $row );
+
+            my $priv = $LJ::SENDMAIL_ACCOUNTS{ $row->{account} };
```
afuna: no space please

cgi-bin/DW/Controller/Admin/SendMail.pm
```diff
@@ -171,5 +171,122 @@ sub index_controller {
...
+            } else {
+                return error_ml( $scope->( '.error.nopriv' ), { priv => $priv } );
+            }
+
+            return DW::Template->render_template( 'admin/sendmail/message.tt', $rv );
+        }
+    }
+
+    # Construct data for dropdown of available email addresses
+    my @account_menu = ( "", LJ::Lang::ml( $scope->( '.select.account.choose' ) ) );
```
afuna: I feel like I saw this last iteration as well. Possible candidate for moving into its own method to reduce duplication.

views/admin/sendmail/lookup.tt
```diff
@@ -0,0 +1,63 @@
+[%# Send emails from site accounts.
```
afuna: psst update comment please ;)

views/admin/sendmail/message.tt
```diff
@@ -0,0 +1,57 @@
+[%# Send emails from site accounts.
```
afuna: Update comment here too.

afuna:
Done. Mostly minor changes requested along the lines of style. But please consider the refactor into two handlers!

cgi-bin/DW/Controller/Admin/SendMail.pm
kareila: It just prints the stored value, which is an email address if it's not a uid.

cgi-bin/DW/Controller/Admin/SendMail.pm
kareila: I thought it would be OK to have the behavior be different for a GET vs. a POST but I can look into that.

cgi-bin/DW/Controller/Admin/SendMail.pm
kareila: It was just under my threshold.  I have one more stage of work on this and if I end up reusing it a third time, I will definitely refactor it.

cgi-bin/DW/Controller/Admin/SendMail.pm
afuna: I think it's clearer the other way -- get vs post is, for me, much useful if the get shows a form that then does a post.

kareila:
All set.  I did the refactoring as a separate commit for clarity.  Thanks for reminding me about create_url :)

afuna:
Thanks!

PR #1205, opened 2015-02-09
[#1202] Use "value" instead of "name" for the dropdown

### LINKED ISSUE #1202: Search not working on some Foundation pages
From http://www.dreamwidth.org/support/see_request?id=30359:

Using Search from "( http://www.dreamwidth.org/communities/list or http://www.dreamwidth.org/entry/new for example) I am taken to http://www.dreamwidth.org/multisearch which just displays a blank page for me. It still works fine on pages that use the old template like http://www.dreamwidth.org/.

...And on further testing as I type this, the Region, FAQ, and Email searches work just fine on both page styles. It's just the Interest, Site and Account, and IM Info searches that send me to the blank page and only on the Foundation-styled pages."

Reproducing. Haven't tested this further to see if this affects other pages.

afuna:
Fixes #1202.

### DIFF
--- a/schemes/global.tt
+++ b/schemes/global.tt
@@ -46,7 +46,7 @@
     [ "email", "email" ]
     [ "im", "iminfo" ]
 ] ~%]
-    <option name="[% option.0 %]">[% "widget.search." _ option.1 | ml %]</option>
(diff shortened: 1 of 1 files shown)


afuna:
This is live now on dreamwidth.org

PR #1206, opened 2015-02-11
[#1126] Tag nav overlay triggered by click

afuna:
* to use: click the expand icon next to the "Tags:" label; select a tag;
  click next/previous

* only allows one tag to be navigated at a time

* on next page load, shows the overlay with the tag selected again


afuna:
Or, to illustrate: 

On page load:
step 1

After clicking the plus icon (note you can't do anything until you select a tag):
step 2

Selected a tag:
step 3

Clicked next/previous (new entry loaded)
step 4

afuna:
Have left open for review/comment for a while. Will merge tomorrow (unless someone comments / objects -- and then we can talk!)

PR #1209, opened 2015-02-13
[#1149] Determine protocol in $u->journal_base from $LJ::IS_SSL

### LINKED ISSUE #1149: Entry links from an https journal page should be https as well

afuna:
Instead of passing in an argument for all instances thereof. This makes
sure that if you're on an https page, all DW-generated journal links
from there aso point to https, and same for http.

This does not affect comment notification emails that are sent (those
are always http)

Fixes #1149.

### DIFF
--- a/cgi-bin/DW/Media/Base.pm
+++ b/cgi-bin/DW/Media/Base.pm
@@ -107,7 +107,7 @@ sub url {
             ( $self->{url_height} // $self->{height} ) . '/';
     }
 
-    return $self->u->journal_base( ssl => $LJ::IS_SSL ) . '/file/' . $extra . $self->{displayid} .
+    return $self->u->journal_base . '/file/' . $extra . $self->{displayid} .
         '.' . $self->{ext};
 }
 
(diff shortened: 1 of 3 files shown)


cgi-bin/LJ/S2/EntryPage.pm
```diff
@@ -555,7 +555,7 @@ sub EntryPage_entry
     $comments->{show_readlink} &&= $get_mode eq 'reply';
 
     my $subject = LJ::CleanHTML::quote_html( $entry->subject_html, $get->{nohtml} );
-    my $event = LJ::CleanHTML::quote_html( $entry->event_html, $get->{nohtml} );
+    my $event = LJ::CleanHTML::quote_html( $entry->event_html( { proxy_insecure_content => $LJ::IS_SSL } ), $get->{nohtml} );
```
zorkian: By the logic of this patch, don't we just want to have `proxy_insecure_content` be based on `$LJ::IS_SSL` in the functions? That way you don't have to pass it in?

afuna:
Oh urgh, that second commit shouldn't be there; this should only be about the journal_base changes. The patch you commented on was committed in #1210. 

I'll update this to only have the relevant commit, and then will revisit the other patch to see about passing proxy_insecure_content around less.

afuna:
Pull request is ready for review again.

Issue #871, opened 2014-08-01
HTTPS connections for reading content


brainwane:
I'd like to be able to browse Dreamwidth journals over HTTPS, and be able to post content over an SSL-protected connection as well. Right now I can log in over HTTPS and read pages on dreamwidth.org via HTTPS, e.g., https://www.dreamwidth.org/nav/explore , but not journals.

(This was http://bugs.dwscoalition.org/show_bug.cgi?id=1239 .)

brainwane:
Hi! [Goodness!](http://dw-news.dreamwidth.org/36470.html) Thank you all for doing so much work on this! I really appreciate it!

afuna:
Thanks for kicking off the discussion so we could be sure to pick it up!

afuna:
I'm not sure why we've kept this open. We've added the proxy and we've added the code that applies the proxying to the various pages. I've looked at the comments and we don't appear to have anything pending, so I'm closing this, 

Feel free to reopen if directly relevant, or (better) let's open a new issue if something related comes up.

PR #1210, opened 2015-02-13
[#871] Proxy insecure content for entries on journals

afuna:
This passes the argument which CleanHTML know that we're interested in
proxying insecure content from journal pages.

For #871.


afuna:
I am literally headscratching right now. I'm not sure why this wasn't the case earlier.

zorkian:
That's a good question.

PR #1214, opened 2015-02-14
Remove uses of smartmatch

anall:
Smartmatch got used before I knew it was not something we wanted in the codebase.
Recent Perls complain loudly about this feature, so remove it.

Also adds a test of the feed canonicalizer to make sure I didn't regress something.

### DIFF
--- a/cgi-bin/DW/FeedCanonicalizer.pm
+++ b/cgi-bin/DW/FeedCanonicalizer.pm
@@ -36,7 +36,7 @@ sub canonicalize {
     $uri_string = $uri_string->[0] if ref $uri_string eq 'ARRAY';
 
     my $uri = URI->new( $uri_string )->canonical;
-    return undef unless $uri->scheme ~~ [ qw/ http https / ];
+    return undef unless $uri->scheme =~ m/^(http|https)$/;
     
     $uri->userinfo( undef );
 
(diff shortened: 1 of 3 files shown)


cgi-bin/DW/FeedCanonicalizer.pm
```diff
@@ -49,49 +49,49 @@ sub canonicalize {
...
-    given ( $uri_str ) {
+    {
         # Let's see if this looks "LJ-ish".
-        when( m!^https?://(?:users|community|syndicated)\.([^/]+)/+([a-z0-9\-_]+)/$LJISH_URL_PART$!i ) {
+        if ( $uri_str =~ m!^https?://(?:users|community|syndicated)\.([^/]+)/+([a-z0-9\-_]+)/$LJISH_URL_PART$!i ) {
             my ( $host, $sub, $feed, $extra, $spare ) = ( $1, $2, $3, $4, $5 );
-            continue if $feed ~~ m/^rss/i && ! $LJISH_SITES{$host};
-            continue if $spare && ! $LJISH_SITES{$host};
-            return make_ljish( $host, $sub, $feed, $orig_uri, $extra );
+            return make_ljish( $host, $sub, $feed, $orig_uri, $extra ) unless 0
```
zorkian: Why the `0 or`? Isn't that a no-op?

anall: I can happily not do that, I was just trying to have all the "continue bail-out cases" have the "or $someCase" form v.s. the first one just being " $someCase"

zorkian: Yeah, what do you think about:

```
return make_ljish(...)
    unless condition
        or condition
        or condition;
```

anall: I like that. Updated.

zorkian:
Nice! Thank you so much for doing this. It's nice to see you in pull requests! <3

PR #1215, opened 2015-02-14
Fix some tests.

anall:
Make these tests also respect extlib, which makes it possible
for people to test new modules on a Dreamhack or other shared
coding environment without having to wait for system modules
to be installed.

extlib was already supported in the codebase in general, but
some tests did not respect it.


zorkian:
Nice! Thanks for fixing these up!

PR #1217, opened 2015-02-15
Fix even more tests.

anall:
Some of these were just warnings, but clean tests are good tests.

### DIFF
--- a/t/esn-journalnewcomment.t
+++ b/t/esn-journalnewcomment.t
@@ -194,7 +194,7 @@ test_esn_flow(sub {
     $LJ::CAP{0...15}->{track_thread} = 0;
     $subsc->delete;
 
-    if (LJ::Event::JournalNewComment->zero_journalid_subs_means eq "friends") {
+    if ( ( LJ::Event::JournalNewComment->zero_journalid_subs_means // "" ) eq "friends") {
         ####### S4 (watching new comments on all friends' journals)
 
         $subsc = $u1->subscribe(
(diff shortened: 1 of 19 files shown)


t/esn-journalnewcomment.t
```diff
@@ -194,7 +194,7 @@ test_esn_flow(sub {
     $LJ::CAP{0...15}->{track_thread} = 0;
     $subsc->delete;
 
-    if (LJ::Event::JournalNewComment->zero_journalid_subs_means eq "friends") {
+    if ( ( LJ::Event::JournalNewComment->zero_journalid_subs_means // "undef" ) eq "friends") {
```
afuna: Why the "undef" here? I think we usually use ""?

anall: ... because I thought that was winding up in an error message if that was a test failure! I wanted it to be a nice failure message ( was undef expected friends ) if it was undef.  Can fix!

afuna: Hmmm I think that's not the case unless it's in an assert, so please do!

zorkian:
@afuna Please remember to use the "review: requested changes" label so we know what's where :)

afuna:
Oh right! Noted, I'll do that next time (sorry)

zorkian:
@anall Ping! Do you think you'll have time to get back to updating this? Should be pretty straightforward :)

anall:
Will try sometime this weekend!

zorkian:
Woah fast. Thank you! :heart:

anall:
Updated. That was not this weekend!

afuna:
But merged before next this weekend! Thanks for the fixes.

PR #1223, opened 2015-02-15
Make cleaner closer more intelligently

zorkian:
There is a problem where the HTML cleaner just randomly closes tags in
some ineffective order. Instead, we should actually use the tag stack to
close tags in the order we saw them.

This also removes autoclose as an option: we always want to close when
we're doing HTML cleaning.

Similarly, this removes tablecheck: we always want that on. If someone
is throwing in table related HTML, we should ensure that it can't get
busted out somehow.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -23,7 +23,6 @@ use LJ::Config;
 
 LJ::Config->load;
 
-
 # attempt to mangle an email address for printing out to HTML.  this is
 # kind of futile, but we try anyway.
 sub mangle_email_address {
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1215,12 +1201,10 @@ sub clean
     # don't close tags that don't need a closing tag -- otherwise,
     # we output the closing tags in the wrong place (eg, a </td>
     # after the <table> was closed) causing unnecessary problems
-    if (ref $opts->{'autoclose'} eq "ARRAY") {
-        foreach my $tag (@{$opts->{'autoclose'}}) {
-            next if $tag =~ /^(?:tr|td|th|tbody|thead|tfoot|li)$/;
-            if ($opencount{$tag}) {
-                $newdata .= "</$tag>" x $opencount{$tag};
-            }
+    foreach my $tag ( @tagstack ) {
```
afuna: We want to reverse this.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1215,12 +1201,10 @@ sub clean
...
+        next if $tag =~ $slashclose_tags;
+        if ($opencount{$tag}) {
+            $newdata .= "</$tag>" x $opencount{$tag};
```
afuna: We probably only want to output one close tag for each tag on the stack.

Example for both cases:

    <h1><h2><h3><h1><h2><h3>

Outputs:

     <h1><h2><h3><h1><h2><h3></h1></h1></h2></h2></h3></h3></h1></h1></h2></h2></h3></h3>

zorkian:
Pushed update with fixes.

afuna:
lgtm

PR #1235, opened 2015-02-17
Fixes #1225 Opera Presto confused by Foundation namespaces

### LINKED ISSUE #1225: Modals in /entry/new don't work in Opera (Presto)

deborahgu:
Opera Presto not loading Foundation. FWIW Foundation is only supporting Opera Presto (<=12) by accident at this point; expect continued degradation. Refs http://ilikekillnerds.com/2014/07/foundation-v5-unrecognized-expression-syntax-namespace-fix/ https://github.com/zurb/foundation/issues/5661

### DIFF
--- a/schemes/celerity.tt
+++ b/schemes/celerity.tt
@@ -105,6 +105,7 @@ the same terms as Perl itself.  For a copy of the license, please reference
                     .find(".top-bar-section").removeClass( "top-bar-section" ).end()
             }
           });
+          Foundation.global.namespace = '';
           $(document).foundation();
         </script>
     </body>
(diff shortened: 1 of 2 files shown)


zorkian:
Opera is not a supported browser, so I'm uncomfortable making a hack like this to Foundation to make it work. Setting something called global namespace seems unwise at best, but potentially disastrous in some unforeseen way. None of the provided links actually analyze whether or not this change is safe or what it does.

deborahgu:
If I do / find an analysis of the safety of the change, will you be willing to reconsider?

PR #1237, opened 2015-02-17
[#1222] display user's timezone instead of UTC on S2 entry pages

### LINKED ISSUE #1222: Site skinned comment pages claim UTC at all times, even when inappropriate
Site skinned comment pages claim the entry was posted in UTC even when the user has selected a timezone. 

Displaying the user's timezone (if set) would be an excellent life choice.

kareila:
This removes the hardcoded 'UTC' display from CSS in favor of
displaying the correct time zone as set by the user viewing the page.
It does so by adding a new 'timezone' property to the Page class in core2.

I don't know if this is the best approach, but it seems to solve the issue.

Fixes #1222.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2328,6 +2328,7 @@ sub Page
         'layout_url' => $layouturl,
         'time' => DateTime_unix(time),
         'local_time' => $tz_remote ? DateTime_tz( time, $tz_remote ) : DateTime_unix(time),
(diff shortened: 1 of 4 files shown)


kareila:
Further testing shows that the time shown on the entry is whatever the poster set, not the local time of the viewer.  So while the current behavior is incorrect, this fix is not correct either - we need to either get the poster's timezone, or convert the time to the viewer's timezone.

kareila:
(Or third option: actually convert the time to UTC as advertised, but I doubt that's the desired behavior)

rahaeli:
Viewer's preferred timezone falling back to poster's timezone if viewer has not set one, plz!

PR #1238, opened 2015-02-18
[#1222] don't display 'UTC' for entry times with unknown timezone

### LINKED ISSUE #1222: Site skinned comment pages claim UTC at all times, even when inappropriate

kareila:
This removes the hardcoded 'UTC' display from CSS and from the
function LJ::S2::sitescheme_secs_to_iso (used to display the
entry time when previewing a comment).  The entry time is
not associated with a timezone, being arbitrarily chosen by the poster.

Fixes #1222.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2647,6 +2647,8 @@ sub sitescheme_secs_to_iso {
     my $has_tz = defined $s2_datetime ? "(local)" : "UTC";
     # if timezone execution failed, use GMT
     $s2_datetime = DateTime_unix( $secs ) unless defined $s2_datetime;
+    # but don't display timezone unless it was requested
+    $has_tz = '' unless $opts{tz};
 
     my @s2_args = ( $s2_ctx, $s2_datetime );
 
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2642,13 +2642,18 @@ sub sitescheme_secs_to_iso {
 
     # convert date to S2 object
     my $s2_ctx = [];  # fake S2 context object
+
+    my $s2_datetime;
+    my $has_tz = '';  # don't display timezone unless requested below
+
     # if opts has a true tz key, get the remote user's timezone if possible
-    my $s2_datetime = $opts{tz} ? DateTime_tz( $secs, $remote ) : undef;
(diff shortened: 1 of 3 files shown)


cgi-bin/LJ/S2.pm
```diff
@@ -2645,8 +2645,11 @@ sub sitescheme_secs_to_iso {
     # if opts has a true tz key, get the remote user's timezone if possible
     my $s2_datetime = $opts{tz} ? DateTime_tz( $secs, $remote ) : undef;
     my $has_tz = defined $s2_datetime ? "(local)" : "UTC";
-    # if timezone execution failed, use GMT
-    $s2_datetime = DateTime_unix( $secs ) unless defined $s2_datetime;
+    # if timezone execution failed, use GMT (but don't display timezone)
```
afuna: When $s2_datetime is not defined, we first $has_tz to "UTC", and then to "". We  probably just set it to "" in the line above.

afuna:
The css changes look good!

cgi-bin/LJ/S2.pm
kareila: Oh, I thought the first assignment to $has_tz was only true if $opts{tz} was true... I was thinking that in the undefined timezone case, it should print UTC if $opts{tz} was true, and nothing if it was false.

kareila: I don't know if that made any sense... I'll submit a new commit that hopefully clarifies things.

cgi-bin/LJ/S2.pm
```diff
@@ -2647,6 +2647,8 @@ sub sitescheme_secs_to_iso {
     my $has_tz = defined $s2_datetime ? "(local)" : "UTC";
     # if timezone execution failed, use GMT
     $s2_datetime = DateTime_unix( $secs ) unless defined $s2_datetime;
+    # but don't display timezone unless it was requested
+    $has_tz = '' unless $opts{tz};
```
afuna: Hrrrm. Better. I can see what it's trying to do now. But how about this instead?

    my $has_tz = '';
    if ( $opts{tz} ) {
        $has_tz = defined $s2_datetime ? ...
    }

kareila:
I updated this yesterday but forgot to leave a comment!

afuna:
Thank you!

PR #1240, opened 2015-02-21
[#1213] Rewrites http:// video embeds to use protocol relative URLs

### LINKED ISSUE #1213: embedded iframes: mixed content issue
When a video / embed uses http instead of https, we'll run into mixed content issues.

So thoughts:

* more popular sites which we know support http, we can rewrite from http -> https 
* less popular sites, placeholder? just live with the mixed content?

We could also try to proxy the videos, but videos are a much bigger space / bandwidth commitment than images so I'm hesitant to start with that!

afuna:
* ... only if they support them (all the big sites do, smaller ones may
  or may not). We don't want to force to https if it isn't supported
  because the behavior is unreliable.

  An http embed on https may throw up a browser warning.

  A non-working https embed may, e.g., show a cryptic error message
  about internal server error, a blank page, or (in one case) an
  unrelated internal service

* added arguments so that we only change the URL if we're going to
  display the embed (as opposed to editing the embed, or cleaning for
  saving in the database).

Fixes #1213.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -433,9 +433,12 @@ sub clean
 
             # force this specific instance of the tag to be allowed (for conditional)
             my $force_allow = 0;
+
             if (defined $action{$tag} and $action{$tag} eq "conditional") {
                 if ( $tag eq "iframe" ) {
-                    $force_allow = LJ::Hooks::run_hook( 'allow_iframe_embeds', $attr->{src} );
+                    my $can_https;
+                    ( $force_allow, $can_https ) = LJ::Hooks::run_hook( 'allow_iframe_embeds', $attr->{src} );
+                    $attr->{src} =~ s!^https?:!! if $opts->{force_https_embed} && $can_https;  # convert to protocol-relative URL
                     unless ( $force_allow ) {
                         ## eat this tag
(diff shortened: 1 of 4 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -433,9 +433,12 @@ sub clean
 
             # force this specific instance of the tag to be allowed (for conditional)
             my $force_allow = 0;
+
             if (defined $action{$tag} and $action{$tag} eq "conditional") {
                 if ( $tag eq "iframe" ) {
-                    $force_allow = LJ::Hooks::run_hook( 'allow_iframe_embeds', $attr->{src} );
+                    my $can_https;
+                    ( $force_allow, $can_https ) = LJ::Hooks::run_hook( 'allow_iframe_embeds', $attr->{src} );
+                    $attr->{src} =~ s!^https?:!! if $opts->{force_https_embed};  # convert to protocol-relative URL
```
zorkian: Doesn't this need to respect `can_https` somehow?

afuna: Oops, yes.

Pushed an update. Also fixed the return value (of can_https) for those
embeds that needed special handling in EmbedWhitelist (e.g., wikimedia)

Issue #1185, opened 2015-02-04
Bad link on /admin/console/index


kareila:
If you visit http://www.dreamwidth.org/admin/console/index and click the command reference link, it tries and fails to load http://www.dreamwidth.org/admin/console/indexreference instead of http://www.dreamwidth.org/admin/console/reference.  The link is correct on http://www.dreamwidth.org/admin/console/ (no explicit index).

kareila:
This didn't get auto-closed.

PR #1242, opened 2015-02-23
(Issue #1185) - Fixes bad link on /admin/console/index

hotlevel4:
Fixes issue #1185 by hardcoding the URL. If there's a design reason this shouldn't be done, please let me know.


afuna:
No, that was mostly thinking "what if admin/console" changes as the subdirectory? But that's  really more of a what-if than anything, and it caused bugs like the one you just fixed.

Issue #1093, opened 2014-12-06
further distinguish the mark all read/delete all notifs buttons in inbox when filtered by entry


rahaeli:
Shorter version of a longer rambling explanation I gave in irc, attempting to put into words instead of just HANDWAVE HANDWAVE:

When in the inbox, if you're in the 'All' view, the buttons over the notifications read "Mark all read" and "Delete all". If you're in a subfolder, any subfolder at all, they read "mark subfolder read" and "delete all in subfolder".

However, there's no way to distinguish, just by the buttons, if you're in the main "Entries and Comments" notifications filter, or if you're further subfiltered to a single entry's notifications. This has bitten me in the ass before, since I use my inbox as a "keep notification this until you reply to it" system a lot, and so I use the "filter to just this entry" view and the "delete all" button to clear out the notifs from the entries I don't need to keep the notifs from. (That way I can clear out, eg, the 100 notifs from the weekly pride thread all at once, while keeping the ones from other entries that I haven't replied to yet.)

It would be lovely if the buttons on top stayed "Mark all read"/"Delete all" when in the 'All' view, "mark subfolder read/delete all in subfolder" when in one of the main (sidebar) filters like Entries and Comments, Circle Updates, Poll Votes, etc, and switch to "Mark this entry's notifications read" and "delete this entry's notifications" when in single-entry view.

kareila:
This didn't get auto-closed.

PR #1244, opened 2015-02-24
(Bug #1093) Distinguish mark all read/delete all buttons in inbox when f...

hotlevel4:
...iltered by entry

Changes the button text to say "Mark All Read for this Entry" and
"Delete All for this Entry". Also widens button by 3em to accommodate
the longer string of text. Fixes issue #1093.


zorkian:
Thank you! :+1:

PR #1245, opened 2015-02-25
[#1007] Use https:// for system-generated links in emails we send

### LINKED ISSUE #1007: Switch emails to use https instead of http for internal links
- comment notifications 
- support emails, etc

(Might be possible to do this by setting the $LJ::SITEROOT, etc, variables to the $SSL* versions while in sendmail -- but will need testing, etc)

afuna:
* ESN notifications now always use https

* paidstatus worker now always uses https (needs setting in worker
  because this uses send_mail directly)

* emails sent from various pages (e.g., password reset): link protocol
  depends on whether the viewed page uses http or https. This should be
  a non-issue once we turn on https everywhere

* refactors the setting of site variables (to start with http or https)
  because it's now used in multiple places

* renames $LJ::USE_SSL_EVERYWHERE to $LJ::USE_HTTPS_EVERYWHERE as a more
  descriptive variable name

Fixes #1007.

### DIFF
--- a/cgi-bin/LJ/NotificationMethod/Email.pm
+++ b/cgi-bin/LJ/NotificationMethod/Email.pm
@@ -69,6 +69,14 @@ sub notify {
     croak "'notify' is an object method"
         unless ref $self eq __PACKAGE__;
 
+    # use https:// for system-generated links in notification emails
+    local $LJ::IS_SSL;
+    local ${$LJ::{$_}} foreach LJ::site_variables_list();
+    if ( $LJ::USE_HTTPS_EVERYWHERE ) {
+        $LJ::IS_SSL = 1;
+        LJ::use_ssl_site_variables();
+    }
+
     my $u = $self->u;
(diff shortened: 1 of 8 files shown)


cgi-bin/LJ/NotificationMethod/Email.pm
```diff
@@ -69,6 +69,14 @@ sub notify {
     croak "'notify' is an object method"
         unless ref $self eq __PACKAGE__;
 
+    # use https:// for system-generated links in notification emails
+    local $LJ::IS_SSL;
```
zorkian: Nice use of `local`.

zorkian:
This seems OK, so I'll merge it; I think of all the patches this one I'm the most worried about, so we'll have to be careful and test it.

afuna:
Okay! I've tested manually that emails, etc, work and we get the expected results with email (whether they're triggered by an ESN system event or by using send_mail from a page) but more is always helpful.

Issue #1246, opened 2015-02-25
audit all uses of "http:" 


afuna:
Example: redirecting to another page if the URL does not check out (where we mistakenly redirect because we're only checking for "http://" instead of "https?://"

afuna:
Otherwise we get into hilarious cases like trying to access /profile doing an infinite redirect loop.

rahaeli:
ahahahaha that sounds like the tiny despairing voice of "....but why is that HAPPENING?" speaking

afuna:
Noting here that I've also looked at the $*font_sources (which have as values external stylesheet urls) and these are all fine because they go through the cssproxy.

PR #1247, opened 2015-02-26
[#1246] Check for https as well as http; use https for generated urls

afuna:


bin/logsummarize.pl
```diff
@@ -96,7 +96,7 @@ BEGIN
 	}
 
 	my $ref = $r->[F_REF];
-	if ($ref =~ m!^http://([^/]+)!) {
+	if ($ref =~ m!^https?://([^/]+)!) {
```
zorkian: Just delete this whole file, wtf is it

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -499,7 +500,7 @@ sub trans
         # -- uppercase usernames
         # -- users with hyphens/underscores, except users from external domains (see table 'domains')
         if ( $orig_user ne lc($orig_user) ||
-            $orig_user =~ /[_-]/ && $u && $u->journal_base !~ m!^http://$host!i && $opts->{'vhost'} !~ /^other:/) {
+            $orig_user =~ /[_-]/ && $u && $u->journal_base !~ m!^https?://$host!i && $opts->{'vhost'} !~ /^other:/) {
```
zorkian: Why not use `$protocol` here?

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -883,7 +884,7 @@ sub trans
             # redirect them to their canonical URL if on wrong host/prefix
             if (my $u = LJ::load_user($user)) {
                 my $canon_url = $u->journal_base;
-                unless ($canon_url =~ m!^http://$host!i || $LJ::DEBUG{'user_vhosts_no_wronghost_redirect'}) {
+                unless ($canon_url =~ m!^https?://$host!i || $LJ::DEBUG{'user_vhosts_no_wronghost_redirect'}) {
```
zorkian: Why no `$protocol`?

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -1602,7 +1603,7 @@ sub mogile_fetch {
             LJ::MemCache::add( $memkey, $paths, $cache_for ) if @paths;
         }
 
-        if ( defined $paths->[0] && $paths->[0] =~ m/^http:/ ) {
+        if ( defined $paths->[0] && $paths->[0] =~ m/^https?:/ ) {
```
zorkian: Why no...

afuna: (I see what you did there ;))

Can't use protocol here; it's in a different function.

cgi-bin/Apache/LiveJournal.pm
afuna: Hmm, I was concerned about making sure we allow both http and https, but $u->journal_base is already going to be either http or https based on the page it's in so that was actually an invalid concern.

afuna:
Updated with requested changes.

zorkian:
Murdered code makes me so happy!

PR #1249, opened 2015-02-26
[#1248] Make sure redirects work for every page

### LINKED ISSUE #1248: Make sure all pages redirect
Journals still don't (we have so many code paths to show pages...)

afuna:
Fixes #1248.


afuna:
Something will undoubtedly prove me wrong, but as far as I can tell, that makes http => https redirects work for everything.

zorkian:
That's the way of it ;)

PR #1254, opened 2015-02-26
Use https for system-generated links when appropriate

afuna:
This takes care of links in the control strip, and in quick reply.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -1122,7 +1122,7 @@ sub create_url {
     $path ||= $r->uri;
 
     # Default SSL if SSL is set and we are on the same host, unless we explicitly don't want it
-    $opts{ssl} = $LJ::IS_SSL unless $opts{host} || exists $opts{ssl};
+    $opts{ssl} //= $LJ::IS_SSL;
 
     my $proto = $opts{proto} // ( $opts{ssl} ? "https" : "http" );
     my $url = $proto . "://$host$path";


cgi-bin/LJ/Web.pm
```diff
@@ -886,7 +886,7 @@ sub create_qr_div {
     }
 
     return DW::Template->template_string( 'journal/quickreply.tt', {
-        form_url                => LJ::create_url( '/talkpost_do', host => $LJ::DOMAIN_WEB ),
```
zorkian: Why are we passing `$LJ::IS_SSL` in to many places? Why can't these functions just assume it?

Inside, I would imagine `$ssl = $opts{ssl} // $LJ::IS_SSL;` that way they default to whatever the global variable is, but you can still force set it to 0/1.

afuna: This is for the use-case when we’re passing in the host option — and
since we can’t guarantee that the new host supports ssl, we want to pass
in the ssl flag.

In most other cases, create_url is smart enough to do things as you
described.

zorkian: I don't get it. I don't think there is a use case for passing `$LJ::IS_SSL` as a value here.

I think that the `ssl` option should default to `$LJ::IS_SSL`. Then, anybody who _does not_ want to use that value would have to pass it in. Would that really break anything?

afuna:
pushed with changes.

Issue #738, opened 2014-06-10
Indicate subject lines in mod queue are truncated by adding "..." back to end


rahaeli:
Tiny nitpicky little thing: the new version of the moderation queue cuts off entry subjects at what looks like 30 characters. Between that and the 1em font size (which will look really large for people used to Tropo's former .75em) on the new Foundation pages, I'm finding it hard to get a sense of what the entry is. Could we up that to 50 characters, maybe?

kareila:
This didn't get auto-closed.

PR #1257, opened 2015-02-27
(Issue #738) Add "..." to end of truncated moderation queue subjects

hotlevel4:
Checks if subject line is 29 characters or greater (since if last character
is a space it's not really stored in the SQL database) and adds "..." to the
end of the subject line in the moderation queue if so. Will be incorrect for
the edge cases where the subject line is exactly 29-30 characters.

Fixes issue #738.


zorkian:
Deferring to @afuna b/c this touches UI design.

afuna:
Hmm, from that screenshot, it looks like there's some CSS missing.

Could you try `compass compile` in both dw-free and dw-nonfree? 

It should look more like this:

screenshot 2015-02-26 23 14 11

hotlevel4:
Dur. Okay, this should be good. Sorry about that!

afuna:
Thanks!

PR #1259, opened 2015-02-27
Fix embed whitelist tests

afuna:
The hook was updated to return a list, the second element of which could
be false. This made the test fail. So make sure we only check the first.


zorkian:
You have extra files in this push.

afuna:
fixt

Issue #1046, opened 2014-11-06
Convert BML files under htdocs/support


afuna:

hotlevel4:
Trying to convert support/faq.

kareila:
Hi @hotlevel4, are you interested in converting any more of these?  If not, let me know so I can put this back up for grabs!

kareila:
lol never mind :stuck_out_tongue_winking_eye:

hotlevel4:
I am pretty sporadic in my work on this, though, so if it's easier for you to unassign and see if anybody else wants to pick it up and complete it in a more timely fashion, I'm happy to pop in and just add things if/when I have time.

kareila:
There's no rush to get it done, though, so as long as you're interested in continuing to work on it, feel free to hang onto it!  There are plenty of unconverted pages to go around...

rahaeli:
Yes, please never apologize for the speed at which you are moving! We just have a tendency for people to think "I don't really want to do this anymore, but I don't want to drop the issue because then I am LETTING PEOPLE DOWN" and we like to make sure people know that it's okay to throw things back in the pool. :)

PR #1263, opened 2015-02-28
(Issue #1046) Converts support/faq.bml to TT

hotlevel4:
Ignores calls to BML::note_mod_time and BML::get_language, which will need
to be added in the future should other languages become available. Attempts to 
use foundation. Related to issue #1046.


cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+
+DW::Routing->register_string( '/support/faq', \&faq_handler, app => 1 );
+
+sub faq_handler {
+    my $r = DW::Request->get;
+
+    my ( $ok, $rv ) = controller( anonymous => 1, form_auth => 1 );
+    return $rv unless $ok;
+
+    my $remote = LJ::get_remote();
```
afuna: You can also get the remote from $rv (that is, $rv->{remote})

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+    my %faqq;
+    my $ret = "";    
+    
+    $sth = $dbr->prepare("SELECT faqcat, faqcatname, catorder FROM faqcat ".
+                         "WHERE faqcat<>'int-abuse'");
+
+    $sth->execute;                         
+                         
+    while ($_ = $sth->fetchrow_hashref) {
+        $faqcat{$_->{'faqcat'}} = $_;
```
afuna: no quotes please

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+
+    my $vars = {};
+    
+    my $dbr = LJ::get_db_reader(); 
+    my $sth;
+    my %faqcat; 
+    my %faqq;
+    my $ret = "";    
+    
+    $sth = $dbr->prepare("SELECT faqcat, faqcatname, catorder FROM faqcat ".
```
afuna: Please add whitespace inside parentheses -- I know that these are copy-pasted, but we might as well update the code style as we convert!

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Controller::Support::Faq;
+
+use strict;
+use warnings;
+use vars qw(%GET);
```
afuna: I didn't realize this would work!

I'd encourage doing it this way though:

    $r = DW::Request->get;
    my $get = $r->get_args;
    ... = $get->{...}

cgi-bin/DW/Controller/Support/Faq.pm
afuna: Note: not going to go over all instances (and I did notice that there are some with the added whitespace already!) but please take another look to clean up the remaining :)

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+    foreach my $faqcat ( sort { $faqcat{$a}->{catorder} <=>
+                                $faqcat{$b}->{catorder} } keys %faqcat )
+    {
+        next if $GET{cat} && $GET{cat} ne $faqcat;
+        my $countfaqs = 0;
+        foreach ( grep {$faqq{$_}->faqcat eq $faqcat } keys %faqq) {
+            $countfaqs++;
+        }
+        next unless $countfaqs;
+        my $faqcatname = LJ::ehtml( $faqcat{$faqcat}->{faqcatname} );
```
afuna: Hmmm. So I'd suggest not doing the ehtml here, but instead doing that in the template.

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -0,0 +1,103 @@
...
+        push @{ $vars->{faqcats} }, {
+            faqcat => $faqcat, 
+            faqcatname => $faqcatname,
+        };          
+        foreach my $faqid ( sort { $faqq{$a}->sortorder <=> $faqq{$b}->sortorder } grep { $faqq{$_}->faqcat eq $faqcat } keys %faqq )
```
afuna: Nice cleanup of the (previously) 2x loop.

afuna:
BML::note_mod_time doesn't have to do with language; it takes the modification time of the FAQ and adds it to the headers so we can return a 304 Not Modified if relevant.

It does not appear we actually *use* it, because we never pass in the appropriate options. Could you open up a bug to investigate further BML::note_mod_time in /support/faq though?

rahaeli:
I'll do it!

cgi-bin/DW/Controller/Support/Faq.pm
afuna: On second thought, I think that this doesn't actually do anything. It initializes the %GET variable but doesn't actually map it to the get arguments (the way that BML does).

You can do that by following the example code I listed above.

*However* looking at the code, this is pretty much a duplicate of the faqbrowse functionality, so I'd rather just remove it entirely.

afuna:
Thank you!

views/support/faq.tt
```diff
@@ -0,0 +1,56 @@
...
+
+<p>[% '.title.text2' | ml( sitenameshort = site.nameshort ) %]</p>
+<p>[% '.search.text' | ml %]</p>
+
+<form method='GET' action='[% site.root %]/support/faqsearch'>
+<div class="row">
+  <div class="large-12 columns">
+    <div class="row collapse">
+      <div class="small-10 columns">
+        <input type="text" placeholder="Search terms" name="q">
```
afuna: Nice. I'd suggest using "type=search" here for semantics (won't affect appearance, just behavior in e.g., screenreaders)

views/support/faq.tt
```diff
@@ -0,0 +1,56 @@
...
+[% FOREACH faqcategory = faqcats %]
+    <h2>[% faqcategory.faqcatname %] (<a href='faqbrowse?faqcat=[% faqcategory.faqcat %]' name='[% faqcategory.faqcat %]'>[% '.view.all' | ml %]</a>)</h2>
+    <ol class='faqlist'>
+    [% faqcatid = faqcategory.faqcat %]
+    [% FOREACH faqq = $faqcatid.faqqs %]
+        <li><a href='faqbrowse?faqid=[% faqq.faqid %]'>
+            [% faqq.q %]
+        </a></li>
+    [% END %]
+    </ol><hr>
```
afuna: Hmmm. I think I prefer this without the <hr>, for cleanliness. The headers are bold enough that they work to demarcate the next section clearly.

views/support/faq.tt
```diff
@@ -0,0 +1,56 @@
...
+
+<h2>[% '.toc.title' | ml %]</h2>
+<ol>
+    [% FOREACH faqcategory = faqcats %]
+        <li><a href = #[% faqcategory.faqcat %]>[% faqcategory.faqcatname %]</a></li>
+    [% END %]
+</ol><hr>
+
+[% FOREACH faqcategory = faqcats %]
+    <h2>[% faqcategory.faqcatname %] (<a href='faqbrowse?faqcat=[% faqcategory.faqcat %]' name='[% faqcategory.faqcat %]'>[% '.view.all' | ml %]</a>)</h2>
```
afuna: we can do the ehtml here using the `| html` filter (since I suggested we remove it from the controller)

views/support/faq.tt
hotlevel4: Shoot, had that in the wrong place. Meant to put it above the backlink.

hotlevel4:
Should be fixed, I think!

afuna:
Thanks!

Issue #1264, opened 2015-03-01
Move t/config-test.pl to ext/local/etc


pauamma:
The checked-in version isn't appropriate for dreamhacks and needs to be modified without polluting diffs. The easiest fix is to move it alongside other config files.

afuna:
Had a sudden thought. Would ext/local/t work to override?

pauamma:
Maybe? But I can't help thinking that having an overridable-configuration for tests would be more than we need?

afuna:
Hmmm but ext/local/etc is not under our control, in the same way that ext/local/t wouldn't be so I'm not sure there's much difference here.

Sophira:
I'm sorry; this whole thing is basically my fault. I said in #931 that I would create test databases for Dreamhack users to use and I haven't done that yet. I've been meaning to do so.

I can get that sorted in the next few days, if that helps any!

afuna:
No that's fine, it's something we have to figure out anyway!

Is there anything we need to do to make it flexible? Do you want separate test dbs for each 'hack user? 

Maybe a t/config-private which only covers the database credentials so we don't need to override the whole file?

PR #1265, opened 2015-03-01
Add aliases with dashes to email_aliases

### LINKED ISSUE #1080: treat - and _ interchangeably in forwarding email addresses
user_name@dreamwidth.org works for the forwarding email address, but user-name@dreamwidth.org doesn't. People don't really think of their usernames as being user_name anymore, since they see it up in the URL bar as user-name, and everything else on the site treats them as interchangeable, so we should also make them interchangeable in forwarding email addresses.

pauamma:
Also refactor a bit

Fixes #1080


cgi-bin/LJ/Console/Command/EmailAlias.pm
```diff
@@ -68,12 +71,28 @@ sub execute {
         $dbh->do("REPLACE INTO email_aliases VALUES (?, ?)", undef, $alias, $value);
         return $self->error("Database error: " . $dbh->errstr)
             if $dbh->err;
+        if ( $alias ne $alias_with_dashes ) {
+            $dbh->do( "REPLACE INTO email_aliases VALUES (?, ?)",
```
afuna: Hmm, a thought, if a user renames, and goes from, e.g, `abc_123` to `abc`, they'll have previously had two entries in the db:

abc-123 => email@example.com
abc_123 => email@example.com

On a rename, they only have "abc". Does that then let the old "abc-123" alias hang around?

Note: this isn't an issue with renames because that does do a delete_email_alias if we're not redirecting emails.

cgi-bin/LJ/Console/Command/EmailAlias.pm
```diff
@@ -45,7 +45,10 @@ sub execute {
 
     # canonicalize
     $alias =~ s/\@.*//;
+    my $alias_with_dashes = $alias;
```
afuna: Why not use the new method you created in LJ::User?

cgi-bin/LJ/User.pm
```diff
@@ -360,7 +360,7 @@ sub delete_and_purge_completely {
 
     $dbh->do("DELETE FROM wt_edges WHERE from_userid = ? OR to_userid = ?", undef, $u->id, $u->id);
     $dbh->do("DELETE FROM reluser WHERE targetid=?", undef, $u->id);
-    $dbh->do("DELETE FROM email_aliases WHERE alias=?", undef, $u->site_email_alias);
+    $u->delete_email_alias;
```
afuna: Good call on creating this function. Can you use it in the console command as well?

cgi-bin/LJ/User.pm
```diff
@@ -4949,6 +4958,12 @@ sub update_email_alias {
               undef, $u->site_email_alias, $u->email_raw );
 
     return 0 if $dbh->err;
+
+    $dbh->do( "REPLACE INTO email_aliases (alias, rcpt) VALUES (?,?)",
```
afuna: Same question here about renames.

t/aliases.t
```diff
@@ -0,0 +1,50 @@
...
+    my ( $rcpt ) = $dbh->selectrow_array( "SELECT rcpt FROM email_aliases WHERE alias=?", undef, $alias );
+    return defined $rcpt;
+}
+
+# Now you see them
+ok( $u->update_email_alias, "update_email_alias successful" );
+ok( check_alias( $u->site_email_alias ), "alias present" );
+ok( check_alias( $u->site_email_alias_with_dashes ), "alias with dashes present" );
+
+# Now you don't
```
afuna: hah

cgi-bin/LJ/Console/Command/EmailAlias.pm
pauamma: Not sure I understand your question right, but the email_alias command alias argument doesn't have to be a username (current or redirected), so in that case, you could

> email_alias delete abc_123

and or

> email_alias delete abc-123

depending on which alias(es) remained hanging around.

My reasoning is that whoever uses that command should know what they're doing. Does that answer your question?

cgi-bin/LJ/Console/Command/EmailAlias.pm
pauamma: Because the alias given in the command may not correspond to a user object at all.

cgi-bin/LJ/Console/Command/EmailAlias.pm
afuna: Hmmm. I can guarantee you that this will be forgotten by the next time anyone uses this tool but I see your point that this is mostly a manual tool and shouldn't be too clever. 

Can we be a little clever though and present useful information reminding the tool user about underscores etc?

alierak: As I understand it, you'd still have both abc-123 and abc_123 aliases in the table after renaming abc_123 to abc, unless something else caused them to be deleted. The update_email_alias call would only affect rows where the alias (first column) is 'abc'.

rahaeli: Yeah, I'm never going to remember that in a million years.

cgi-bin/LJ/User.pm
pauamma: No, because there may be no user object corresponding to the alias entered there.

cgi-bin/LJ/User.pm
pauamma: If I understand you right this isn't a new problem introduced by this change, because it was true of abc_123 even before. Perhaps this should be another bug?

cgi-bin/LJ/Console/Command/EmailAlias.pm
pauamma: OK, what should the reminder say and when? In the help text for the command, when the alias contains a _, when the alias contains a -, when the alias contains neither?

afuna: I'm actually reconsidering this approach (sorry!) and thinking of going back to your original approach of modifying the select for the postfix.

The main problem with that is testing,  but it also means that we don't have to keep multiple things updated.

pauamma: Let me know when you decide either way. (FWIW, I think tackling this in postfix configuration is cleaner.)

zorkian:
Hi! Sorry for being slow here -- two points of feedback.

1. Whatever we put in the database should be exactly the username. Usernames have underscores. If we want to build some way of allowing the system to accept hyphens, that's great, but I don't think the correct approach would be to put hyphens in the database.

2. Updating the SQL so it basically converts the email being sent to down to underscores to match what we put in the database makes the most sense, IMO. That'd be pretty easy to do operationally assuming #1 is true and we can predict what alias would be in the DB.

Make sense? Thank you for working on this!

afuna:
@zorkian that makes sense to me. Have you seen https://github.com/dreamwidth/dw-free/issues/1080#issuecomment-64789967 ?

zorkian:
Yes I have. That's what I said in my second point: basically we should use the SQL that @pauamma tested and proved and then that should cover all cases.

Re: the comment on the issue (please try keep all comments about code on the pull requests!), I can definitely make the SQL change for postfix. Wouldn't some of the code here be useful, though? Any of the test stuff? Or does it not make sense now?

afuna:
My initial thought was no, but you know, you're right about the tests. I don't think any of the code changes are applicable, but we can at least run tests against the SQL that is suggested for the postfix config.

pauamma:
Also maybe the refactoring?

afuna:
Sure, less duplication would be nice!

pauamma:
OK, here's my part of it.

t/aliases.t
```diff
@@ -0,0 +1,52 @@
...
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+use strict;
+use warnings;
+
+use Test::More tests => 6;
+
+use lib "$ENV{LJHOME}/cgi-bin";
+# FIXME: use $LJ::_T_CONFIG = 1 when test config unbroken on dreamhacks
```
afuna: With the changes in #1264, can we remove this flag?

pauamma:
Forgot to comment then, but I removed the FIXME and temporary setting

zorkian:
Thanks for doing this :)

PR #1269, opened 2015-03-02
[#1154] Add quick reply to the Database layout

### LINKED ISSUE #1154: Quicker Reply not working in Database
Database has custom code for the reply link because it uses an image instead of text. Doesn't seem like it was updated to work with Quicker Reply. This may be true of other styles but I don't think there are others doing this. Worth checking, though.

afuna:
Fixes #1154.


zorkian:
Seems legit.

PR #1271, opened 2015-03-03
[#1005] Switch doctype to use HTML5

### LINKED ISSUE #1005: switch doctype to HTML5 instead of XHTML
- site pages 
- core2 / layouts

afuna:
Fixes #1005.

### DIFF
--- a/styles/zesty/layout.s2
+++ b/styles/zesty/layout.s2
@@ -1775,9 +1775,8 @@ function Page::lay_footer() {
 function Page::print() {
 
 """
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
-        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-loose.dtd">
-       <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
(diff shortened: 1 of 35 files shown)


styles/zesty/layout.s2
```diff
@@ -1775,9 +1775,8 @@ function Page::lay_footer() {
 function Page::print() {
 
 """
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
-        "http://www.w3.org/TR/xhtml1/DTD/xhtml1-loose.dtd">
-       <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
```
zorkian: lol zesty.

zorkian:
This doesn't change rendering or anything? I'd be concerned that this would affect what browsers do to us :)

afuna:
Nope, it affects validation mostly.

Technically — XHTML is stricter than HTML5, so for a page with broken
HTML, something with the XHTML doctype may just not render the page
fully. Also, technically, some of the things we do for accessibility are
invalid XHTML (but valid HTML5). But in practice, browsers will just
handle it.

PR #1273, opened 2015-03-04
[#1113] Show the read comments link when you reply via QR

### LINKED ISSUE #1113: make sure to show "1 comment" when we post inline from reading page to an entry which had 0 comments
Right now we just regex replace the existing link, and since when there's no comments there's no link yet, this doesn't work when going from 0 -> 1 comment. We should be smarter about that.

afuna:
-- this is for the case where we go from 0 -> 1 comment, and we used to
not display a link (because we had nothing to display)

Fixes #1113.


afuna:
What. Sorry my pull request script bugged out on me.

PR #1275, opened 2015-03-04
[#953] Add labels when there are multiple navigation landmarks

### LINKED ISSUE #953: Label the navigation landmarks
We should add add a label (either using aria-label or aria-labelledby) to the role=navigation landmarks because we have multiple of those on a page

I think it's not *strictly* required because the [examples](http://www.w3.org/WAI/GL/wiki/Using_ARIA_landmarks_to_identify_regions_of_a_page) say "more than two", but it'd be nice to be clear.

afuna:
Fixes #953.

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -3767,6 +3767,10 @@ sitescheme.footer.sitemap=Site Map
 
 sitescheme.footer.suggestion=Make a Suggestion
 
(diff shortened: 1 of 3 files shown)


zorkian:
This issue has extra commits on it. Closing. Please fix and reopen.

afuna:
Reduced to one.

PS there's a corresponding dw-nonfree patch btw.

PR #1278, opened 2015-03-04
[#1095] Accept passed-in remote for watch_items

### LINKED ISSUE #1095: API 'getreadpage' function indirectly (and incorrectly) uses LJ::get_remote()
It appears that when we implemented the `getreadpage` function in our API, we used the `watch_items` method from `DW::Logic::LogItems`. However, that method uses `LJ::get_remote()` in order to determine the remote user, rather than allowing the protocol to pass through its own authenticated remote user.

This is a problem because `LJ::get_remote()` relies on a session object built using the cookies of the HTTP request, which can legitimately be non-existent when using the API challenge/authenticate system. As a result, `watch_items` only returns public posts, even if the user has successfully authenticated via the protocol.

This issue seems to have existed ever since commit 3d38a7af on 2011-12-29, when we implemented the new `getreadpage` function. It appears that `watch_items` was only built with the Web interface in mind.

I *believe* the issue can be worked around by using the `ljsession` cookie returned by the `sessiongenerate` API call to authenticate instead of using the challenge-response system, but I haven't tested this.

The fix should probably be to modify `watch_items` to allow the remote user to be passed from the protocol, and then to modify the `getreadpage` function to do this.

afuna:
This is primarily useful for getreadpage from the protocol, where
LJ::get_remote() does not work.

Fixes #1095.

### DIFF
--- a/cgi-bin/DW/Logic/LogItems.pm
+++ b/cgi-bin/DW/Logic/LogItems.pm
@@ -52,7 +52,7 @@ sub watch_items
 
     # we only allow viewing protected content on your own reading page, if you
     # are viewing someone else's reading page, we assume you're logged out
-    my $remote = LJ::get_remote();
+    my $remote = LJ::want_user( delete $args{remote} ) || LJ::get_remote();
     $remote = undef if $remote && $remote->id != $u->id;
 
     # prepare some variables for later... many variables


zorkian:
Wait a second, how is this safe? If someone just passes in `?remote=afuna` that would set the `$remote` which seems sketchy.

afuna:
We need to pass in the remote somehow because LJ::get_remote() only works in web context ( LJ::Protocol is definitely not web context), but in all cases where we pass in remote to watch_items, we only ever do so by setting it to the user being viewed and only if we've authenticated the remote user by other means.

Importantly, we don't pass in remote based on user input at all.

zorkian:
Why does `LJ::get_remote()` not work for protocol? We have a remote user, they're making a request, we should be able to make it just work?

It seems that the ideal situation is that we stop plumbing through "passing around the remote user" and instead just have the `get_remote` function Do The Right Thing Everywhere.

That might be too much for this particular pull request though, so I'll merge it; but we should consider this kind of simplification when we patch things.

afuna:
Hmmm, if we wanted we could use `LJ::set_remote()` instead, and
`LJ::get_remote()` will work magically everywhere.

(It won’t work without set_remote because get_remote normally relies on
cookies — and protocol only has cookies if the authentication method is
cookies instead of password auth.)



On Sat, Mar 7, 2015, at 10:27 PM, Mark Smith wrote:
> Why does LJ::get_remote() not work for protocol? We have a
> remote user, they're making a request, we should be able to make
> it just work?


> It seems that the ideal situation is that we stop plumbing through
> "passing around the remote user" and instead just have the get_remote
> function Do The Right Thing Everywhere.


> That might be too much for this particular pull request though, so
> I'll merge it; but we should consider this kind of simplification when
> we patch things.


>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/1278#issuecomment-77735571

PR #1282, opened 2015-03-05
[#997] ?ssl=no turns off redirects from http->https

### LINKED ISSUE #997: Non-HTTPS login page
*not* default. Discoverable via FAQ?

Need to make sure we still have the login js but only if no ssl

afuna:
Fixes #997.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -304,6 +304,8 @@ sub trans
     my $host = lc( $apache_r->headers_in->{"Host"} );
     my $hostport = ( $host =~ s/(:\d+)$// ) ? $1 : "";
 
+    my %GET = LJ::parse_args( $apache_r->args );
+
     # Allow hosts ending in . to work properly.
     $host =~ s/\.$//;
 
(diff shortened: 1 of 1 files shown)


zorkian:
Hrmmm.

This allows you to turn it off on any page, but since we don't automatically add that to URLs, that means every page you have to do it again.

I'm actually not sure that it's worth doing this at all. Allowing one user to bypass security means you're exposing other user content to all the intermediaries. That's kind of shady, so maybe we just say "no, we require modern SSL"?

afuna:
I am fine with this as well.


On Fri, Mar 6, 2015, at 07:00 AM, Mark Smith wrote:
> Hrmmm.


> This allows you to turn it off on any page, but since we don't
> automatically add that to URLs, that means every page you have to do
> it again.


> I'm actually not sure that it's worth doing this at all. Allowing one
> user to bypass security means you're exposing other user content to
> all the intermediaries. That's kind of shady, so maybe we just say
> "no, we require modern SSL"?


>


Links:

  1. https://github.com/dreamwidth/dw-free/pull/1282#issuecomment-77572054

kaberett:
Removing the "requested changes" label seems to have triggered this showing up in the code tour generator -- something to investigate/tweak, maybe?

PR #1286, opened 2015-03-07
move /search to DW::Controller::Search::Journal

### LINKED ISSUE #1284: convert journal search page from BML to TT
This conversion doesn't look difficult, but setting up a working test environment for Sphinx on a dreamhack might be...

kareila:
I tested this by adding `sub _do_search( return $mock_results )` to the end of the controller file in lieu of the actual _do_search which was copied basically verbatim from the old search page, so I don't anticipate any issues that aren't due to kinks in the translation from straight Perl to TT plus Perl.

I did run into an issue where the 'noescape' parameter for LJ::html_check wasn't supported by the equivalent method in DW/Template/Plugin/FormHTML.pm, so I added that in.

Fixes #1284.


cgi-bin/DW/Controller/Search/Journal.pm
```diff
@@ -0,0 +1,185 @@
...
+use DW::Template;
+use DW::FormErrors;
+use Storable;
+
+DW::Routing->register_string( '/search', \&search_handler, app => 1 );
+
+my $gc = LJ::gearman_client();
+
+sub _do_search {
+    return unless $gc && @LJ::SPHINX_SEARCHD;
```
afuna: Nitpick: Think this is redundant given the code in search_handler

cgi-bin/DW/Controller/Search/Journal.pm
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+
+            # the arguments to the search (userid=0 implies global search)
+            my $args = { userid => $su ? $su->id : 0, remoteid => $remote->id,
+                         query => $q, offset => $offset, sort_by => $sby,
+                         ignore_security => $ignore_security, allowmask => $allowmask,
+                         include_comments => $wc };
+
+            my $result = _do_search( Storable::nfreeze( $args ) );
```
afuna: Nitpick: I'd prefer this inside the _do_search, for symmetry / clear association with the Storable::thaw in there.

afuna:
I've tested with sphinx and can confirm that the search looks fine!

cgi-bin/DW/Controller/Search/Journal.pm
kareila: To clarify, you want me to change this to `_do_search( $args )` and have the `nfreeze` happen inside the function?

cgi-bin/DW/Controller/Search/Journal.pm
kareila: You can't be too careful...

cgi-bin/DW/Controller/Search/Journal.pm
afuna: Yup!

kareila:
That should do it :)

PR #1291, opened 2015-03-14
Multiple email addresses per support request:

### LINKED ISSUE #945: track multiple email addresses on emailed-in support requests
Scenario that crops up pretty often in Abuse: Party A emails in a request to abuse@, we reply, they fwd our reply to Party B, Party B uses the replyto link to add another comment to the request, we reply again, the email reply only goes to Party A.

If someone using a different email address than the one specified at creation comments to a request using the auth code and the email gateway, their email address should also receive a notification of any future replies. (This should *only* happen with the email gateway: someone leaving a comment on a request through the webform shouldn't get emailed notifs of future replies.)

pauamma:
- Add From: email address to supportprops if never seen
- Send response notification email to each stored email address in turn

Fixes #945

### DIFF
--- a/cgi-bin/LJ/Support.pm
+++ b/cgi-bin/LJ/Support.pm
@@ -915,7 +915,7 @@ sub add_email_address {
     my ( $sp, $address ) = @_;
 
     # Already present?
-    return if grep { $_ eq $address } @{all_email_addresses( $sp )};
+    return if grep { $_ eq $address } all_email_addresses( $sp );
 
     # Add
     my $props = load_props( $sp->{spid} + 0 );
@@ -938,7 +938,7 @@ sub all_email_addresses {
         push @emails, ( $u->email_raw || $sp->{reqemail} );
     }
 
-    return \@emails;
+    return @emails;
 }
 
 sub mail_response_to_user
@@ -1026,7 +1026,7 @@ sub mail_response_to_user
         $body .= "\n\n" . LJ::Lang::ml( "support.email.update.noreply" );
     }
 
-    foreach my $email ( @{all_email_addresses( $sp )} ) {
+    foreach my $email ( all_email_addresses( $sp ) ) {
         LJ::send_mail( {
             to => $email,
             from => $fromemail,
--- a/cgi-bin/LJ/Support.pm
+++ b/cgi-bin/LJ/Support.pm
@@ -907,6 +907,40 @@ sub touch_request
     return 1;
 }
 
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/Support.pm
```diff
@@ -907,6 +907,40 @@ sub touch_request
...
+sub add_email_address {
+    my ( $sp, $address ) = @_;
+
+    # Already present?
+    return if grep { $_ eq $address } @{all_email_addresses( $sp )};
+
+    # Add
+    my $props = load_props( $sp->{spid} + 0 );
+    my $nb_extra_addresses = $props->{nb_extra_addresses} || 0;
+    set_prop( $sp->{spid}, 'nb_extra_addresses', $nb_extra_addresses + 1 );
```
afuna: Clever!

Now I'm torn here because this does work, but it breaks a sort of hidden expectation of props having definite keys.

*But*, and importantly, from the point of view of *not* losing data this is better than concatenating extra email addresses into one property value.

pauamma: I'm not sure that expectation is warranted in the long term, considering this comment in LJ::Support::set_prop:

> # TODO:
> # -- delete on 'undef' propval
> # -- allow setting of multiple

I also considered putting addresses in a table of their own, but that feels overwrought to me.

And I think that data safety is the overriding bit here, considering the possible legal consequences of not emailing one address you thought you had. (This is for ToS, after all.)

rahaeli: Don't worry too much about that, email is not a guaranteed form of contact in any fashion.

pauamma:
Does that mean we can enforce/expect that total length of email addresses + number of email addresses <= 256?

zorkian:
I'm not sure if @afuna's comments are asking for changes or not.

I do agree that this patch is pretty "clever" which I generally feel is an anti-goal. But I'm not sure how else to make this work easily otherwise without either overdesigning (a separate table) or running into that 256 character limit. So, I guess this is OK? We'll never have the ability to remove an address from a list (code to do that would get more complex and racy).

Also, it seems a bit like you'd want the return of that list function to be an array, not an arrayref. Since you only ever de-reference it immediately, just return it as a list so you reduce sigil usage?

afuna:
Agree with @zorkian's comment above; adding the requested changes tag for the array versus arrayref.

pauamma:
OK, this should take care of the requested change

PR #1293, opened 2015-03-16
Handle redirects from no ending slash -> ending slash

afuna:
This allows us to redirect from https://.../foo -> https://.../foo/,
instead of:

	https://.../foo/ -> http://.../foo/

Or, with $LJ::USE_HTTPS_EVERYWHERE = 1:

	https://../foo/ -> http://.../foo -> https://.../foo/


afuna:
Also did some cleanup where we had URLs foo that just redirect to foo/ -- we should just link to foo/

zorkian:
just approving things

justapprovingthings

PR #1295, opened 2015-03-16
Use indexOf instead of startsWith for browser compatibility

afuna:


kareila:
Someone commented about an hour ago saying the startsWith was still a problem. Is it possible this didn't go live on the site yet, even though it was committed on the release branch?

afuna:
Yeah looks like we didn't -- but now we have. Thanks for the reminder.

PR #1300, opened 2015-03-17
[#1298] Make sure we close the table tag

### LINKED ISSUE #1298: HTML cleaner should autoclose table tags
I don't know why exactly this changed behavior with the most recent push, but we have never automatically closed TD and TR tags when closing a table.  Look at t/cleaner-tables.t:
```
# VALID: table without closing tr/td tags
$orig_post = "<table><tr><td>Cell 1<td>Cell 2<tr><td>Cell 3<td>Cell 4</table>";
$clean->();
ok($orig_post eq $clean_post, "Table okay if td and tr tags aren't closed");
```
All of a sudden this is breaking pages where it wasn't before. We should implement the following rules in the cleaner:

- If a second TD is opened, close the first TD
- If a second TR is opened, close the first TR
- If a table is closed, close any open TDs and TRs

afuna:
Fixes #1298.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -950,7 +950,8 @@ sub clean
                             # so rather than mess with it, let's just ignore those
                             # and only deal with non-self-closing tags
                             # which are not in a table
-                            push @tagstack, $tag unless $slashclose || @tablescope;
+                            # (but we still want to close <table>; that's not yet inside the table)
+                            push @tagstack, $tag if ! $slashclose && ( $tag eq "table" || ! @tablescope );
                         }
                         else { $newdata .= "&gt;"; }
                     }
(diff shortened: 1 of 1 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -950,7 +950,8 @@ sub clean
                             # so rather than mess with it, let's just ignore those
                             # and only deal with non-self-closing tags
                             # which are not in a table
-                            push @tagstack, $tag unless $slashclose || @tablescope;
+                            # (but we still want to close <table>; that's not yet inside the table)
+                            push @tagstack, $tag if ! $slashclose && ( $tag eq "table" || ! @tablescope);
```
zorkian:
```
! @tablescope);
```

should be

```
! @tablescope );
```

afuna:
made changes requested.

PR #1307, opened 2015-03-18
fix HTML tags being mismatched after doing LJ::html_trim

afuna:
Primary problem was that in some cases closing tags are optional (such
as with tr/td) and so if you had input such as the first given test
case:

    <table><tr><td>test</table>

You'd get:

     <table><tr><td>test</table>[</tr></table>]

Note the unnecessary added ending tags in the bracket.

This patch changes html_trim so we only pop / append the ending tag
in-text when it matches the last opened tag in the stack.

Then we close all open tags in the stack at the end.

It also adds a couple test cases that demonstrate the behavior of
htmltrim when HTML tags are mismatched.

### DIFF
--- a/cgi-bin/LJ/TextUtil.pm
+++ b/cgi-bin/LJ/TextUtil.pm
@@ -653,8 +653,10 @@ sub html_trim {
 
         } elsif ($type eq 'E') {
             # end tag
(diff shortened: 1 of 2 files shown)


zorkian:
Nice catch!

PR #1319, opened 2015-03-26
[#1313] Support contextual hover menu on icon urls with old format

### LINKED ISSUE #1313: hovering over icons posted in comments/entries no longer triggers ctxpopup
http://dw-maintenance.dreamwidth.org/67313.html?thread=2389745#cmt2389745

At least, I think that's what they're reporting!

afuna:
Fixes #1313.


afuna:
By necessity, much looser than before (not checking against, e.g., siteroot because we can't guarantee protocols will match)

PR #1321, opened 2015-03-26
[#1303] Trigger the dialog onload not onready

### LINKED ISSUE #1303: investigate report of wandering tagnav element
http://dw-maintenance.dreamwidth.org/67313.html?thread=2378225#cmt2378225 includes screenshots

Browser appears to be Firefox on most recent Mac OS, journal style appears to be Five AM.

afuna:
This ensures that the DOM has been calculated properly by the time we
try to figure out whether to display the tags or not.

This fixes the tag nav showing up inconsistently positioned in
Chrome/Safari.

Fixes #1303.

### DIFF
--- a/htdocs/js/journals/jquery.tag-nav.js
+++ b/htdocs/js/journals/jquery.tag-nav.js
@@ -95,14 +95,18 @@
 
 jQuery(document).ready(function() {
     $(".tag-text[data-journal]").tagnav();
+});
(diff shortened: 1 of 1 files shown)


afuna:
Take two. #1320 fixed one case (the image loading too slowly). This fixes another (DOM not being fully loaded).

Issue #1323, opened 2015-03-28
break LJ/User.pm into submodules


kareila:
LJ/User.pm is by far the largest code module in the repository.  It's almost 10k lines long.  The only other file that's even close is jQuery.
```
% find . -type f -not -path "./.git/*" -not -path "*.jpg" -not -path "*.s2" -print0 | xargs -0 wc -l | sort -r | grep -v ' total$' | head -10

    9866 ./cgi-bin/LJ/User.pm
    9471 ./htdocs/js/jquery/jquery-1.8.3.js
    5328 ./bin/upgrading/en.dat
    4837 ./cgi-bin/LJ/S2.pm
    4179 ./bin/upgrading/update-db-general.pl
    4175 ./cgi-bin/LJ/Protocol.pm
    3923 ./cgi-bin/LJ/Talk.pm
    3829 ./cgi-bin/LJ/Web.pm
    2767 ./cgi-bin/LJ/Entry.pm
    2763 ./bin/moveuclusterd.pl
```
I think it would improve maintainability to break it into submodules.  We could keep the current categorization system to organize the submodules and help people used to the old system to find things. It would look something like this:
```
### Categories:
###  1. Creating and Deleting Accounts
###  2. Statusvis and Account Types
###  3. Working with All Types of Account
use LJ::User::Account;

###  4. Login, Session, and Rename Functions
use LJ::User::Login;

###  5. Database and Memcache Functions
use LJ::User::Data;
```
Etc.

rahaeli:
i am for anything that will let me actually find things in that giant heap of code. (Seriously, I have written new functions in the past only to realize at the very last moment there was one in there that did what I needed it to do after all.)

pauamma:
See also DW::User::*

kareila:
Good point, @pauamma. I will look at that before closing this.

pauamma:
Also, should something similar be done with those files other than cgi-bin/LJ/User that define functions and methods in LJ::User? Like cgi-bin/LJ/Keywords.pm, cgi-bin/LJ/Community.pm, cgi-bin/DW/Mood.pm. and cgi-bin/DW/Shop.pm?

kareila:
I think those are tightly integrated enough that it makes more sense to leave them where they are.

kareila:
[...]

PR #1326, opened 2015-03-30
move user methods and functions into submodules

kareila:
This should probably be merged relatively quickly as any other incoming changes to User.pm will break automatic mergeability on this.  :)

I did some basic testing - ran the compile test, made sure the web server started up, loaded the home page and a journal page.  Everything seemed to work as expected.

The only other refactoring I did on the base commit was to move various "use module" statements to the specific submodule where they were being used, in the cases where I was able to make that determination.

Some of the sections that were moved into the same module weren't sequential in the original catalog order.  If they were close together, I rearranged the numbers to make them sequential.  If they were far apart, I left the original numbers intact.

I removed the `is_minor` user method from the "99 obsolete" section and put it in with the rest of the age-related code, since that method seems to be in pretty heavy use still.

One aspect of the original User.pm I glossed over was the fact that it explicitly turns off uninitialized variable warnings, even in development environments.  When the code was moved into submodules, that pragma was not inherited, and I quickly logged at least a dozen warnings in my server log.  The second commit fixes those warnings and turns the pragma back on for the submodules.

These are the new line counts for User.pm and the submodules:
```
     131 cgi-bin/LJ/User.pm
    1670 cgi-bin/LJ/User/Account.pm
     373 cgi-bin/LJ/User/Administration.pm
     478 cgi-bin/LJ/User/Age.pm
    1241 cgi-bin/LJ/User/Data.pm
     706 cgi-bin/LJ/User/Display.pm
     885 cgi-bin/LJ/User/Icons.pm
     932 cgi-bin/LJ/User/Journal.pm
     583 cgi-bin/LJ/User/Login.pm
     582 cgi-bin/LJ/User/Message.pm
    1617 cgi-bin/LJ/User/Permissions.pm
     925 cgi-bin/LJ/User/Styles.pm
```
There's definitely room for further reorganization and cleanup, but I think this is a good first pass and achieves the goal of breaking down the monolith into chunks that are easier to work with and debug.


kareila:
I forgot to mention that this fixes #1323.

afuna:
I agree that this can be finetuned more wrt placement, but deem it acceptable (and easier to finetune once things are already broken down, than risk a massive code merge in the future).

PR #1347, opened 2015-04-15
[#1345] restore missing button on dialog to insert image (RTE)

### LINKED ISSUE #1345: "Insert" button to insert image on RTE no longer exists
Problem: When you pull up the insert image dialog on the RTE, there is not a "insert" or "submit" button, only a cancel button. There used to be, but it disappeared at some point. There have been at least three reqs about it in the last week.
Workaround: once you've put the image URL into the box, hit enter.
Fix: please make there be a nice tidy button again, or at least a note on the box that says something like, "when you're done, hit enter."

insert
http://www.dreamwidth.org/support/see_request?id=30743

kareila:
Restores initialization of the SiteConfig variable, which is used by fck_image.js.

(This just readds a line that was removed in af3dba5 - I realized that the variable was still being referenced elsewhere. It did make the missing button reappear when I tested it.)

Fixes #1345.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -3129,6 +3129,8 @@ sub rte_js_vars {
                 }
             }
         }
+
+        var SiteConfig = new Object();
     </script>^;
 
     return $ret;


afuna:
Oops, thanks for tracking that down!

zorkian:
Nice catch! My bad. :)

PR #1348, opened 2015-04-15
fix refundtopoints error message for case where use attempted twice within 30 days

### LINKED ISSUE #1340: convert-to-points system does not distinguish between "hasn't been 30 days since last conversion" and "not enough time to convert"
http://www.dreamwidth.org/shop/refundtopoints is supposed to not let you convert paid time to points in either of the following cases:

1. It's been less than 30 days since the last time you converted paid time to points.
2. You have fewer than 30 days of paid time on the account.

However, no matter which of the two conditions it's failing on, the error message is always /shop/refundtopoints.tt.toofew ("Sorry, you have less than 30 days of paid time on your account.") In condition 1, the error should be "Sorry, you've converted your paid time to points within the last 30 days. You won't be able to use this tool again until [date]."

kareila:
The first commit corrects the error message, the second commit adds config values needed to test the fix, and the third commit adds the requested date info.

Fixes #1340.

### DIFF
--- a/etc/config.pl
+++ b/etc/config.pl
@@ -390,6 +390,7 @@
             '_key' => 'paid_user', # Some things expect that key name
             '_visible_name' => 'Paid Account',
             '_account_type' => 'paid',
+            '_refund_points' => 30,
             'activeentries' => 1,
             'bookmark_max' => 500,
             'checkfriends' => 1,
@@ -446,6 +447,7 @@
             '_key' => 'premium_user',
             '_visible_name' => 'Premium Paid Account',
             '_account_type' => 'premium',
+            '_refund_points' => 41,
             'activeentries' => 1,
             'bookmark_max' => 1000,
             'checkfriends' => 1,
--- a/views/shop/refundtopoints.tt.text
+++ b/views/shop/refundtopoints.tt.text
@@ -9,6 +9,8 @@
 
 .complete=Your exchange has been completed.
 
+.nextrefund=You won't be able to convert paid time to points again until after [[date]].
+
 .noteligible=Sorry, your account type is not eligible for a points conversion.
 
 .refund=You are eligible to exchange <strong>[[days]] days</strong> of [[type]] for <strong>[[points]] [[sitename]] Points</strong>.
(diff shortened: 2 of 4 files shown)


views/shop/refundtopoints.tt.text
```diff
@@ -9,6 +9,8 @@
 
 .complete=Your exchange has been completed.
 
+.nextrefund=You won't be able to use this tool again until after [[date]].
```
afuna: Suggest wording change: "You'll be able to [apply refunds?] again after [[date]]"

etc/config.pl
```diff
@@ -390,6 +390,7 @@
             '_key' => 'paid_user', # Some things expect that key name
             '_visible_name' => 'Paid Account',
             '_account_type' => 'paid',
+            '_refund_points' => 30,
```
afuna: Noting that these are in the prod configs (wanted that to be clear because I double-checked that that was the case / these didn't need to be added!)

views/shop/refundtopoints.tt.text
kareila: I would check with @rahaeli since that was her requested phrasing.

rahaeli: Yeah, that was just totally off the cuff, people should always feel free to rewrite anything I specify!

"You won't be able to refund paid time to points again until [[date]]"?

kareila:
"Sorry, you can only refund paid time once every thirty days. You won't be able to convert paid time to points again until after [[date]]."

rahaeli:
s/refund/convert in the first sentence for consistency (and because 'refund' can be a touchy word; I should go make sure we're using it properly, actually) and we're good.

kareila:
Bah, rewriting existing translation strings is so much more annoying than adding new ones. :)

rahaeli:
LOL. I can go fix it on the site instead. :)

rahaeli:
Huh, I went to go fix on site and can't find the text string, by item code OR text. Was it just added? Either way, it doesn't seem to exist on the site, so you can just change it in the file.

kareila:
Probably because it was never successfully displayed before.

*makes ritual warding gestures*

rahaeli:
Ia! Ia! Translation system f'thag'n! In its lake in Gi'th'ub the translation system lies sleeping!

PR #1350, opened 2015-04-17
Fix some tests

kareila:
- use-strict.t needed to skip t/config-test-private.pl
- create-url.t was failing to predict SSL response from user domains
- various errors in post.t (most, but not all)


afuna:
thank you!

PR #1351, opened 2015-04-17
[#1220] t/post.t: don't test for removed altlogin functionality

### LINKED ISSUE #1220: t/post.t is broken
I have been fixing ALL the tests after the move to a separate test DB.

However, this seems like code-break, not test-break.

kareila:
We no longer allow remote users to post to an alternate personal account while
logged in, so now the tests for that functionality are failing.  Since this is
deliberate, remove the relevant tests.

Fixes #1220.

### DIFF
--- a/t/post.t
+++ b/t/post.t
@@ -15,7 +15,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 136;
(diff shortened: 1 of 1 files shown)


afuna:
thanks for taking care of that!

PR #1357, opened 2015-04-24
[#772] English-stripping and refactoring for OpenID account creation

### LINKED ISSUE #772: a few highly user-unfriendly error messages in talkpost_do.bml
Was browsing talkpost_do.bml to look for the code for fixing issue #742 and realized there are a bunch of bad/unclear/overly-technical error messages, like:

            return LJ::bad_input("Invalid return_to");
            return LJ::bad_input("Can't verify identity: ".$csr->err) unless $vident;
            return LJ::bad_input("Invalid characters in identity URL.") if $url =~ /[\<\>\s]/;
            return LJ::bad_input("Couldn't vivify your account (but we verified that you're " . LJ::ehtml($url) . ")")

...etc, etc. (That last one was what made me open this bug, because whaaaaaat. I think the only reason nobody's opened a support request to ask wtf is because it's a very rare error.)

Ideally this should be English-stripped (parts of it already are) but at the very least the error messages should be more understandable.

kareila:
It was pointed out that some obscure OpenID account creation errors
were English-stripped in one place (/openid/login.bml) and not another
(/talkpost_do.bml).  This updates the English-stripping in both places
to be consistent and slightly more user-comprehensible.

Some of the validation process has been refactored into a new class
function, LJ::User::load_from_consumer, to reduce code duplication.

Fixes #772.

### DIFF
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -1049,6 +1049,38 @@ sub load_identity_user {
     return $u;
 }
 
+# class function - refactoring verification of Net::OpenID::Consumer object
+# returns the associated user object, or undef on failure
+sub load_from_consumer {
+    my ( $csr, $errorref ) = @_;
+
+    my $err = sub { $$errorref = LJ::Lang::ml( @_ ) if defined $errorref && ref $errorref };
+
+    my $vident = eval { $csr->verified_identity; };
+
+    unless ( $vident ) {
+        my $msg = $@ ? $@ : $csr->err;
+        $err->( '/openid/login.bml.error.notverified', { error => $msg } );
+        return;
+    }
+
+    my $url = $vident->url;
+
+    if ( $url =~ /[\<\>\s]/ ) {
+        $err->( '/openid/login.bml.error.invalidcharacters' );
+        return;
+    }
+
+    my $u = load_identity_user( "O", $url, $vident );
+
+    unless ( $u ) {
+        $err->( '/openid/login.bml.error.notvivified', { url => LJ::ehtml( $url ) } );
+        return;
+    }
+
+    return $u;
+}
+
 # journal_base replacement for OpenID accounts. since they don't have
 # a journal, redirect to /read.
 sub openid_journal_base {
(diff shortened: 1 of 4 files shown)


kareila:
I was able to test the changes to /openid/login but couldn't quite figure out how to enter the relevant code block in talkpost_do (something about ecphash?) - anyway, since they're so similar, I don't think it's a huge problem.

afuna:
ecphash -- that's from when you're replying via the comment form in the
email (in this case that's all dead code)

kareila:
Oh, well, never mind testing that then. :P

rahaeli:
Ooh, can you open a bug to put that under the code machete?

kareila:
I will do that.

@afuna, do you want to commit this as is or do you want me to pull out the changes to talkpost_do and undo the refactoring?

afuna:
@kareila I'm in favor of pulling as much out of BML as possible so let's keep the refactor!

kareila:
Actually, with the ecphash stuff removed, I can see it's actually not related - it's in a block that's only executed if `$GET{'pendcid'}` is true.  I'm not sure what's going on with that either, though.

afuna:
Hmm, oh that's some weird code... okay basically what happens is that if
an openid user comments for the first time, we have to take a detour to
create their journal for them. So we save their comment in (the
journal's) pendcomments where they wait, pending.

Pending is such a weird word.

On Fri, Apr 24, 2015, at 05:58 PM, Jen wrote:
> Actually, with the ecphash stuff removed, I can see it's actually not
> related - it's in a block that's only executed if $GET{'pendcid'} is
> true. I'm not sure what's going on with that either, though.


>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/1357#issuecomment-96113585

kareila:
I can kind of follow it back and forth, but I still can't figure out how to invoke it ¯\_(ツ)_/¯

afuna:
It should happen when you make a comment with an openid account that's never logged in before (so start logged out -- then click the openid option and use an account that's new to your hack)

kareila:
If I'm logged out, the only option that's selectable is "logged in account" - the OpenID selection is greyed out and displays a link to /openid, which goes through /openid/login.bml.  (I tried that first.)

afuna:
Hrrrrm can you double-check your comment settings for your journal? On mine (in DW), I do get all the options (anonymous, openid, username/password)

azurelunatic:
OpenID is treated like anonymous if not email verified, like logged in if
verified. So to allow new OpenID you need to allow anonymous.

kareila:
Thanks for clearing that up for me - I think I've managed it now, and it seems to be doing the right thing.

afuna:
Thanks @kareila  (and @azurelunatic!)

PR #1364, opened 2015-04-28
[#1363] Optionally treat "+" as spaces

### LINKED ISSUE #1363: tagnav memory doesn't handle special-characters
Navigating using the tagnav links is intended to 'remember' which tag you had selected. When the tag in question has any sort of special characters (including, notably, space characters), the new page does not remember the tag selection and it must be reselected.

as best as I can tell it's something gone sideways in the 'compare with the URL string' logic. As an example: `http://synecdochic.dreamwidth.org/353197.html#tagnav-broken+wings` doesn't work but `http://synecdochic.dreamwidth.org/353197.html#tagnav-broken wings` (hand-edited to such) works as expected.

afuna:
DW's urlencode function does percent encoding for most special
characters except spaces; make the JS match this behavior.

Fixes #1363.

### DIFF
--- a/htdocs/js/journals/jquery.tag-nav.js
+++ b/htdocs/js/journals/jquery.tag-nav.js
@@ -102,11 +102,12 @@ jQuery(document).ready(function() {
 var hash = location.hash;
 if ( hash.indexOf( "#tagnav-" ) == 0 ) {
     $(window).load(function() {
-        var tag = hash.slice(8);
+        var tagnav_tag = hash.slice(8);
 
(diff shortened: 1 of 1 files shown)


afuna:
I am honestly torn on this one, because historically that treatment of spaces in tags has caused edge cases to pop up every once in a while (like right now!), so this feels like a bandaid on top of a bandaid. But the other choice is Yet Another Confusingly Similarly Named function in LJ::TextUtil, and we have so many of those too!

zorkian:
Seems legitimate to me.

PR #1365, opened 2015-05-06
Send offsite mood links through proxy.

### LINKED ISSUE #1342: Mood themes aren't https
They need to be run through the proxy I think! Because they break https on read pages ;)

anall:
Fixes #1342

[ This is untested except for the fact that the URL form looks correct.  I do not have the means or ability to test HTTPS proxying because there is no information on how the proxy end of things work ]

### DIFF
--- a/cgi-bin/DW/Mood.pm
+++ b/cgi-bin/DW/Mood.pm
@@ -16,6 +16,7 @@ package DW::Mood;
 use strict;
 use warnings;
 
+use LJ::CleanHTML;
 
 ### MOOD (CLASS) METHODS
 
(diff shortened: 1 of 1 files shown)


afuna:
I suggest LJ::CleanHTML::https_url instead! That does a couple extra checks for if it's already https or if we're hosting it.

anall:
That work?

[ again, untested besides the URL looking like it is getting mangled properly ]

afuna:
Hmm it still seems to be using DW::Proxy::get_proxy_url.

anall:
Try now?  I fail at git.

afuna:
Looks good from an eyeball, let me see if I can confirm (but maybe not tonight!)

rahaeli:
I sometimes think "I fail at git" is the default state. I know I do too. :)

afuna:
Hmmm I think that this always converts to HTTPS, whether or not the page being viewed is in HTTP or HTTPS. But I think that given:

a) the url is cached
b) we plan to switch to https everywhere

That actually is a non-issue.

PR #1369, opened 2015-05-07
Fixes issue 1346

momijizukamori:
Using an inline element (img) inside a block element (div)
causes whitespace to appear below the baseline for alignment.
Changing the image display to 'block' fixes this.

### DIFF
--- a/styles/planetcaravan/layout.s2
+++ b/styles/planetcaravan/layout.s2
@@ -457,6 +457,10 @@ h2#pagetitle, h2#subtitle {font-weight: normal; padding:.25em; margin:0; color: 
         }
 }
 
+.userpic img {
+    display: block;
+}
+
 $userpic_css
 
 .no-userpic .userpic {border: 0;}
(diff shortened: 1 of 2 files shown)


afuna:
Oh huh. Nice fix. 

I'd like the commit message to be clearer though, so that it's obvious what the commit is about when you run git log (or git log --oneline)

The first line should describe the fix, e.g.,:

    Fix whitespace appearing below the icon in Planet Caravan

Then a longer description (what you have for the next lines works great)

And then "Fixes #1369"

(Github is strict about the formatting and won't do its magic otherwise).

momijizukamori:
Hope that did it - this is what I get for not doing anything for 11 months (and the internet pulled up the solution!)

afuna:
Almost perfect! (Typo: 1346 not 1347)

afuna:
It's all good, I typoed my suggestion to you as well ;) (And by typoed I mean I gave you the entirely wrong bug number)

PR #1375, opened 2015-05-09
Add shorter sitenames to username tag

### LINKED ISSUE #1417: Add shorter sitename options to username tag
As discussed in IRC (but not in dw_suggestions as far as I can tell, at least since 2011), it would be really nice for the <user name="x" site="livejournal.com"> (e.g.) tag to also permit  site="lj" or site="livejournal" (and similar for all tags).

kaberett:
Adds common abbreviations to the site="" argument of the user tag
(e.g. livejournal and lj both now map to livejournal.com).

Fixes #1417.

### DIFF
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -31,7 +31,8 @@ LJ::ModuleLoader->require_subclasses( "DW::External::Site" );
 my %domaintosite;
 my %idtosite;
 
-# static initializers
+### static initializers
+# with tld
 $domaintosite{"livejournal.com"} = DW::External::Site->new("2", "www.livejournal.com", "livejournal.com", "LiveJournal", "lj");
 $domaintosite{"insanejournal.com"} = DW::External::Site->new("3", "www.insanejournal.com", "insanejournal.com", "InsaneJournal", "lj");
 $domaintosite{"deadjournal.com"} = DW::External::Site->new("4", "www.deadjournal.com", "deadjournal.com", "DeadJournal", "lj");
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/External/Site.pm
```diff
@@ -96,7 +130,8 @@ sub get_site {
                 map { lc $_ }
                 split( /\./, $site );
 
-    return $domaintosite{"$parts[-2].$parts[-1]"} || DW::External::Site::Unknown->accepts( \@parts ) || undef;
+    return $domaintosite{"$parts[-2].$parts[-1]"} || $domaintosite{"$parts[-1]"} ||
```
afuna: If I'm reading this right, I think we only need the list of aliases (above), and the accepts (below). 

The main thing being that once the $domaintosite{"parts[-1]") matches, we return the relevant DW::External::Site instance directly -- never end up calling DW::External::Site::Unknown->accepts 

(huh in fact it seems like we never actually do call any of the DW::External::Site->accepts() no matter what).

kaberett: So those trailing conditions pre-exist my touching the code, sooo I'm not entirely sure what they're for. Happy to remove them if you're confident we don't need them but I assumed they wanted leaving well enough alone because they'd been put there for a reason! :-)

afuna: Yes! Sorry I wasn't clear I meant we can cut out a lot of the complexity of the patch by removing the changes in the individual DW::External:::Site classes :)

Also! I forgot to say this earlier but please include the issue number in your commit message (so we can track which one it was)

kaberett: Oh, sure, okay - I'll remove those then.

The reason I didn't include an issue number is that to the best of my knowledge there isn't an open issue for this - I discussed it with folk in IRC, we agreed it would be nice to have, but I didn't ever file an issue. Let me know if you'd like me to file one retroactively?

rahaeli: Yes, please do file an issue for it so you have one to hang the PR on. :)

rahaeli: PS: I think there might be a suggestion for something like this. If you have a quick look-through and see, please also include the suggestions link in the issue that you open!

kaberett:
@rahaeli: if there was a suggestion it's not been made since 2011 (or if it has I've not put *either* of the sensible tags on it). However we have discussed it in IRC a couple of times. Going to open an issue (and implement requested changes...) now...

kaberett:
@afuna Done!

afuna:
Thank you!

PR #1376, opened 2015-05-09
Use client_ip.

### LINKED ISSUE #923: $r->connection->remote_ip removed on Apache 2.4
Apache 2.4 / mod_perl 2.4 have removed `$r->connection->remote_ip` -- Replacements are either `$r->connection->client_ip` ( closest load balancer, I think ) and `$r->useragent_ip` ( actual user IP -- may not be secure depending on how the `X-Forwarded-For` headers are trusted in upstream proxies ) seem to be the possible replacements.

`$r->useragent_ip` seems to be the correct replacement, however I am unsure if that works on pre-2.4 and do not have easy access to check.

anall:
Note: this breaks everything for anything before Apache 2.4

Fixes #923 

Per comments on #923, we are not going to do this in a "works on 2.4 and 2.2" way.

**NOTE**: We will have to make sure the hack box is prepared to be updated before this is merged, as well as making sure production is prepared.


afuna:
There are a couple instances of  `$apache_r->connection()->client_ip`, do those need to be changed as well?

anall:
Ugh yes they do. Those never caused an error for me so I missed them.

When will we have DW::Request doing *everything*? ;_;

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -95,7 +95,7 @@ sub handler
             # by the fact that mod_proxy did nothing, requiring mod_proxy_add_forward, then
```
anall: Do we still even need this block.

If I understand the difference between `remote_ip` and `client_ip` correctly, this negates the need for us doing this ourselves.

cgi-bin/Apache/BML.pm
```diff
@@ -1916,7 +1916,7 @@ sub get_path_info
 sub get_remote_ip
 {
     my $apache_r = BML::get_request();
```
anall: Any objections to making these functions in `Apache::BML` be a lightweight wrapper to `DW::Request`?

BML is probably, sadly, not going to die anytime soon.

anall:
{ as an aside, I don't want to even consider doing anything this børk-causing before Jun 1 -- don't want to prevent people from hitting their PR counts }

afuna:
That's fair. I think that it probably would be okay though, since first
thing most people do (or should) is update their repos.

cgi-bin/Apache/BML.pm
zorkian: Wouldn't it be easier to update callers to use DW::Request?

If not, then I have no objections.

cgi-bin/Apache/LiveJournal.pm
zorkian: Yeah, I thought about removing it, and think you may be right. I didn't feel like I had the :brain: to sit down and puzzle it all out when I was working on it though...

zorkian:
Okay so should we not merge it yet then? or is it OK?

It's only two weeks to 6/1, so I'm fine waiting.

anall:
We need to time upgrading the hack box ( @Sophira ) with merging this pull request.

I really do not want to hurt anyone's chances of coming to OSB because this patch prevented them from doing devwork.

cgi-bin/Apache/BML.pm
anall: Might want to see how many uses there are, and how many of them are in BML files.

Sophira:
Thanks for the ping!

This is going to require either a custom-compiled Apache or an update to a newer Ubuntu release. I've been meaning to do the latter for some time and had initial talks with @zorkian about it, but I haven't followed up on it. Mark, we should talk. :)

Unfortunately, though, my laptop died in the last few days and I don't yet have another computer to work from. (I'm typing this on my mobile.) I estimate that I won't have a computer for about two weeks or so, and preparing migration to a new server is something I'd really rather not try to do from my phone, though if needs must I can try.

I sincerely apologise that I haven't mentioned this before; it's only happened very recently and things have been moving too quickly for me to keep up.

There are migration scripts in the repository (although I wasn't able to take a copy of the latest migration-send script from the old machine so it might need a bit of an update) and there's full instructions on setting up a new Dreamhack host in the /INSTALL file in the repo (it doesn't cover migrating, but that should hopefully be simple enough with the scripts) , so if anybody else wants to head up the project for now, that could work. You'll want to talk with Mark about it, and I can give explanations of anything you need. Most of the tougher bits of the process will be the organisation.

If speed is the primary concern then a version of Apache 2.4 built for Ubuntu Lucid is probably the best bet. If anybody's willing to make a .deb then that would make things easy. Otherwise, it probably wouldn't be fantastically difficult to compile from the phone, but it certainly won't be comfortable. ;p

Let me know what you'd like to do. I'm about to go to sleep so won't respond for a few hours but I'll read any responses when I wake up.

zorkian:
We're at OSBridge now can I land this? :)

anall:
This will break everyone on hack boxes.

zorkian:
Okay! Landing. Now that we're at OSB (which was your concern) we can make sure people update their environments. It's also after the date @Sophira said she'd get a computer back and have time to poke at things etc, so I think we're good to go.

PR #1380, opened 2015-05-10
Additional link on sent-PM splash page

### LINKED ISSUE #1232: Add a link to view the message to the "Message sent" page
From http://dw-suggestions.dreamwidth.org/1360416.html:

> Suggest adding a link to the "Your message has been sent successfully" page, so the sender can easily view the sent message.

woggy:
Provides an additional link on the splash page that occurs after sending
a PM, that redirects to an inbox view for the just-sent message.
Explicit override on the expand/collapse view expands the first sent
message even though it's marked as read.

Known Issue: If the user clicks through too quickly, the page will load
before the worker has finished sending the PM. A refresh will cause the
new PM to show up.

Fixes #1232

### DIFF
--- a/cgi-bin/LJ/NotificationInbox.pm
+++ b/cgi-bin/LJ/NotificationInbox.pm
@@ -140,6 +140,14 @@ sub usermsg_sent_items {
     return $self->subset_items(@events);
 }
 
+sub usermsg_sent_last_items {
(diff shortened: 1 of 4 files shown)


afuna:
I think the best possibility given technical limitations.

PR #1382, opened 2015-05-12
[#1381] Close resp.Body after we're done with it

### LINKED ISSUE #1381: Proxy causes system to run out of file descriptors when left running too long
http://www.dreamwidth.org/support/see_request?id=30863

An image in an entry won't display.

The image lives over at photobucket.com/albums/bb317/AndrewDucker/2015-infographic.jpg.png 

The URL we're showing is https://p.dreamwidth.org/12b0d34ed254/i207.photobucket.com/albums/bb317/AndrewDucker/2015-infographic.jpg.png

So somewhere between when they posted and when we served the content, something went sideways. I am probably explaining this poorly, so hopefully someone can figure out what happened. Fu said to open a ticket.

afuna:
So that we don't risk leaking file descriptors.

Fixes #1381.

### DIFF
--- a/src/proxy/main.go
+++ b/src/proxy/main.go
@@ -171,6 +171,7 @@ func getProxyFile(token, url string) (string, error) {
 		log.Printf("Failed to fetch %s: %s", url, err)
 		return "", err
 	}
+	defer resp.Body.Close()
 
 	// If it's too large, we don't want it!
 	if resp.ContentLength > MAXIMUM_SIZE {


afuna:
Pretty new to Go, so I'd love to have someone double-check me on this.

Basis for patch: code example on https://golang.org/pkg/net/http/: "The client must close the response body when finished with it"

I've built and run the proxy with this code, and I can confirm that this at least does not break things. :roller_coaster:

zorkian:
Nice catch! I think this is great.

PR #1393, opened 2015-05-23
[#1315] allow spamreports to handle IPv6 addresses

### LINKED ISSUE #1315: spamreports by IP does not recognize IPv6 IPs
We're getting IPv6 IPs now, and /admin/spamreports by IP doesn't recognize 'em; it gives a "No such IP" error when you try to load the "by IP" view (eg, http://www.dreamwidth.org/admin/spamreports?mode=view&by=ip&what=2001:db8:0:1&state=open ) (IP replaced by an example IP).

kareila:
This patch does two things:

1. Widens the column for IP addresses in the spamreports table from 15 characters to 39.
An IPv6 address may contain 8 16-bit colon-separated hexadecimal codes for a maximum
width of 8*5-1 = 39 characters.

2. Allows the spamreport page to load for any given IP value that is 39 characters or less.
If the IP does not exist in the spamreports database, the "noreports" error will be shown.
If the IP exists, the report may be viewed and closed, but the sysban option will not be
presented unless it is a valid IPv4 address.  Once we verify sysban support for IPv6,
this clause may be removed.

Fixes #1315.

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -4173,6 +4173,13 @@ EOF
         do_sql( 'DELETE FROM codes WHERE type = "country"' );
         set_dbnote( 'remove_countries_from_codes', 1 );
     }
+
+    # widen ip column for IPv6 addresses
+    if ( column_type("spamreports", "ip") eq "varchar(15)" ) {
+        do_alter( "spamreports",
+                  "ALTER TABLE spamreports ".
+                  "MODIFY ip VARCHAR(39)" );
+    }
 });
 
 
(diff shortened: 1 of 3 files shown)


kareila:
For the verify_ipv6 subroutine I followed the IPv6 address format explanation from http://computernetworkingnotes.com/ipv6-features-concepts-and-configurations/ipv6-address-types-and-formats.html

kareila:
I am also open to making verify_ipv6 a system function, if we can figure out where it should live in the codebase :)

azurelunatic:
Yay!!!

anall:
Sysbanning ipv6 addresses is very likely quite useless unless we've got a way to block prefixes. I've got easy access to a /64, which gives me 18,446,744,073,709,552,000 addresses.  I could probably get a /48 with a little more work and have 1,208,925,819,614,629,500,000,000 addresses.

kareila:
Right, that's why I didn't address that here. It needs to be opened as a separate issue.

kareila:
Bump!

zorkian:
Thank you for the bump, and sorry for the delay. I agree that making this just work, even if it means we can't sysban ipv6, is probably the right tradeoff here.

PR #1394, opened 2015-05-23
Remember status of "Use post time" checkbox.

### LINKED ISSUE #1229: Allow "Use the time when entry is posted" to be set as default
From http://dw-suggestions.dreamwidth.org/1394378.html

>Allow "Use the time when entry is posted" to be set as default when posting entries.

woggy:
A new userprop: displaydate_check which tracks what the last status of
this checkbox was, and sets the initial state on new/edited entry
appropriately.

Fixes #1229

### DIFF
--- a/bin/upgrading/proplists.dat
+++ b/bin/upgrading/proplists.dat
@@ -221,6 +221,14 @@ userproplist.disable_comm_promo:
   indexed: 0
   multihomed: 0
   prettyname: Administratively disable community promo features for a user
+  
+userproplist.displaydate_check:
+  cldversion: 4
+  datatype: bool
+  des: Use time of posting by default for this user.
+  indexed: 0
+  multihomed: 0
+  prettyname: Automatically sets entry page to use time of posting for timestamp.
 
 userproplist.dismissed_page_notices:
   cldversion: 4
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -293,6 +293,7 @@ sub _init {
     my $panels;
     my $formwidth;
     my $min_animation;
+    my $displaydate_check;
     if ( $u ) {
         # icons
         @icons = grep { ! ( $_->inactive || $_->expunged ) } LJ::Userpic->load_user_userpics( $u );
@@ -344,6 +345,7 @@ sub _init {
         $panels = $u->entryform_panels;
         $formwidth = $u->entryform_width;
         $min_animation = $u->prop( "js_animations_minimal" ) ? 1 : 0;
+        $displaydate_check = $u->displaydate_check ? 1 : 0;
     } else {
         $panels = LJ::User::default_entryform_panels( anonymous => 1 );
     }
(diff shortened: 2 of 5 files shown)


bin/upgrading/proplists.dat
```diff
@@ -221,6 +221,14 @@ userproplist.disable_comm_promo:
   indexed: 0
   multihomed: 0
   prettyname: Administratively disable community promo features for a user
+  
+userproplist.displaydate_check:
+  cldversion: 4
+  datatype: bool
+  des: Status of the update_displaydate checkbox, which sets which posting time to use.
```
afuna: Since this is a boolean, I think that "which sets which posting time to use" is a little vague. How about something like "Automatically updates the posting time in the (new) entry page"?

cgi-bin/DW/Controller/Entry.pm
```diff
@@ -344,8 +345,10 @@ sub _init {
         $panels = $u->entryform_panels;
         $formwidth = $u->entryform_width;
         $min_animation = $u->prop( "js_animations_minimal" ) ? 1 : 0;
+        $displaydate_check = $u->displaydate_check ? 1 : 0;
     } else {
         $panels = LJ::User::default_entryform_panels( anonymous => 1 );
+        $displaydate_check = 0;
```
afuna: This doesn't need to be set explicitly, btw, because it defaults to 0 when you declare $displaydate_check above.

cgi-bin/DW/Controller/Entry.pm
```diff
@@ -1203,6 +1209,8 @@ sub _persist_props {
     my ( $u, $form ) = @_;
 
     return unless $u;
+    
+    $u->displaydate_check( $form->{update_displaydate} ? 1 : 0);
```
afuna: style nitpick: consistent whitespace please (don't need it after the parentheses above)

htdocs/js/pages/entry/new.js
```diff
@@ -673,6 +673,7 @@ var postForm = (function($) {
         initSticky(entryForm);
 
         $("#js-usejournal").triggerHandler("change");
+        $("#js-entrytime-autoupdate").triggerHandler("click");
```
afuna: Oh nice, I am guessing this is to make sure that when the checkbox is checked by default, the associated actions / states come into play right?

woggy: Yes, specifically the disabling of the date/time entry boxes just above it. :)

views/entry/module-displaydate.tt
```diff
@@ -74,12 +74,22 @@ the same terms as Perl itself.  For a copy of the license, please reference
 
     <div class="row">
         <div class="columns">
-            [%- form.checkbox_nested( label = dw.ml( ".label.autoupdate" )
-                name = "update_displaydate"
-                id = "js-entrytime-autoupdate"
-
-                value = "1"
-            ) -%]
+            [% IF "$displaydate_check" %]
```
afuna: So this is completely functional (and sorry for not being available on IRC when you were problem-solving this!)

I'd like to suggest a cleaner alternative:

    default = displaydate_check

afuna:
Mwahaha I like when other people touch the new entry page :)

PR #1396, opened 2015-05-24
Preserve sticky-entry value when module unloaded

### LINKED ISSUE #1343: Beta update: sticky entry getting unstickied when edited
http://www.dreamwidth.org/support/see_request?id=30644

Sticky an entry; edit it; refresh. The entry is not longer stickied. The URL is gone from Account settings.

woggy:
Before, there was no way to distinguish between "user left checkbox
unticked" and "checkbox doesn't exist" for the sticky-entry checkbox,
resulting in formerly-sticky entries losing their status.  This fix
wraps the remove logic in another if statement, so that it will only
perform the remove if the checkbox existed to be unticked.

Fixes #1343


afuna:
That's number 5 now isn't it? :)

woggy:
IT SURE IS.

PR #1403, opened 2015-05-31
Hide banned users from profile (fixes #1279)

### LINKED ISSUE #1279: banning a user does not remove them from your profile
Expected behavior (ie, the way LJ does it and the way I thought we did it too): when you ban a user from your journal, that account should no longer appear on your Subscribers/Has Access From lists.

Actual behavior: account still appears on your profile.

(This shouldn't affect the banned user's profile -- that is to say, if I susbcribe to Alice, and she bans me from her journal, my name should not appear on Alice's profile under 'Subscribers', but Alice's name should still appear on my profile as one of my subscriptions.)

kaberett:
Prevents banned users from appearing in watch/trust/subscribe lists
on profile pages. (Thanks Dre and Jen! <3)

### DIFF
--- a/htdocs/profile.bml
+++ b/htdocs/profile.bml
@@ -695,6 +695,7 @@ body<=
     my ( @trusted_userids, @trusted_by_userids, @mutually_trusted_userids, @not_mutually_trusted_userids, @not_mutually_trusted_by_userids );
     my ( @watched_userids, @watched_by_userids, @mutually_watched_userids, @not_mutually_watched_userids, @not_mutually_watched_by_userids );
 
+
     # no mutual trust lists for identity accounts since they can't trust anyone
     @trusted_by_userids = $u->trusted_by_userids if $u->is_identity;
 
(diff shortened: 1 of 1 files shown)


htdocs/profile.bml
```diff
@@ -742,6 +743,18 @@ body<=
         @posting_access_to_userids, @posting_access_from_userids,
     );
 
+    # pull in list of banned users
+    my $banned = LJ::load_rel_user($u, 'B') || [];
+    my %banned_users = map { $_ => 1 } @$banned;
+
+    # remove banned users from lists to display
+    @$_ = grep ( !defined $banned_users{ $_ }, @$_ )
```
zorkian: You probably want `exists` here instead of `defined`.

Also, our style is to use `grep { }` (with braces instead of parenthesis).

htdocs/profile.bml
```diff
@@ -753,6 +766,7 @@ body<=
 ################################################################################
 ##### PEOPLE LISTS
 
+
```
zorkian: I'd prefer not to add extra blank lines.

htdocs/profile.bml
```diff
@@ -695,6 +695,7 @@ body<=
     my ( @trusted_userids, @trusted_by_userids, @mutually_trusted_userids, @not_mutually_trusted_userids, @not_mutually_trusted_by_userids );
     my ( @watched_userids, @watched_by_userids, @mutually_watched_userids, @not_mutually_watched_userids, @not_mutually_watched_by_userids );
 
+
```
zorkian: Extra blank line.

zorkian:
As this is a change in behavior, would be great to get confirmation from @zarhooie or @rahaeli that this is a desired change.

kareila:
@rahaeli opened #1279 which this is the fix for. I would call that an endorsed change in behavior.

kareila:
Fixes #1279.

(This needs to be on a line by itself in a comment for Github to recognize it and close the issue when the PR is accepted.)

PR #1404, opened 2015-05-31
Give the user a hand fixing UTF-8 errors

### LINKED ISSUE #1384: Have a more descriptive UTF-8 error
http://www.dreamwidth.org/support/see_request?id=30868

When a user attempts to post something that has a funky character, they get the following error:
 "Bad Unicode Input
Your browser sent some text which isn't recognised as valid text in the UTF-8 encoding. This might happen if you forced your browser to view the previous page in some encoding other than UTF-8. It may also indicate a bug in the browser. If you can't get around this error, contact us." 

When this comes through Support, we usually tell them to paste the contents of the entry into something like Notepad, and then try to paste from there. It would be better if the error message told them to do that, or maybe referred to a FAQ.

me-and:
When displaying an error due to UTF-8 encoding errors in submitted
strings, rather than just punting the user straight to support, give
them a hand to fix the problem themselves.

Fixes #1384

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -7,7 +7,7 @@ Backlink=[<a href="[[link]]">&lt;&lt;</a> [[text]]]
 
 bml.badcontent.body=One or more errors occurred processing your request.  Please go back, correct the necessary information, and submit your data again.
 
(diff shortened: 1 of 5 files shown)


zorkian:
Does this workaround work, and is this the recommended way? @zarhooie?

rahaeli:
Yep, it's the text that @zarhooie requested in the original issue. (It's not a perfect method of resolving the problem, but 90% of the time the issue is "wrote the entry in Word, Word inserted Wordbarf" and pasting into Notepad will work for that.)

zorkian:
Thanks !:)

rahaeli:
In conclusion, I hate both character encodings and Word.

PR #1405, opened 2015-05-31
Provide title text for tag navigation button

### LINKED ISSUE #1332: add title text for the tagnav image
http://www.dreamwidth.org/support/see_request?id=30621 brought up that the tagnav image (the green plus sign that brings up the forward/back arrows) has alt text, but not title text. It would probably be helpful to have title text that describes what to do with the image, for people for whom it isn't intuitive.

Let's say title text of "Tag navigation: select this, then select a tag to browse forward or backward through that tag." (A bit long for title text, but it's the most succinct I can make that explanation!)

me-and:
The little green tag navigation button doesn't have any title text.  It
would be nice if there were something there to explain how to use the
thing, which will hopefully pop up on mouseover in browsers.  Let's add
that.

Fixes #1332

### DIFF
--- a/htdocs/js/journals/jquery.tag-nav.js
+++ b/htdocs/js/journals/jquery.tag-nav.js
@@ -8,6 +8,7 @@
     function collapsed($trigger) {
         $trigger.attr("aria-expanded", "false");
         $trigger.attr("alt", "Show tag navigation");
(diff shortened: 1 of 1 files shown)


afuna:
Good call on having additional info.

Issue #1398, opened 2015-05-25
Admin Console Reference missing CSS styling


kareila:
The list of console commands in the console reference guide used to have different styling depending on whether or not you were allowed to use the command.  Since it was converted to Foundation/SCSS, that styling has disappeared.

The two div classes are "enabled" (for commands you can use) and "disabled" (for unavailable commands).  The current styling is fine for enabled, but disabled commands used to be printed in much lighter text to indicate that they were unavailable to use.

We need to readd some CSS so that enabled and disabled commands can once again be differentiated from each other.

me-and:
Claimed

PR #1406, opened 2015-05-31
Add different formatting for unavailable commands

me-and:
Apply different formatting in the Admin Console for enabled and disabled
commands, to help differentiate the two.

This is the free part of the fix for #1398

### DIFF
--- a/cgi-bin/LJ/Console.pm
+++ b/cgi-bin/LJ/Console.pm
@@ -197,7 +197,9 @@ sub command_reference_html {
 
         $ret .= "<hr /><div class='$style'><h2 id='cmd.$cmd'><code><b>$cmd</b> ";
         $ret .= LJ::ehtml($class->usage);
-        $ret .= "</code></h2>\n";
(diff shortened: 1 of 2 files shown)


me-and:
The corresponding non-free pull request is dreamwidth/dw-nonfree#100

afuna:
I'd like to put this styling somewhere in htdocs/scss/skins. (_skin-colors is probably a good 
one).

Reason being that I'd like to avoid modifying the files under htdocs/scss/foundation*, to make it easier to upgrade if needed / when desired. 

I also think that makes adding stuff to dw-nonfree unnecessary -- unless the colors need to be different!

me-and:
New pull request per @afuna's suggestions; there's now also no need for the changes to dreamwidth/dw-nonfree#100.

I've also gone a bit further and modified `Console.pm` in the name of accessibility, so that the difference between available and unavailable commands is highlighted by something more than just a colour change.

afuna:
Awesome, thanks for going the extra mile.

PR #1408, opened 2015-05-31
Correct Google Spreadsheets whitelist regex

### LINKED ISSUE #1344: Fix Google Docs spreadsheet embeds
At some point (likely within the last couple code pushes), we broke Google Docs spreadsheet embeds. I was able to replicate this issue, but it looks like text-type Google Docs embed fine. The original issue for this seems to be here: https://github.com/dreamwidth/dw-free/pull/573

http://www.dreamwidth.org/support/see_request?id=30609 is the Support request.

me-and:
The regex for allowing Google Spreadsheets to be embedded in posts should refer to docs.google.com/spreadsheets, which is the actual path used by Google.  Fix it so it does, and add a test to cover the behaviour.

Fixes #1344

### DIFF
--- a/cgi-bin/DW/Hooks/EmbedWhitelist.pm
+++ b/cgi-bin/DW/Hooks/EmbedWhitelist.pm
@@ -61,7 +61,7 @@ my %host_path_match = (
     "www.google.com"        => [ qr!^/calendar/!, 1 ],
     # drawings do not need to be whitelisted as they are images.
     # forms arent being allowed for security concerns.
-    "docs.google.com"       => [ qr!^/(document|spreadsheet|presentation)/!, 1 ],
+    "docs.google.com"       => [ qr!^/(document|spreadsheets?|presentation)/!, 1 ],
 
     "www.kickstarter.com"   => [ qr!/widget/[a-zA-Z]+\.html$!, 1 ],
 
--- a/t/embed-whitelist.t
+++ b/t/embed-whitelist.t
@@ -15,7 +15,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 54;
+use Test::More tests => 55;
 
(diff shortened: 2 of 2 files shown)


zorkian:
Both of the URLs in the test seem to work. Shouldn't we support both?

afuna:
Not sure what you mean about both URLs working  -- when I visit the /spreadsheet/ (old) version, it 404s.

zorkian:
`https://docs.google.com/spreadsheet/pub?key=0ArL0HD_lYDPadEkxSi1DTzJDa09GUmtzWEEwUDd4WFE&output=html&widget=true` redirects, it works fine.

afuna:
Ah gotcha. I'd gone to the new url, and just substituted "spreadsheet"
for "spreadsheets". That one didn't work, but the original form of the
URL does work.

Soooooo. in short I agree, let's support both please.


On Sun, May 31, 2015, at 04:23 PM, Mark Smith wrote:
> https://docs.google.com/spreadsheet/pub?key=0ArL0HD_lYDPadEkxSi1DTzJDa09GUmtzWEEwUDd4WFE&output=html&widget=true
> redirects, it works fine.


>



Links:

  1. https://github.com/dreamwidth/dw-free/pull/1408#issuecomment-107256753

me-and:
Huh. I was pretty sure I'd tested that and found it didn't work, but
evidently not. I'll fix up as soon as I'm in front of a computer rather
than my phone.

me-and:
Pull request updated to accept and test both `spreadsheet` and `spreadsheets`

afuna:
Great, thanks!

afuna:
Mmmph, maybe we should have the bot remove the "request changes" label when someone comments back with update or updated ;)

PR #1409, opened 2015-05-31
Fix typo in siteskins.blueshift.alt

### LINKED ISSUE #1407: siteskins.blueshift.alt contains typo
"siteskins.blueshift.alt=Blueshift: Black text ond white background; blue highlights" should read "on" not "ond"

kaberett:
Fixes #1407.

Unfortunately changing the name of the text string isn't an option:
it's referenced only from cgi-bin/LJ/Settings/SiteScheme.pm via

$scheme_alt_ml ||= "siteskins.$scheme->{scheme}.alt"
    if LJ::Lang::string_exists( "siteskins.$scheme->{scheme}.alt" );

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -3775,7 +3775,7 @@ siteskins.celerity.alt=Celerity: Black text on white background; olive color hig
 
 siteskins.celerity.desc=Default-sized sans-serif font; vertical, fixed, non-expanding menus.
 
-siteskins.blueshift.alt=Blueshift: Black text ond white background; blue highlights
+siteskins.blueshift.alt=Blueshift: Black text on white background; blue highlights
 
 siteskins.blueshift.desc=Vertical, fixed, non-expanding menus.
 


kaberett:
(after discussion with @anall, apparently my best bet at this point is to sacrifice a dictionary to the translation gods...)

woggy:
Staring at the referencing code and I *think* it's possible for a code change to support a new string. I can't quite figure out how to make it look Elegant with Perl postfix conditionals, though. (or whether we want to tread down the road of special-casing like this)

Currently:
```
$scheme_alt_ml ||= "siteskins.$scheme->{scheme}.alt" if LJ::Lang::string_exists("siteskins.$scheme->{scheme}.alt" );
```

Idea:
```
if LJ::Lang::string_exists("siteskins.$scheme->{scheme}.alt2") {
    $scheme_alt_ml ||= "siteskins.$scheme->scheme.alt2";
} else {
    $scheme_alt_ml ||= "siteskins.$scheme->{scheme}.alt" if LJ::Lang::string_exists("siteskins.$scheme->{scheme}.alt" );
}
```

rahaeli:
Actually, it looks like this string was never loaded onto the site proper, so you're good to just make the change in code without bumping the string. (I was going to go edit it live and then you could just change the text in git so future installs would have the correct string, but ... it ain't there.)

woggy:
Oh, well, then. Fortunate in this case. :)

rahaeli:
Yes! Your way would have worked, but probably a bit complex to put in the code when the simple hacky workaround will also work :)

woggy:
if only everything involving the translation system worked out so nicely.

afuna:
*makes the necessary warding signs against the translation system*

PR #1410, opened 2015-05-31
Remove redundant multisearch nav code

### LINKED ISSUE #1079: Investigate/remove dead code in views/multisearch.tt and cgi-bin/DW/Controller/Search/Multisearch.pm
So, probably as a result of https://github.com/dreamwidth/dw-free/commit/c292ef5d6f66b18849a284160afe4dceda537240 , it looks to me like /multisearch is no longer called with type=nav, and code around https://github.com/dreamwidth/dw-free/blob/develop/views/multisearch.tt#L16 and https://github.com/dreamwidth/dw-free/blob/develop/cgi-bin/DW/Controller/Search/Multisearch.pm#L200 is (I think) now dead and can be removed.

me-and:
c292ef5 seems to have removed all calls to `multisearch` that had
`type=nav`.  Certainly there aren't any such calls now, anyway.  Remove
code that's now redundant as a result.

Fixes #1079 

I've not tested this: I obviously can't test dead code, but I'd like to have done a basic sanity check that I've not broken anything else in this area, but I just can't find where this code is used on the site. Instead, I've discussed this with @anall on #dreamwidth-dev, who thinks the code looks fine (and has promised me a pony if it breaks things to boot).

### DIFF
--- a/cgi-bin/DW/Controller/Search/Multisearch.pm
+++ b/cgi-bin/DW/Controller/Search/Multisearch.pm
@@ -203,8 +203,7 @@ sub multisearch_handler {
     };
 
     # set up dispatch table
-    my $dispatch = { nav          => $f_nav,
-                     nav_and_user => $f_nav,
+    my $dispatch = { nav_and_user => $f_nav,
                      user         => $f_user,
                      int          => $f_int,
                      email        => $f_email,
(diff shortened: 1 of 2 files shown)


afuna:
Also happy with it. Hopefully no pony for you!

PR #1414, opened 2015-06-16
(Bug 1046) Convert Support High Scores Page to TT

hotlevel4:
Converts Support High Scores page to TT and Foundation. Related to issue 1046.


hotlevel4:
Redid the pagination to not use LJ::Paging, would appreciate a closer look at those changes.

afuna:
I think that looks good; you've switched to using the tt component and passed in all the expected data.

PR #1416, opened 2015-06-23
This line prevents ext/local/cgi-bin from overriding modules in cgi-bin

anall:
( this line is a bug as is, and getting this committed will make things easier for me until #1376 is merged )


zorkian:
Paths are hard.

anall:
And unfortunately this is not good enough. There will possibly be more incoming, depending on what the status of getting #1376 merged is ( we're blocked on the hack box currently, I believe ).

Issue #1355, opened 2015-04-24
OpenID logins from providers that use HTTPS URLs are broken


kareila:
There has been a recent increase in reports of OpenID login failures.  Upon closer inspection, I have realized a common thread between the different providers is that they have all switched to using HTTPS by default, so I believe the problem is SSL-related.

These are some of the OpenID providers that are currently not working:
- Twitter
- Tumblr
- Etsy
- Yahoo

(UPDATE: it has been rightly pointed out that Twitter, Tumblr and Etsy do not provide OpenID logins; that was a false conflation on my part.  Yahoo is still a legitimate problem, however.)

As reported here: http://www.dreamwidth.org/support/see_request?id=29682

The error message displayed when attempting to log in is "no_identity_server: Could not determine ID provider from URL".  This is the value of `$csr->err` set when the following statement is executed in /openid/login.bml (currently line 108):

`my $claimed_id = eval { $csr->claimed_identity($url); };`

Here `$csr` is a Net::OpenID::Consumer object.  I expect the fix will likely involve upgrading some of our installed OpenID and/or SSL modules, or we may need to implement a workaround.

I found this article helpful in researching the problem: http://ikiwiki.info/bugs/openid_login_fails_wirth_Could_not_determine_ID_provider_from_URL/

I tried the `$ENV{PERL_LWP_SSL_CA_PATH} = '/etc/ssl/certs';` workaround mentioned there, but it didn't seem to fix the problem.  They also mention `$ENV{PERL_LWP_SSL_VERIFY_HOSTNAME} = 0;` but that is already in ljlib.pl.

I also tried locally installing newer versions of the relevant modules, with no apparent change in behavior.

zorkian:
OpenID is a mostly deprecated protocol, AIUI, and I don't think people are really supporting it anymore. We might continue to have increasing trouble supporting it (read: hacks on hacks) if we do.

kareila:
I realize it's not a high priority, but I suspect this may be more of an SSL problem than an OpenID problem, and as such is worth investigating, especially if the fix turns out to be as simple as installing newer versions of some modules.

kareila:
Making a note to retest this soon.

kareila:
I just tried an OpenID login on prod and got a new but hopefully easier to diagnose error:

    Not an ARRAY reference at /home/dw/current/extlib/lib/perl5/Net/DNS/Paranoid.pm line 91.

kareila:
[...]

PR #1418, opened 2015-06-23
Fix SSL in get_useragent

### LINKED ISSUE #786: Feeds won't accept HTTPS URLs
I don't know if this has ever been fixed of if there's an issue with this particular URL but I can"t edit the URL of bruce-schneier-feed to https://www.schneier.com/blog/atom.xml although this feed is valid. I get an Invalid URL error message.
http://www.schneier.com/blog/atom.xml redirects to the https version but schneier_no_tracking_feed, which seems to use this alternate URL, won't update.

Request: http://www.dreamwidth.org/support/see_request?id=29076

anall:
Switching LWPx::ParanoidAgent to LWP::UserAgent::Paranoid
Fixes #786

( this may also fix the following bugs: #1355 )

( @kaberett this is JUST FOR YOU! )

### DIFF
--- a/bin/checkconfig.pl
+++ b/bin/checkconfig.pl
@@ -235,7 +235,6 @@ my %modules = (
                     deb => 'libio-aio-perl',
                     opt => 'Required for Perlbal',
                  },
-               "LWPx::ParanoidAgent" => { ver => '1.10' },
                "MogileFS::Client" => {
                    ver => '1.12',
                },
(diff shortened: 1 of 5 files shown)


anall:
( one more commit coming )

zorkian:
Okay; waiting for one more commit -- I assume to purge all the existing usage of `LWPx::ParanoidAgent`? (I assume there's stuff to remove.)

PR #1423, opened 2015-06-24
Add seperate header link color variables to all layouts

### LINKED ISSUE #677: Use color_header_link properties for links in journal headers
Having header links such as navigation links use module colors can cause readability issues. When links are displayed in the header they should use their own set of colors, distinct from module link colors as header and module can have different background colors.
I think this is implemented everywhere but in Bases and Practicality.
Note1: props in Core2 already exist. You don't need to create them.
Note2: users shouldn't see any changes if possible. You'll probably have to set the options to whatever colors were used instead. Nobody will blame you if you decide you don't want to do this for Practicality seeing as it has about 100 themes...

momijizukamori:
Adds variables for header link colors to Practicality and Bases
and adds them to existing themes, using the same colors that were
being used.

Note: this will not set custom colors if users are using them,
and they will have to set the new variables themselves in the wizard.

Fixes #677


zorkian:
I looked at this but I think this is all @afuna ... :)

PR #1424, opened 2015-06-24
Cleanup Routing API endpoints.

anall:
Clean up accidentally duplicated function.
Make lookup-routing work with API endpoints.


zorkian:
Thanks for finding this.

PR #1425, opened 2015-06-24
[#1419] Tweak font size/space when displaying fallback text

### LINKED ISSUE #1419: New Update Page degrades weirdly without Javascript
I've been informed that we're not supporting Javascript-less use on the new update page. However, since it's going to come up on my end, I wanted to file this anyway.
See below for picture of what is going on:
javascriptless new page

Version 41.0.2272.76 Built on Ubuntu 14.04, running on LinuxMint 17 (64-bit), no Javascript.

Of note: when using the same browser with Javascript on, the offset is greatly reduced but not eliminated. Woggy is replicating on his system.

afuna:
Fixes #1419.

### DIFF
--- a/htdocs/scss/pages/entry/new.scss
+++ b/htdocs/scss/pages/entry/new.scss
@@ -4,6 +4,11 @@
     text-align: left;
 }
 
+.no-js button.postfix.fi-icon--with-fallback {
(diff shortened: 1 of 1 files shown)


zorkian:
noideadog

PR #1426, opened 2015-06-24
[#1412] Add onchange for browsers that don't support onkeyup

### LINKED ISSUE #1412: Captcha failing on Android mobile platform
http://www.dreamwidth.org/support/see_request?id=31005 (Peterstein req)

The user wrote in complaining about /lostinfo not working on Android phones.

Steps to replicate:
On a Droid phone, go to www.dreamwidth.org/lostinfo
Answer CAPTCHA
The "proceed" button will not become clickable

I was able to replicate using my own Droid Maxx, but was NOT able to replicate on my friend's iPhone 6+. I do not have access to any other mobile devices, or I'd have checked on those too.

Possibly related: the captcha fails to load at all if Javascript is not enabled on a browser. This is not the exact issue that we're having here, but I figured I'd mention it just in case. I can also open a separate bug for that if needed.

afuna:
Fixes #1412.

### DIFF
--- a/htdocs/lostinfo.bml
+++ b/htdocs/lostinfo.bml
@@ -42,7 +42,7 @@ body<=
 
         $ret .= "<div class='form-basic form-fix'><div class='tl'><div class='tr'><div class='br'><div class='bl'>\n";
         $ret .= "<fieldset><label>$ML{'.enter_username'}</label>\n";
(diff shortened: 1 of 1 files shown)


afuna:
Apparently some mobile browsers don't support keyup (and don't plan on supporting it?). This falls back onto onchange so when they're done typing username, they can still recover info.

zorkian:
M-m-m-m-m-m-merged!

PR #1427, opened 2015-06-24
Fixes calendar module overflow in Leftovers. Fixes #829

### LINKED ISSUE #829: Calendar overflows in Leftovers when font is big enough
Typical issue caused when you have:

 #secondary > .inner:first-child {
padding: 1em;
}

(set in TR), modules styled as individual boxes and padding in .module-calendar TDs because the limits are bigger than the module box. This isn't an issue when there is no background because the overflow doesn't matter then. CSS and illusions of perfection...

If you refactor so the padding is set to 0 and the spacing is done differently the calendar will adapt\* (to a point; setting the sidebar width to a small value often makes it display badly no matter what you do).
- especially if you set the padding in em instead of px, which isn't the case here.

momijizukamori:


afuna:
Tweaking the padding seems reasonable, but why is the padding defined twice? I think the first just overrides the second entirely?

zorkian:
(Form letter incoming.)

PR #1442, opened 2015-06-25
Add footer text color variable for Planet Caravan

### LINKED ISSUE #1067: Tags label invisible in Planet Caravan/some themes
http://www.dreamwidth.org/support/see_request?id=29745

.tag-text or some parent element is black and the entry footer bkg is also black making the 'Tags:' label is invisible in Cheerfully, Oceanic and Sunrise, and users have no way of changing that color via the customize UI as there is no corresponding prop for it. I see
```
.entry .footer, .comment .footer {
    background: $*color_entry_footer_background;
    padding: .75em;
}
```

but no corresponding text color(s).

momijizukamori:
There were instances of black tag labels on black footers,
with no way of setting the tag label in the wizard. This adds a
footer text color, and sets it in layouts that had the issue.

Fixes #1067


styles/planetcaravan/layout.s2
```diff
@@ -41,6 +41,7 @@ set margins_unit = "%";
 
 propgroup colors_child {
     property Color color_entry_footer_background { des = "Entry footer background color"; }
+    property Color ccolor_entry_footer_text { des = "Entry footer text color"; }
```
afuna: psst ccolor_entry

momijizukamori: Okay, typo fixed and squashed into the previous commit

PR #1443, opened 2015-06-25
Switch links to https

### LINKED ISSUE #1006: Update links in site copy to point to HTTPS
If we don't do this immediately, it won't break anything because we'll have redirects in place, but... maybe nice? :)

Will comment with what grep shows me

me-and:
Wherever there are http links to elsewhere on Dreamwidth, use https instead.

Also remove some (apparently) unused strings that include http links, because there's no point updating them.

Along with the corresponding fix in dreamwidth/dw-nonfree, this fixes #1006.


CONTRIBUTING.md
```diff
@@ -3,10 +3,10 @@
 We love having new contributors join us! Here's the quick summary of where to find information on contributing to Dreamwidth:
 
 * Make sure you have a [GitHub account](https://github.com/signup/free)
-* Dreamwidth is a highly complicated web app with tons of dependencies. You don't need to get it working locally just to contribute: we do it all for you. You can do it if you want, but it's a lot easier if you let us do it! The easiest thing to do is to sign up for a [Dreamhack account](http://hack.dreamwidth.net/) -- our hosted developer service.
+* Dreamwidth is a highly complicated web app with tons of dependencies. You don't need to get it working locally just to contribute: we do it all for you. You can do it if you want, but it's a lot easier if you let us do it! The easiest thing to do is to sign up for a [Dreamhack account](https://hack.dreamwidth.net/) -- our hosted developer service.
```
zorkian: Does the Hack server actually work with HTTPS?

me-and:
...no. No it doesn't. I broadly assumed anything with `dreamwidth.(net|org)` would support HTTPS, but evidently that's not true. Will repatch imminently.

CONTRIBUTING.md
rahaeli: Not at the moment, no -- that should stay a http link, not https.

me-and:
New merge request now available.

Get it in quickly and @kaberett will need to add it to the code tour, and their rage amuses me...

anall:
The following strings are still being used. 

    widget.betafeature.updatepage.off
    widget.betafeature.updatepage.on
    widget.betafeature.updatepage.title

me-and:
Right. So the problem here is that I assumed `etc/config-local.pl` in dreamwidth/dw-nonfree is the version that's used on the live site. That's not actually the case, and the live site lists `updatepage` in `%LJ::BETA_FEATURES`. Normally I'd want to rename the key in the translation system, but the only obvious bit of the key to change is the `updatepage`, which would require a change at the same time to the live `config-local.pl`.

However @anall tells me that this change should be fine without changing the key, so I'll just go with that option.

PR #1454, opened 2015-06-27
Better status field in support history page

### LINKED ISSUE #1402: Status field on support history page less useful than it might be
The 'status' column on http://www.dreamwidth.org/support/history only distinguishes between 'open' and 'closed'. This is not very useful! It should use the same as the support board itself: 'open', 'answered, awaiting close', 'answered, still needs help', and 'closed'.

me-and:
As well as "open" and "closed", this will mean the list of tickets at <http://dreamwidth.org/support/history> also includes "awaiting close" and "still needs help" in the status column.

This fixes #1402, which @kaberett passed over to me for the actual coding bit. I've implemented it as @rahaeli requested, moving the code to determine this state to a function that can be called from elsewhere, and do that.

I haven't done the English stripping bit, because the entire files need English stripping (and/or converting away from BML), and I believe @kaberett is going to put that up somewhere as a good introductory job for a newbie dev.

### DIFF
--- a/cgi-bin/LJ/Support.pm
+++ b/cgi-bin/LJ/Support.pm
@@ -203,6 +203,20 @@ sub fill_request_with_cat
     $sp->{_cat} = $cats->{$sp->{'spcatid'}};
 }
 
+sub open_request_status {
+    my ($timetouched, $timelasthelp) = @_;
+    my $status;
+    if ($timelasthelp > $timetouched+5) {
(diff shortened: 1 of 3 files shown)


afuna:
Appreciate the cleanliness of the code. Also, I thought it wasn't possible to do this with the data we had on hand. I'm glad to be proven wrong.

rahaeli:
&hearts;

PR #1457, opened 2015-06-27
[#1451] Nest questions in an extra hash to avoid namespace conflicts

### LINKED ISSUE #1451: Site Options faqcat's faqs have gone missing
Lost: one faqcat.

http://www.dreamwidth.org/support/faq shows all the faq categories and their associated faqs. Since the code push, the faqs in the Site Options category are no longer displaying in the full faq list -- the category header displays, but not the actual faqs.

Things I have done to try to force a refresh: moving the faqs out of the category and then back in, changing the sortorder of the faqcat. 

@afuna thinks it might be related to the TT-ization.

afuna:
"site" is used as a reserved top-level variable name in TT pages, e.g.,
"site.root". So when we tried to use "site" (the category name) as a
top-level variable, it was overshadowed. This moves the categories off
the top-level to avoid namespace conflicts

Fixes #1451.

### DIFF
--- a/cgi-bin/DW/Controller/Support/Faq.pm
+++ b/cgi-bin/DW/Controller/Support/Faq.pm
@@ -87,7 +87,7 @@ sub faq_handler {
             next unless $q;
             $q =~ s/^\s+//; $q =~ s/\s+$//;
             $q =~ s!\n!<br />!g;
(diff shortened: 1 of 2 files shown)


afuna:
oops I want this on the hotfix. closing temporarily

PR #1460, opened 2015-06-27
Fix OpenID login error on live site

kareila:
Per discussion in #1355, trying to log in with OpenID in production currently
results in the following error:

    Not an ARRAY reference at /home/dw/current/extlib/lib/perl5/Net/DNS/Paranoid.pm line 91.

My understanding of the problem is that LWP::UserAgent::Paranoid has a different argument
format for its blocked_hosts method than LWPx::ParanoidAgent did, and the arguments need
to be enclosed in an arrayref.  I can't reproduce the error exactly in testing on my
development environment, but I believe this will at least get us back to where we were
before, where at least non-SSL OpenID logins would still work.


afuna:
Okay, going to get this live tonight.

PR #1474, opened 2015-07-01
[#1468] check for SKIP_PARANOID_USERAGENT instead of IS_DEV_SERVER

### LINKED ISSUE #1468: let LJ::OpenID use the paranoid user agent on dev servers
Related to #1418: when we switched modules from LWPx::ParanoidAgent to LWP::UserAgent::Paranoid, there were some API changes which made it to the live site undetected because LJ::OpenID does not use the paranoid user agent on dev servers. The error was fixed in #1460, but in the future, I believe such heartache could be avoided if we removed the dev server special casing.  The split dates from LJ days and I haven't seen any good arguments in favor of keeping it.

kareila:
For certain LJ::OpenID functions that require a UserAgent object,
use LWP::UserAgent::Paranoid instead of the default LWP::UserAgent,
unless $LJ::SKIP_PARANOID_USERAGENT is enabled in the site config.

Fixes #1468.

### DIFF
--- a/cgi-bin/LJ/OpenID.pm
+++ b/cgi-bin/LJ/OpenID.pm
@@ -64,7 +64,7 @@ sub consumer {
     my $get_args = shift || {};
 
     my $ua;
-    unless ($LJ::IS_DEV_SERVER) {
+    unless ( $LJ::SKIP_PARANOID_USERAGENT ) {
         $ua = LJ::get_useragent( role => "OpenID",
                                  timeout => 10,
                                  max_size => 1024*300, );
@@ -193,7 +193,7 @@ sub hmac {
 sub blocked_hosts {
     my $csr = shift;
 
-    return do { my $dummy = 0; \$dummy; } if $LJ::IS_DEV_SERVER;
+    return do { my $dummy = 0; \$dummy; } if $LJ::SKIP_PARANOID_USERAGENT;
 
     my $tried_local_id = 0;
     $csr->ua->blocked_hosts( [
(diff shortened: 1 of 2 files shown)


zorkian:
Did you consider using a library to detect RFC 1918 IPs and just disabling it in that case? Or is that too magical?

kareila:
I didn't consider that approach, but I think the principle of least surprise should apply here, and the code should behave consistently unless an exception is requested. An unexpected special case is what got us here in the first place. :)

zorkian:
I hate options, I'm not convinced this should be an option. Why not just make dev servers not use Paranoid agent, and make it print something to the Apache error log when it hits that code path, so the developer will notice it?

kareila:
I guess that's a matter of opinion. If the dev servers aren't using the paranoid agent, I can't reproduce or debug the problems we have with it.

afuna:
But are we even sure that dev servers need to avoid Paranoid Agent? The only case that's been brought up was hypothetical.

kareila:
No, that was the hypothetical of whether the paranoid agent should ever legitimately NOT be used.

afuna:
Right sorry I meant the other way around -- are we sure we need to disable paranoid agent in this case :)

kareila:
The simplest thing would be to just remove the special case. That's what I was going to do until @alierak mentioned the hypothetical scenario where it might be useful.

zorkian:
If it's only hypothetical, let's not add an option please!

PR #1475, opened 2015-07-01
[#1453] FormHTML functions automatically ehtml their arguments

### LINKED ISSUE #1453: site search over-escaping characters
Go to dreamwidth.org/search and enter something with quotes in it (or any other characters treated as needing escaping). Hit 'search'. The text in the search box is then escaped.

For instance, searching for "and the" (with the quotes) turns into &amp;quot;and the&amp;quot;. And if you hit 'more results', it becomes &amp;amp;quot;and the&amp;amp;quot;. Etc.

kareila:
The BML version of this page used hand-coded HTML which necessitated
manually escaping the displayed query text using LJ::ehtml. However,
the TT version uses form.textbox and form.hidden, which escape their
values automatically.  Using both together resulted in doubly escaped
text, so we should no longer manually escape the query text.

Fixes #1453.

### DIFF
--- a/views/search.tt
+++ b/views/search.tt
@@ -44,8 +44,7 @@
     %]
 
     <br>
-    [% eq = q | html;
-       form.textbox( name = "query", maxlength = "255", size = "60", value = eq );
+    [% form.textbox( name = "query", maxlength = "255", size = "60", value = q );
        " ";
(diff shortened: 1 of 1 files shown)


zorkian:
Nice catch! :)

PR #1476, opened 2015-07-01
[#1468] let LJ::OpenID always use the paranoid useragent

### LINKED ISSUE #1468: let LJ::OpenID use the paranoid user agent on dev servers

kareila:
This removes the IS_DEV_SERVER special casing that resulted in the
paranoid useragent only being used in production for some functions
of LJ::OpenID.

The bypass for the blocked_hosts check is left intact as a comment,
in case it is needed for testing purposes at a later date.

Fixes #1468.

### DIFF
--- a/cgi-bin/LJ/OpenID.pm
+++ b/cgi-bin/LJ/OpenID.pm
@@ -63,12 +63,10 @@ sub server {
 sub consumer {
     my $get_args = shift || {};
 
-    my $ua;
-    unless ($LJ::IS_DEV_SERVER) {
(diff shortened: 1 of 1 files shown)


zorkian:
Thank you!

PR #1487, opened 2015-07-13
[#1483] explicitly specify UTF-8 charset

### LINKED ISSUE #1483: compass compile errors on Unicode chars
From http://dw-dev-training.dreamwidth.org/59889.html (thanks @ninetyd)

After upgrading the dreamhack server (such that Ruby is now >= 1.9), compass compile is producing errors like:
`error htdocs/scss/components/pickadate/time.scss (Line 6: Invalid US-ASCII character "\xE2")`

Hopefully setting the first line of affected files to this will help:
`@charset "UTF-8"`

as described at http://sass-lang.com/documentation/file.SASS_REFERENCE.html#encodings

kareila:
If a developer's LANG environment variable uses something other than
a version of UTF-8, compiling SCSS files that contain Unicode characters
will fail with invalid character errors.  The easiest and best way to
fix this is to explicitly specify the expected UTF-8 charset on the
first line of each affected file.

Fixes #1483.


zorkian:
Thanks for tracking that down!

kareila:
I did very little, @alierak deserves all the credit for diagnosing the issue and proposing the fix and @ninetyd brought the problem to our attention.

PR #1490, opened 2015-07-14
Limit the old form of the icon url recognized by the contextualhover

afuna:
We should only recognize the old dreamwidth url, not one for any other
site


afuna:
This could conceivably be made generic by using Site.siteroot -- but:

a) JS doesn't allow you to ignore special characters. You'd have to write a regexp to convert those periods into back-slashed periods. Meh

b) Site.siteroot includes http which brings us to http versus https

c) This is pretty much an edge case handilng site-specific user behavio

rahaeli:
Yeah, I'm okay with not genericizing it given the constraints.

kareila:
Works for me! Thanks :heart:

PR #1494, opened 2015-07-17
Fixes# 1467.

dteklavya:
Replaced /support/submit.bml with /support/submit. Also verified that the link works.

### DIFF
--- a/bin/upgrading/deadphrases.dat
+++ b/bin/upgrading/deadphrases.dat
@@ -1596,3 +1596,5 @@ general widget.search.msn
 general /manage/profile/index.bml.chat.msnusername
 general /profile.bml.im.msn
 general /profile.bml.label.msnusername
+
+general /site/opensource.bml.text.development.directions3
--- a/htdocs/site/opensource.bml
+++ b/htdocs/site/opensource.bml
@@ -38,7 +38,7 @@ body<=
 
 <p><?_ml .development.main2 _ml?></p>
 
-<p><?_ml .development.directions3 _ml?></p>
+<p><?_ml .development.directions4 _ml?></p>
 
 <=body
 windowtitle=><?_ml .title _ml?>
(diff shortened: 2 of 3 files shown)


rahaeli:
So, this patch would be perfect, except for one little thing: it touches the translation system. (Any of the files that end in .text are the strings for the translation system. The translation system is horrible. I can't even defend it, it's the *absolute worst*, I am so sorry.) 

Because the translation system is absolutely horrible, you can't just change a string in the .text file and have it change on the live site. (There is a long and boring explanation, but it basically boils down to: because some strings are changed on the live site using the admin tool for the translation frontend, we don't let any changes to the .text files override the strings on the live site, because there's no way to tell which one of the two versions of the string is more recent. The translation system is horrible. I am so sorry.)

So, in order to make a change like this, you can't just edit the string. Instead, you have to:

* Change the name of the string, in this case, ".development.directions3" -- you'll notice there's already a 3 on the end of it. That's because that string has gone through this at least twice before. Convention is to stick the next highest number on the end, so you'd change the name to .development.directions4.

* Put the name of the old string name in bin/upgrading/deadphrases.dat. (This will remove the old version from the live site.)

* Change the reference to the string name in the code that calls it, which in this case is htdocs/site/opensource.bml, to your new .development.directions4 string.

* Then you can finally make the change to your new .development.directions4 string to remove the .bml from the link.

(I'm sorry I missed that this was that kind of bug or I would have made a note in the bug itself!)

zarhooie:
While fixes are still needed, please know that I appreciate the heck out of
you. Thank you!

bin/upgrading/deadphrases.dat
```diff
@@ -1596,3 +1596,5 @@ general widget.search.msn
 general /manage/profile/index.bml.chat.msnusername
 general /profile.bml.im.msn
 general /profile.bml.label.msnusername
+
+general /site/opensource.bml.text.development.directions3
```
kareila: Almost there! This string reference needs to be `/site/opensource.bml.development.directions3`, not `/site/opensource.bml.text.development.directions3`.  If you'll remove the .text portion I can get this in!

kareila:
I corrected the string name and merged this in manually.  Thanks for working on it!

PR #1495, opened 2015-07-18
"Use current time" unchecked by default when editing

### LINKED ISSUE #1480: Beta Update page: new "use current time" preference persists when editing entries
http://www.dreamwidth.org/support/see_request?id=31185

The preference should not apply when editing existing entries, only when creating new entries.

@kaberett expressed interest in working on this with @woggy.

kaberett:
Fixes #1480.

The sticky-ticky remembering whether a user has checked the
"use current time" option on the Create Entries beta should not
be checked by default when editing entries (or Christmas happens
in June).

Also one typo fix because I couldn't help myself.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -257,7 +257,7 @@ sub new_handler {
 
 
 # Initializes entry form values.
-# Can be used when posting a new entry or editing an old entry. .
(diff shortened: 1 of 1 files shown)


kaberett:
Couldn't think of a tidier way to do it; willing to be told there is one! Thanks to @woggy for the hand-holding :-)

kareila:
This seems to do the right thing in a reasonable manner!  Thank you for fixing it :smile:

PR #1498, opened 2015-07-24
Refactor text on beta features page

kareila:
- Moved existing strings from en.dat to views/beta.tt.text. Confirmed
they were not referenced elsewhere in the code before doing so.

- Added referenced but undefined strings for `.betafeature.${handler.key}.cantadd`.

- Corrected placement of period after closing paragraph tag.


kareila:
Lightly tested; will self-commit if someone else is willing to do a sanity check.

afuna:
I do not believe that these text strings are used anywhere else !

kareila:
Thanks for looking :heart:

PR #1503, opened 2015-08-01
(Issue 1046) Convert Support History to TT + Foundation

hotlevel4:
English-strips the page and converts it to TT and Foundation.


kareila:
Thanks for doing this!  I love that you put in the extra effort to restyle the search form using the Foundation toolkit, and the rest of it looks like a pretty straightforward TT conversion.

I do have a few remarks for things that should be tweaked before I can merge this, but nothing major. I'll drop those in shortly, and when you've addressed them, leave a comment letting me know I should look at this again.  Thanks!

views/support/history.tt.text
```diff
@@ -0,0 +1,49 @@
...
+
+.table.state=State
+
+.table.answeredby=Answered By
+
+.table.openedby=Opened By
+
+.table.timeopened=Time Opened
+
+.viewing=Viewing support requests for [[username]]
```
kareila: Your history.tt.text file needs to have its keys in alphabetical order, for easy reference.

cgi-bin/DW/Controller/Support/History.pm
```diff
@@ -0,0 +1,175 @@
...
+                    answered => $answered,
+                    catname => $cats->{$reqs{$id}->{spcatid}}{catname},
+                    openedby => $reqs{$id}->{requserid} ? LJ::ljuser($us->{$reqs{$id}->{requserid}}) : $reqs{$id}->{reqemail},
+                    timeopened => LJ::mysql_time( $reqs{$id}->{timecreate} )
+                };
+            }
+        } else {
+            $vars->{noresults} = 1;
+        }
+    } else {
```
kareila: This else block should only be executed if $fullsearch is false, instead of just the final statement.  e.g.:
```
} elsif ( ! $fullsearch ) { ... }
```
(I realize this is sacrificing a bit of readability by changing an 'unless' to an 'if !', but it's the easiest way to do it, I think, given that Perl doesn't have an 'elsunless' ;)

cgi-bin/DW/Controller/Support/History.pm
```diff
@@ -0,0 +1,175 @@
...
+                if ($reqs{$id}->{state} eq 'closed' && $reqs{$id}->{winner}) {
+                    $answeredby = LJ::ljuser( $reqs{$id}->{winner} );
+                    $points = $reqs{$id}->{points};
+                    $answered = 1;
+                }
+                $ret .= $reqs{$id}->{requserid} ? LJ::ljuser($us->{$reqs{$id}->{requserid}}) : $reqs{$id}->{reqemail};
+                $ret .= "</td>\n<td align='center'>";
+                $ret .= LJ::mysql_time($reqs{$id}->{timecreate});
+                push @{ $vars->{reqs} }, {
+                    id => $id,
```
kareila: I don't think your 'id' value is actually used in the template, so it should probably be removed here.

cgi-bin/DW/Controller/Support/History.pm
```diff
@@ -0,0 +1,175 @@
...
+                    reqemail => LJ::ehtml( $row->[8] ),
+                };
+                push @userids, $row->[4] if $row->[4];
+            }
+            my $us = @userids ? LJ::load_userids( @userids ) : undef;
+
+            # get categories
+            my $cats = LJ::Support::load_cats();
+
+            $ret .= '<table><thead><tr><th>Summary</th><th>State</th><th>Answered By</th>' .
```
kareila: These $ret lines are an artifact from the BML file and aren't used in the template output.  To reduce confusion, please remove all lines assigning to $ret from the controller as well.

cgi-bin/DW/Controller/Support/History.pm
```diff
@@ -0,0 +1,175 @@
...
+            $reqlist = $dbr->selectall_arrayref( 
+                'SELECT spid, subject, state, spcatid, requserid, ' .
+                       'timecreate, timetouched, timelasthelp, reqemail ' .
+                'FROM support WHERE reqtype = \'user\' AND requserid = ?',
+                undef, $userid );
+        } elsif ( $get->{email} ) {
+            # try by email, note that this gets requests opened by users and anonymous
+            # requests, so we can view them all
+            my $email = LJ::trim( $get->{email} );
+            $vars->{email} = $email;
```
kareila: You dropped an important `LJ::ehtml` call here.  User input must always be escaped before being redisplayed.
```
$vars->{email} = LJ::ehtml( $email );
```
It's also possible to do the html escaping in the template, but it's probably simpler to do it here.

hotlevel4:
Should be fixed. (I think!)

I meant to do the html escape for the email in the template and couldn't get it to work right, then forgot to add it back into the controller. For future reference, how would I do that in the template?

kareila:
My TT knowledge is not comprehensive and there might be a less verbose way, but the way I know would work is to save the html-escaped email as a new variable, and use that as the argument to the function.
```
[% ELSIF get_email %]
    [% escaped_email = email | html %]
    <h3>[% '.viewing' | ml(username = escaped_email) %]</h3>
[% END %]
```

PR #1505, opened 2015-08-01
Convert rename-related pages to Foundation

### LINKED ISSUE #972: Convert rename.tt and rename/swap.tt to Foundation

### LINKED ISSUE #1504: [Safari+JS] problem with input focus on /rename page
If I visit dreamwidth.org/rename and try to input text into the "Rename token" text field, the focus immediately jumps up to the "Desired username" field.  Using control-click to paste text is the only workaround I've figured out so far.

I'm only able to reproduce this in Safari.  If I use Firefox or Chrome, the focus doesn't jump away.  I haven't checked to see if earlier versions of Safari are affected, but the same thing seems to happen in Safari on my iPad running iOS 8.3 as well.

There's no special JavaScript being loaded on the page, just the default site-wide stuff.

kareila:
This fixes #972 by converting the two pages named there, but also includes all the rename admin pages as well, since most of them used the same error handling mechanism, which I updated to use DW::FormErrors.

This is lightly tested; I checked the default display of most of the pages, and generated a few error messages to make sure they appeared as expected.

This series of changes also fixes #1504, which I uncovered in the process of testing.


cgi-bin/DW/RenameToken.pm
```diff
@@ -109,7 +109,7 @@ Verifies if this could be a valid format for the rename token. Checks length and
 sub valid_format {
     my ( $class, %opts ) = @_;
 
-    my $string = uc $opts{string};
+    my $string = defined $opts{string} ? uc $opts{string} : '';
```
zorkian: Or,

```
my $string = uc($opts{string} // '');
```

cgi-bin/DW/User/Rename.pm
```diff
@@ -57,34 +58,36 @@ Return true if this user can be renamed to the given username
 
 Optional arguments are:
 =item force     => bool, default false
-=item errref    => array ref of errors
+=item errors    => DW::FormErrors object
```
zorkian: Slight preference to `errobj` in these kinds of things, just so it's clear it's an object. (You don't have to change things unless you want, but just a note...)

htdocs/scss/pages/rename.scss
```diff
@@ -25,15 +20,6 @@
...
-}
-
-ul.error-list li { 
-    list-style: disc outside;
-}
-
-p.detail {
-    margin-left: 2em !important;
+#content p.note {
+    font-style: italic; font-size: 80%;
```
zorkian: I believe the style is to have a newline after semicolons, so they're on separate lines?

views/admin/rename_edit.tt
```diff
@@ -13,16 +13,11 @@ the same terms as Perl itself.  For a copy of the license, please reference
...
 
-[% IF error_list %]
-    <p>Error</p>
-    <ul class='error-list'>
-        [% FOREACH error = error_list %]
-            <li>[% error %] </li>
-        [% END %]
-    </ul>
-[% END %]
+[%- dw.need_res( { group => "foundation" }
```
zorkian: I can't recall off-hand, but is there really supposed to be no trailing comma here?

afuna: The comma is optional there in TT.

htdocs/scss/pages/rename.scss
kareila: You're right, I missed that when pasting it from somewhere else.

views/admin/rename_edit.tt
zorkian: Aren't you on vacation? ;)

kareila: She loves us too much to stay away entirely :wink:

zorkian:
LGTM nothing blocking this from merging, but I won't hit the button given the style nit. Once you're happy with it go ahead and merge etc.

Thanks for doing this! Yay forward progress on our conversions :)

views/admin/rename_edit.tt
afuna: I am not even writing any code! Just... keeping my hand in a bit :)

kareila:
Thanks for the quick review!

views/admin/rename_edit.tt
rahaeli: We are like the Hotel California. You can check out anytime you'd like ...

PR #1506, opened 2015-08-01
Fixes #1463. Adds external site support for lj.rossia.org

### LINKED ISSUE #1463: add lj.rossia.org to username tag
For username 'username':

Journal: http://lj.rossia.org/users/username/
Profile: http://lj.rossia.org/userinfo.bml?user=username

(It's very oldstyle LJ URLs, before subdomaining.)

pseudomonas:
Added a DW::External::Site::LJRossia.pm ; updated DW::External::Site.pm to cope with three-part URLs (only if no match for two-part URL) and to link the LJRossia module.


rahaeli:
Hi pseudomonas! Do you have a Contributor Licensing Agreement (CLA) on file? (If you do, drop me an email privately to remind me what the name you filed it under was -- denise@dreamwidth.org.) If you don't, we'll need a signed copy of the agreement in order to accept your contribution. You can see the instructions here: http://wiki.dreamwidth.net/wiki/index.php/CLA

rahaeli:
Aha! Returning contributor whose GH username I did not have connected with the CLA on file. Sweet; welcome back!

kareila:
Thanks for doing this - seems fine on first look, but I can't accept it as is because of the config.pl additions. Can you remove those and resubmit?

It's also traditional to update the header of any new modules to include your own name and email address, but that's your call.

kareila:
Upon closer testing of this, I've realized that the journaltype lookups are failing because you've given the new site the servicetype of "ljr" instead of "lj". Also, we've done local caching of the userhead types for the other LJ-based sites in the htdocs/img/external/ directory, to avoid sending traffic to the remote site every time the userhead is displayed - please consider doing the same here.

pseudomonas:
OK, will fix. Is their codebase similar enough to *modern* lj that the lj servicetype will do the right thing?

rahaeli:
It should -- that identifier just describes how usernames are constructed, IIRC.

pseudomonas:
I thought the issue with this site was that it was constructing them differently (old-style), though?

rahaeli:
I am *pretty* sure it should still work! It worked for Journalfen, which also used old-style URLs.

pseudomonas:
Is this any better? Still getting used to git...

kareila:
This does look better!  I am still testing - I think both of the existing DW::External::Userinfo checks are failing due to site differences, but at least the relevant code is executing now :) I will try to get that fixed later tonight.

One thing I would love if you are feeling up to it, is to check out your branch and do a `git rebase -i develop` to squash this into one commit.  Here's a page that outlines what I mean by that: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Once the commits are squashed, you'll need to do a `git push --force` to update your pull request.

pseudomonas:
Squashed (I think!)

kareila:
Thanks so much!  This is good to go now! :+1:

pseudomonas:
Yay! Baby's first bug squished!

Issue #1464, opened 2015-06-29
admin tool to reset post-by-email token


rahaeli:
We had a case come up of "user's email account is emailing in spammy comments as reply-by-email" -- what probably happened there was that someone compromised the email account and harvested all the email addresses out of it. 

After a bit of thought, I realized the only way to stop the spam would be to reset the post-by-email secret token. (Even if the user resecures their email account, the addresses have already been harvested, and all the spammer has to do is forge their email address as the From: on the message in the future. They don't need ongoing access to the email account in order to do that, so just resecuring the account doesn't necessarily stop the spam.)

Right now, the only way to reset a user's PBEM token for them is to impersonate them and hit the button while logged in as them. So, since we try to restrict uses of impersonate to the absolute bare minimum, we should have an admin console command that will let a site admin reset the token for someone.

It's a relatively low-impact action to take -- resetting someone's PBEM token wouldn't require any access to user data -- so I don't think we need to restrict it to one of the more restricted privs or create a new priv for it. Instead, let's hang it off the same priv that we already use to check whether someone can use the reset_email console command: the reset_email priv. I don't think we even need a sub-priv of it; they're highly similar concepts.

Let's make it a new console command called reset_token, hung off the reset_email priv, and it should take two args:

`reset_token <username> <reason>`

The text supplied for 'reason' should be logged into statushistory. Running the command should generate a new random token. It should also send an email to the user, from 'antispam@dreamwidth.org' and a subject line of "Your Dreamwidth reply-by-email token has been reset":

> Dear username,
>
> A $sitename administrator has reset the secret token for your account to use to post comments by email. We had to reset the token, because your account was being used to email spam to the unique reply-to address for one or more comment notifications you received.
>
> Usually this happens because someone got access to your email account and 'harvested' the addresses from messages that were in your inbox. We strongly suggest that you change the password for your email account. We don't know for sure that a spammer broke into your account, but that's usually how a situation like this happens, and it's better to be safe than sorry.
>
> Any old comment notification emails you have in your inbox that were sent to you before your token was changed will no longer work to reply by email. You'll need to follow the links in those notification emails to reply directly on the website. Any comment notification email sent to you after the token was changed will work as usual.
>
> If you have any questions, you can reply to this email.
>
> Best,
>  The $sitename Team

(You'll have to define a new $ANTISPAM_EMAIL config option and set it to antispam@dreamwidth.org so we don't get misdirected email from daughter sites using this command.)

[...]

PR #1507, opened 2015-08-03
Reset pbem token - Fixes bug #1464

pseudomonas:
This should fix bug 1464.

### DIFF
--- a/cgi-bin/DW/Console/Command/ResetPBEMToken.pm
+++ b/cgi-bin/DW/Console/Command/ResetPBEMToken.pm
@@ -5,7 +5,7 @@
 # Console command for resetting PBEM tokens
 #
 # Authors:
-#      Adam Bernard <user name="pseudomonas" site="dreamwidth.org"">
+#      Adam Bernard <https://pseudomonas.dreamwidth.org>
 #
 # Copyright (c) 2015 by Dreamwidth Studios, LLC.
 #
@@ -45,7 +45,7 @@ sub execute {
 
     my $newauth = $u->generate_emailpost_auth();
 
-    return $self->print( "Token not reset" ) unless $newauth;
+    return $self->error( "Token not reset" ) unless $newauth;
 
     my $sitename = $LJ::SITENAME;
     my $rv = LJ::send_mail( {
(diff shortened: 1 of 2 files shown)


kareila:
Hey @pseudomonas, I wanted to let you know that we ask our contributors to follow the syntax outlined in https://help.github.com/articles/closing-issues-via-commit-messages/ which is picky about what it accepts. "This should fix #1464" works for making sure the issue gets assigned to you and closed when the PR is merged, but the similar syntax you used above doesn't. Sorry for the trouble :)

cgi-bin/DW/Console/Command/ResetPBEMToken.pm
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/perl
+#
+# DW::Console::Command::ResetPBEMToken
+#
+# Console command for resetting PBEM tokens
+#
+# Authors:
+#      Adam Bernard <user name="pseudomonas" site="dreamwidth.org"">
```
kareila: Don't use the user tag syntax here. The URL syntax you used in your previous PR is fine, though!

cgi-bin/DW/Console/Command/ResetPBEMToken.pm
```diff
@@ -0,0 +1,85 @@
...
+    return $self->error("This command takes two arguments. Consult the reference.")
+        unless $username && $reason && scalar(@args) == 0;
+
+    my $u = LJ::load_user($username);
+    return $self->error("Invalid user $username")
+        unless $u;
+
+    my $newauth = $u->generate_emailpost_auth();
+
+    return $self->print( "Token not reset" ) unless $newauth;
```
kareila: Should this be `return $self->error` instead of `return $self->print`?

kareila:
This seems to work fine! If you'll address the two minor issues I brought up, I'll get this merged. Thanks!

cgi-bin/DW/Console/Command/ResetPBEMToken.pm
rahaeli: Yes, please use $self->error, it prints differently and someone (*raises hand*) is a Bear of Very Little Brain and does not always notice that something didn't succeed unless it's printed in error styling. :)

kareila:
Sorry, I just saw you updated this! Adding a commit to an open PR doesn't fire off a notification, sadly!

PR #1511, opened 2015-08-04
[#1502] in setTimeToNow(), don't multiply minutes by 100

### LINKED ISSUE #1502: Timestamp weirdness: multiplying minutes by 100 then truncating causes strange behaviour
Dre has diagnosed the problem (see comments). Apparent lack of in-code comments, however, means it is not immediately apparent why we're doing this or what will break if we stop. Nonetheless it should probably be fixed.

---

We've had several reports now of weird stuff going on with timestamps for people using the "Use time entry is posted" checkbox on the Create Entries beta since the sticky-ticky was introduced.

Support request: " Sometimes (possibly related to the above, as it happens every time I compose two posts one after the other) a post will get a future time stamp. I think, but I'm not sure, that it gets marked into the future by the same interval as it took me to compose the post. (i.e. if I started composing a post at 12:00 and posted at 12:45, the post will get tagged with 13:30.)" (http://www.dreamwidth.org/support/see_request?id=31153)

Conversation in IRC: 
07:05 < AlexSeanchai> see http://alexseanchai.dreamwidth.org/946340.html which says posted at 01:50 am? I posted it at 01:05.
07:06 < AlexSeanchai> right before I clicked post, the timestamp box of course had the time at which I opened the post window. which was actually 01:02, but the box said 01:20.
07:06 < AlexSeanchai> single-digit minute numbers are showing in the tens column instead of the ones column.
07:06 < AlexSeanchai> how to fix is beyond me.
07:09 < AlexSeanchai> ...that is to say, I figured out *a* problem with my freaking timestamps. I was gonna post another post to test, editing after 1:10 hits, but the timestamp box on the update page says 02:30. dunno where that came from.

I've also observed some weirdnesses but haven't had the cope/brain to work out what on earth is going on. Filing per @rahaeli's comment over in dw-dev in the interests of diagnosing what the problem is.

kareila:
The lion's share of the credit for this fix goes to AlexSeanchai
for pattern recognition and exor674 for diagnosing.

Fixes #1502.

### DIFF
--- a/htdocs/js/pages/entry/new.js
+++ b/htdocs/js/pages/entry/new.js
@@ -579,7 +579,7 @@ var postForm = (function($) {
                         ].join("-");
             $("#js-entrytime-date").val(date).trigger("change");
 
-            var time = [ zeropad(now.getHours()), zeropad(now.getMinutes() * 100) ];
+            var time = [ zeropad(now.getHours()), zeropad(now.getMinutes()) ];
             $("#js-entrytime-time").val(time).trigger("change");
 
             $("#js-trust-datetime").val(1);


kaberett:
HURRAH. Will make sure they get namechecked in the code tour :-)

zorkian:
TO THE PUSH WITH YOU

PR #1513, opened 2015-08-05
Refactoring poll code

pseudomonas:
First stab at refactoring. The code hasn't been beautified, just moved out into different modules so that there's (hopefully) no if `($question->type eq "tickybox") { ...` left in `Poll.pm` or `Question.pm`


cgi-bin/LJ/Poll/Question.pm
```diff
@@ -15,6 +15,26 @@ package LJ::Poll::Question;
...
+
+### I don't quite see why I should need to do the use here, but it does
+### seem to be necessary - asb
+use LJ::Poll::Question::CheckBox;
+use LJ::Poll::Question::RadioButton;
+use LJ::Poll::Question::MultiChoice;
+use LJ::Poll::Question::Scale;
+use LJ::Poll::Question::DropDown;
+use LJ::Poll::Question::Text;
+
```
kareila: This is common and annoying enough that we have a shortcut:

`LJ::ModuleLoader->require_subclasses( "LJ::Poll::Question" );`

etc/config.pl
```diff
@@ -349,7 +349,7 @@
             google_analytics => 0,
             'inbox_max' => 2000,
             'interests' => 150,
-            'makepoll' => 0,
+            'makepoll' => 1,
```
kareila: I'm sure you needed to change this for your testing, but it probably shouldn't be included in the pull request.

rahaeli: Always a good idea to put etc/config*.pl in your .gitignore! It means manually updating the configs if we add a new config variable, but it prevents having to be paranoid about what changes creep in.

kareila:
I'm really excited that you did this! I left a few style comments based on my initial impressions, and will take a closer look at the functional changes on my next pass.

etc/config.pl
pseudomonas: aha, the gitignore was something I didn't know about. Good plan!

pseudomonas:
@kareila thanks so much for your comments! The refactoring has actually made it really quite easy to make a new poll type by creating a new subclass of `LJ::Poll::Question::MultiChoice` so I think it's served its purpose for me at least.

pseudomonas:
Oh, bottoms. The push of the latest code (implementing runoff polls) should probably be separated from the refactoring per se.

pseudomonas:
I'm still getting used to git :disappointed:

zorkian:
(Form letter incoming.)

Issue #1455, opened 2015-06-27
No version control on the FAQ system


me-and:
Apparently there is no version control on the FAQ system. This is the sort of thing that makes me sad, and I'd quite like to fix it.

I expect this isn't trivial, but I can't imagine it being *too* horrific...

azurelunatic:
Oh god why do you *say* things like that?!

me-and:
I should probably claim this...

me-and:
Right, I have the outline of a solution, which is basically to implement [this Stack Exchange solution](http://stackoverflow.com/a/323109). Short version goes:

* Add a new table `faqhist`, which has all the fields the `faq` table has, plus `revision`, where the `revision` field is an additional index field. (Timestamps are also useful but they're already present in that table.)
* Every time an FAQ entry is created or edited, add it to the `faqhist` table as well as the `faq` table, incrementing the `revision` from the previous entry with the matching index fields.
* Bob's your uncle/aunt/appropriate gender-neutral relative.

rahaeli:
I leave this quote from the qdb as being relevant: http://qdb.dreamwidth.net/dw/72

zarhooie:
Kab and Adam have both been prepped on the signs of being possessed by an Elder God, or whatever else may be shambling in the deep, dark depths of the FAQ system.

azurelunatic:
So when the badgers go all wrong and someone manages to save some FAQ as
blank, what's the rollback process?

me-and:
The rollback process will just be pulling the data out of the database manually and copying it into place. Once the information is being recorded, though, it'll be possible to add a web interface into the table to view old entries, restore them, view diffs, etc.

me-and:
There, see, not too horrific at all!

rahaeli:
I really, really misread 'faqhist'. ;)

zorkian:
Right?!

--
Mark Smith
mark@qq.is
 
 
On Thu, Aug 6, 2015, at 05:56 PM, Denise Paolucci wrote:
> I really, really misread 'faqhist'. ;)


> —
> Reply to this email directly or view it on GitHub[1]. 


>  


Links:

  1. https://github.com/dreamwidth/dw-free/issues/1455#issuecomment-128551762

me-and:
...it took me far longer than I'd like to admit to work out what you could have possibly misread it as ><

Want me to change it?

rahaeli:
[...]

PR #1516, opened 2015-08-07
Add FAQ revision control

me-and:
Create an `faqhist` table which will record changes made to FAQ entries for posterity.

This PR covers starting to store the information; it's not providing any sort of interface for recovering it -- for now that'll require direct database access -- and it isn't going to do anything like spin through the existing FAQ entries and store them all.

Reverting an FAQ to a previous revision is a feature that could be added, but I don't know if it'd see enough use to be worth the effort; someone with database access can always pull of the old versions manually if necessary.

Coping all the current FAQ entries into the database is probably easiest done with a one-off command on the database itself; my SQL-fu isn't strong enough to work out what that command _is_ (although I probably could find out if necessary), but I'm pretty sure it's possible.


me-and:
This is #1455. Which I'm primarily mentioning because GHI doesn't seem to have made the connection itself and I'd like a handy link between the two.

kareila:
I like this approach.  Two comments:

1. I don't think it is necessary or helpful to duplicate current faqs into the faqhist table, only to store the previous revisions that are no longer displayed on the site.  To that end, I would remove the faqhist update statement from the INSERT section and only keep it in the UPDATE and DELETE sections.

2. If you follow my advice above, this no longer breaks my rule of repeated threes, but in general: long complicated identical statements should be abstracted out to avoid having them at a later date changed in one place and not another.  SQL statements are absolutely the poster child for this.  For two not-quite-identical queries close together, though, it's probably OK - just something to bear in mind for the general case. :)

Thank you for doing this!  If you'll remove the statement that duplicates newly created FAQs into the faqhist table, I'll give this a quick spin on my dreamhack to make sure I don't see any unanticipated issues, then hopefully get this in before the next code push.

kaberett:
(Adam's away from sensible computers and internet connections for the weekend - I believe he's being rained on in a field somewhere - so he's unlikely to be able to get to this pre-push. On the other hand, he's also unlikely to mind someone else doing the tidy-up, so I guess see if he contradicts me before Friday afternoon...?)

kareila:
I'll stay posted for further updates then. I certainly don't mind making the alteration myself post-merge as long as he signs off on it.

me-and:
I am indeed in a field being rained on this weekend, so I'm not going to be
able to look at this myself until Monday. Entirely happy for someone else
to make the mark-ups, though.

kareila:
Okay, I just realized that this is a bigger issue than just removing the update-on-insert.  Because it's always adding the new data to the faqhist table, not the data being replaced.  Which, okay, that's one way to do it, but I think a better way would be to save the old data when entering the new data, instead of just always making two copies of the new data. So the steps would be to look up the existing row, then add THAT row to faqhist, then either delete the existing row or replace it with the new data depending on what was requested. And I think the "add the existing row to faqhist" procedure will always be the same no matter which action is being taken, although if you also want to continue to log a row full of nulls to indicate who and when a FAQ is deleted, that's cool too.

Also: when two or more db statements are happening in a row, we're only checking for an error on the last statement executed.  We should check for errors after every statement, and print the error and stop if we get one.  I think the typical procedure looks something like `$dbh->do(...); if ($dbh->err) { return $dbh->errstr; } else { $dbh->do([the next thing...]); }`

Anyway, let me know your thoughts, and sorry I wasn't able to get this in today!

me-and:
On rewriting the points at which we write to the database, I've not given your comments much thought yet; I'll do so in the next couple of days, after I've recovered from running around in a field all weekend.

On checking the return of the previous database action before returning, my code already does that. I've done it with what appeared to be the paradigm used in other bits of the codebase I've seen: adding `unless $dbh->errstr` at the end of database actions after the first.

kareila:
Yes, I saw that you added it where it wasn't happening before, and thank you!  What I'm suggesting is that it be checked before the new statement as well as afterward.

me-and:
Okay, I'm getting confused. In all three cases there are now three possible behaviours:

* Both database transactions succeed, in which case `$dbh->errstr` is never set, and the string returned is the "All good" or "FAQ item deleted" confirmation string.

* The first database transaction fails. In that case, `$dbh->errstr` will be set after that first transaction, the second transaction will not be attempted due to being gated by `unless $dbh->errstr`, and the string returned will be that error string.

* The first database transaction succeeds so the second one is attempted, but that fails. In that case, the string returned will be the error string from that second transaction.

What do you think should be handled differently?

kareila:
Oh, you're absolutely right! I wasn't seeing the other conditionals because of the postfix notation. My bad!

me-and:
(I dislike postfix notation for exactly that reason, but it seems to be the common style for such things in the DW code, so...)

me-and:
Taking another look at this, and in particular this comment from @kareila (which I believe is now the only thing stopping this being merged):

> I don't think it is necessary or helpful to duplicate current faqs into the faqhist table, only to store the previous revisions that are no longer displayed on the site. To that end, I would remove the faqhist update statement from the INSERT section and only keep it in the UPDATE and DELETE sections.

By my understanding, it's not as simple as just removing the code from the INSERT section: if we don't want the current faqs to be listed in the faqhist table, the UPDATE and DELETE sections also need to be updated so they're writing the old faq entry to the table, rather than the new entry / a dummy "deleted" entry.

As such, I think I have two objections to making that change, one pragmatic, one principled:

* At the moment, the writes to the database only need the new faq contents, which are written to both databases. Writing the incoming faq contents to the faq table and the outgoing contents to the faqhist table makes the code more complex: it would need to read the current data from the faq table, write that to the faqhist table, then write the new data to the faq table, rather than just writing the same data to both tables.

* I'm considering the faqhist table to be a simple version control system, and normally a version control system will have the tip pointing to the latest version of the content, rather than the version before. It seems unintuitive to me that the faqhist table would exclude the most recent history of each faq entry.

Is there anything I'm missing that makes writing the current faq entry to the faqhist table a problem?

kareila:
If we decide to stay with this approach, I think the only obstacle to this being merged is putting something in update-db.pl to initially populate the new table if there are existing entries in the current faq table.

I am going to ping @zorkian to take a look at this when he has a few minutes. Since it's a backend change, it's best for him to review it in any case.

bin/upgrading/update-db-general.pl
```diff
@@ -198,6 +198,23 @@
...
+register_tablecreate("faqhist", <<'EOC');
+CREATE TABLE faqhist (
+    rev mediumint(8) unsigned NOT NULL auto_increment,
+    faqid mediumint(8) unsigned NOT NULL,
+    question text,
+    summary text,
+    answer text,
+    sortorder int(11) default NULL,
+    faqcat varchar(20) default NULL,
+    lastmodtime datetime default NULL,
```
zorkian: For all times we store them as `INT UNSIGNED` and use epoch time.

zorkian:
So this adds new rows to some `faqhist` table, but what do we do with them? It seems like we're just storing them forever in a table but there's no way people can see them?

alierak:
Correct - see the opening comment of the PR.

zorkian:
Saving data "for posterity" seems kind of bogus. Why are we saving these things? What is the purpose, what is the intent? I read the original issue filed but it seems it was just because it caused sadness that they aren't stored?

Every byte stored costs us money. It'd be good to make sure we do it for a reason and we actually have a plan to do something with the data. Is there an issue filed with the plan?

azurelunatic:
The problem is that it's way too easy to blow away a FAQ by accident. When I was doing docs on LJ I did not know a single docadmin who had not done it at least once. 

Decreasing the amount of damage it's possible to do by accident may also open up granting edit privs to more people; one of the current documentation bottlenecks is that there are relatively few people who can edit.

zorkian:
"Store every revision forever" seems much different from "provide the last one so you can rollback from accidental oopses".

Also, in that case, I don't want to take the patch until there's a UI to pull the data out. Otherwise someone is going to have to bug the sysadmins to go select out FAQs and that's not going to fly.

bin/upgrading/update-db-general.pl
me-and: The code to use `datetime` and `NOW()` rather than `INT UNSIGNED` and `UNIX_TIMESTAMP()` was taken from the existing code for the faq table. I'd prefer to keep the timestamps used by the new faqhist table in the same format as those used for the existing faq table.

me-and:
@zorkian If we take this patch now, we can start populating the database now and write a UI later. I agree a UI will make this feature vastly more useful, but I don't believe waiting until this feature is perfect is a good reason to refuse having the improvements in the first place.

I'd very much support opening a new issue to cover adding a front end once this back-end patch is merged; indeed I can volunteer to write up a front-end once this is merged.

Storing every revision forever is the paradigm of every version control system I've worked with, and this is just the back end for another simple version control system. The immediate purpose is to catch accidental blatting of an FAQ, but it may also be useful to see how FAQs have changed over time, or to see how DW explained users should do some task five years ago to work out why their database record is in a bad state now, etc.

Yes, every byte costs money, but this is not a table that is going to grow rapidly or store large amounts of data, particularly compared to, say, providing image storage for all users. And if it _does_ turn out that you want to delete old entries because they're taking up too much disk space, that's very easy to do if and when it becomes a problem, rather than trying to anticipate all possible problems now.

zarhooie:
To touch on the version storing... It would make sense to me to store the five most recent versions, and dump anything older than that. Storing only the immediately previous change won't do us a whole lot of good, as any small changes (such as typo fixes, bad links, etc) will make that useless. The most times I can remember needing to jump back into a FAQ to edit it is maybe 2-3 times, so five should do it.

This seems to strike a nice, tidy, happy medium between "store everything ever" and our current status of "store nothing," and is what I'd prefer to see done as one of the people who has, in the past, wiped out an entire FAQ or three.

Mark, the spec for this was always just the back end. If storing the five most recent versions is acceptable, can we please push this? It's been one of Support's Most Wanted since before I started, and it's important to me to see it live. Kab is working up the spec for the front end, and will file the issue for it in a bit here.

(And, for what it's worth, if I needed to ask a sysadmin to pull a FAQ out of the database because it got oops'd, it would be once or twice a year, tops. The FAQs get touched just often enough that having a backup system is important, but not often enough that not having a UI for it should block this from being pushed.)

me-and:
Okay. I've discussed this offline with @zarhooie. It makes sense to us to take this patch now and open a new issue to cover limiting the history of each entry; she's asserted it'll be some time before the database grows to any particular size anyway.

zorkian:
I'm sorry if I wasn't clear, but I am not taking the patch until we get a way for support admins/someone to pull the data out. I don't care what that method is, but it cannot rely on sysadmins to pull from the database directly.

kaberett:
I really don't understand why that's the case. If this PR is merged now, the absolute worst-case scenario is that backups exist and a frontend can be written. I can't see any reason why backups existing in a way that's accessible only to sysadmins is incompatible with a policy that accidental deletions are handled via the existing methods, for all that they're stressful and inefficient, while a frontend is prepared. Additionally, I really can't see that data storage is a meaningful concern here, given that I'm currently the main person editing FAQs and I update my journal *way* more frequently than I ever edit FAQs.

I'd much rather have backups (and need to write a way to get them out) than not have backups: when this idea was discussed at OSB it was met with enthusiasm. I am absolutely certain a frontend can be written in fairly short order, but you didn't state that it was a requirement when planning for this feature began and from here it honestly looks as though you've made an arbitrary decision that something you'd never previously mentioned was important is a non-negotiable. I don't see that there's any incentive at all for people to work on a frontend prior to this PR being merged, since - again, from my perspective - there is the reasonable expectation that you'll again make an arbitrary decision that merging is blocked by something never previously mentioned as a requirement. At least if this *is* merged there's some confidence that the effort put into writing a frontend won't be wasted as new requirements arise as if from the aether.

zorkian:
The original ticket says "there is no version control". It does not specify a plan nor any sort of really critiqueable element, and I'm not going to arbitrarily write up a design doc for every random ticket someone files. Unfortunately, based on the text in the original ticket there was no feedback that could be provided. It was my belief that any reasonable implementation would be a "version control system" for FAQs.

Simply recording every version forever as this patch does is not a version control system. A version control system should contain (at a brief glance): storage of prior revisions, the ability to view the prior revisions, and the ability to roll back/forward to prior versions. These are all features of any version control system. If this patch had actually implemented what is the minimum reasonable set of features provided by a version control system then it would have been accepted. Instead, this patch does not, so it is not yet accepted.

Yet being the key word -- I totally agree the rest will come, but I want to see it before landing this. I can tell you now that the DB schema in this patch is going to need some work, but we won't know what kind of work until there's a frontend using it so we know what kind of queries are being written so we can make sure the indexes are created correctly to minimize database work. Landing the DB schema first is almost never the right decision in my experience; it is the wrong end to start the problem on if you are going to do work piecemeal.

This patch aside: if you would prefer to have less arbitrary feeling "needs more work" feedback rounds, then you should consider following a standard development methodology. As a suggestion, you could write up a design and post it on the ticket and then get sign-off from one of the committers (@kareila, @afuna, @zorkian) before implementing. If neither happens before you lay down code then you will always run the risk that the code will not end up being acceptable to the project for some reason or another.

I'm not trying to be capricious or arbitrary here and I'm sorry it sounds that way. I really do encourage people to ask for feedback, write up designs, and make sure the plan is acceptable before laying down code. You will ultimately save a lot of time, effort, and aggravation. (It will also let me feel less like the bad cop when I have to say no for some reason... so selfishly, I really do prefer when people follow good practice.)

me-and:
Right. Thank you for explaining why you think having a front-end is necessary; up until this point it had read very much as if you were looking for excuses to reject the PR, and I was on the verge of announcing I was entirely done with attempting to get this feature in.

I had understood @kareila's [comment above](https://github.com/dreamwidth/dw-free/pull/1516#issuecomment-153828962) to constitute sign-off, even excluding the conversations I'd had at OSBridge discussing and agreeing this design, so it's not clear to me how getting sign-off from one of the committers would have avoided this situation.

In the name of avoiding further rounds of my work being rejected for no obvious reason, can you please confirm that you're happy that my code does actually conform to the conventions regarding using `datetime` and `NOW()`? 

Likewise, can you please explain why you think the additional (I'd guess maximum few hundred kB per year) data storage the new table would require is more expensive than the additional server load to find and remove old records, the additional maintenance costs of the more complex code required for that, the additional cost every time someone does have an unanticipated reason for wanting to see an older entry, and the additional cost of everyone's time in arguing over it in the first place?

zarhooie:
@zorkian Thank you for spelling all that out, especially the stuff about why we need the front end in place first.

kareila:
I asked for @zorkian's opinion because I wasn't sure if this was the best approach for implementing the backend for this feature.  He makes a good point that it would have been better all around to start by hashing out a full design for how the feature would work from start to finish. I blame myself for not thinking to request that sooner, and hopefully will learn from this going forward.

As a baseline, I still believe this is a good start, but it's become clear that more work needs to be done before it is sufficiently featured enough to merge in.  Thank you for the work you've done so far, and let me know how best I can help you to proceed with this.

me-and:
Hi @kareila, thanks for the offer of help.

There's still two things I need before I can get back to/on with coding this, both clarifications on some of @zorkian's objections above.

Firstly, there was an issue to do with the disk space the new table would use. If that's really a problem, I'd like to understand a bit more about why so that I can address those concerns; it seems to me that the space this table would use would be completely trivial compared to DW's existing disk usage.

Secondly, there was a comment on some of the code about it not matching code conventions. As best I can tell, my code does match the conventions in use in this area of the codebase, so I'd appreciate guidance on whether all that code needs refactoring or my usage is fine, or something in between.

In both cases, all I need is either some explanation about why the problem is a problem, or confirmation that it isn't actually an issue. I just want to make sure I'm going in the right direction in further work.

kareila:
Personally I'm a big fan of consistency, so I agree with using the same data types for the faqhist table that the faq table uses.  While there's merit to the argument of following different conventions for new tables, in this case we're basically creating a duplicate of an existing table, so I do think they should exactly match what is already in use, which of course we inherited from LJ.  There's also the question though of table indices, the choice of which will depend on how the queries for fetching the data are structured.

As for disk space, I think the concern was that the table would grow without bound, which would eventually impact performance.  It would be best to just save the past 5 revisions or so of any given faq ID.

me-and:
Thank you for confirming you agree about my usage of date stamp formats.

I understand that concern about disk space, but I don't understand how it balances with other concerns:

* Adding code to check the number of historical entries and delete old ones will make the code more complicated, requiring more database queries and incurring higher code maintenance costs.

* The disk space required for this table, given what I understand of how frequently the FAQs are edited, will be at least an order of magnitude lower than the disk space used by a single reasonably active DW journal or community.

* If it turns out in future that disk space usage is a problem, it's one that's easily fixed by adding the code to delete old entries; doing it now seems like a premature optimization.

* If there's a limit on the entries stored, it means that if/when there's a need for an older entry, we won't be able to get it. If we don't delete old entries, it's easy to start deleting them if disk usage becomes a problem. If we do delete old entries, it's probably impossible to recover those entries if we decide we need them for some unanticipated reason.

kareila:
Hmm, this should probably be on the wiki somewhere if it isn't already:

DW uses two basic types of database tables - global and clustered.  The reason
that journals can grow without bound is that they are clustered, meaning that
they can be easily spread across multiple machines/database instances to improve
performance.  Meanwhile, global tables like the FAQs share one central database
which must be managed more carefully.

While obviously "disk space" per se is cheap, search and replication across the
site will be negatively impacted by large global tables.  So it is not premature
to manage the overall usage of resources here.

However, we can reduce the overhead required to expire older revisions by outsourcing
it to one of the daily worker scripts, instead of checking every time a query is run.

The number of revisions retained should be governed by the needs of the FAQ
maintainers.  The mentions I've seen gave me the impression that it would be
quite uncommon to need to go more than one or two revisions back, so keeping
five seemed like a good limit to set.

Let me know if anything still isn't clear!

me-and:
Okay, thank you. That's useful and gives me enough to get on with :)

kareila:
Excellent, glad I could help!

me-and:
Right, have a UI!

A few concerns/limitations on this, which I want to flag:

- If an FAQ is deleted, a subsequent new FAQ ID will reuse the deleted ID. The revision history of a FAQ follows the FAQ ID, so the new FAQ's history will include the history of the old FAQ, up to and including its deletion.

- The code for generating the version history is duplicated in `faqedit.bml`. Ideally I'd put the common code into a function and just call the function twice, but the duplicated code is very much generating the user-visible content, so it doesn't seem to be appropriate to put them in the Perl modules, and it doesn't seem to be possible to define functions and then call them in `.bml` files.

- If you go to a page for a FAQ that has never existed, as opposed to one that used to exist but has been deleted, the versions drop-down is empty, and the "Load" button effectively just refreshes the page.

me-and:
And now with a worker thread for deleting old FAQ entries (with thanks to @kaberett for reminding me about it!), although I'm far from convinced I've set it up to run on a regular basis in the correct fashion.

zorkian:
(Form letter incoming.)

me-and:
Finally resurrecting this (I *think* if the PR gets reopened, it will automatically update with the current contents of my branch).

A few caveats:

1.  I've moved the cleanup into `clean_caches.pl`, and switched it to using `$db->selectrow_array` rather than `$db->prepare`, but I can't immediately work out how to kick off a run of that script to check it does what it's supposed to. I've verified the SQL commands with `ljdb`, but that's obviously only a small part of the code here.

2.  I've not moved the other changes to use `selectrow_array` or similar, as (a) that wouldn't be consistent with the existing code in those files and (b) I'm getting myself very confused trying to understand how to actually make that style of code work in those cases.

3.  When I've been testing this code on kaberett's Dreamhack, I'm seeing what looks like some very odd caching behaviour that I'm going to need some help debugging: after creating a FAQ entry and making a bunch of edits to populate the history table, the edit page seems to load a random entry from the recent history of that FAQ. 

    Inserting a bunch of `die` commands implies it's not loading these from the faqhist table (which is correct: it should only load from there if there's a `rev=` argument in the URL), and using `ljdb` to edit the database tables *doesn't* affect what's loaded on the web front-end: the front-end will load things that have been deleted from both the faq and the faqhist tables.

    I see this behaviour across computers: if I edit a FAQ entry, then change the tables in the back-end with `ljdb`, kaberett's computer will load the edit page with values that no longer appear anywhere in the FAQ. Restarting Apache does get it to load the actual data from the database, however, so my supposition is this is to do with caching on the Dreamhack server, but I've *no idea* where to start with debugging that.

    The editing the backend test is the most obvious and probably the easiest to debug, but the more real-world-likely problem is that if there are multiple edits to a FAQ since Apache was restarted, the edit page seems to load any one of those versions without any pattern I can see.

So, could y'all can help me test `clean_caches.pl`, confirm you're happy with leaving the `$sth`-style code in faqedit.bml (which I assume you are given the comment about moving away from that style was optional), and help me work out what's causing the edit page to load incorrect versions of the FAQ?

alierak:
I think you're looking for "$LJHOME/bin/ljmaint.pl clean_caches".

alierak:
I also think your LJ::Faq->load ignores the rev option if loading a faq in the site's default language.

cgi-bin/LJ/Faq.pm
```diff
@@ -88,7 +91,7 @@ sub load {
         return undef unless $f;
         $faq = $class->new(%$f, lang => $lang);
 
-    } else { # Don't load fields that lang_update_in_place will overwrite.
+    } elsif ($rev == 0) { # Don't load fields that lang_update_in_place will overwrite.
```
alierak: Okay, so there are two cases in this if statement already: loading a faq in the site's default language, and loading a faq in some other language. Adding this elsif here gives you three cases: loading a faq in the site's default language, loading a faq in some other language at revision 0, and loading a faq in some other language at a nonzero revision. Shouldn't there be a fourth case for loading a faq in the site's default language at a nonzero revision?

cgi-bin/LJ/Faq.pm
```diff
@@ -98,6 +101,16 @@ sub load {
...
+    } else {
+        my $f = $dbr->selectrow_hashref
+                ("SELECT faqid, question, summary, answer, faqcat, lastmoduserid, ".
+                 "DATE_FORMAT(lastmodtime, '%M %D, %Y') AS lastmodtime, ".
+                 "UNIX_TIMESTAMP(lastmodtime) AS unixmodtime, sortorder ".
+                 "FROM faqhist WHERE faqid=? AND rev=?",
+                 undef, $faqid, $rev);
+        die $dbr->errstr if $dbr->err;
+        return undef unless $f;
+        $faq = $class->new(%$f, lang => $lang);
```
alierak: I'm assuming this part has to be revised to deal with calling lang_update_in_place since it's supposed to be loading a faq in something other than the site's default language, given its position at the end of the if statement.

etc/workers.conf
```diff
@@ -121,3 +121,4 @@ reference-list:
     talklib-gm: 1           # loads comments using a worker (performance thing)
     t-memlimit: 1           # for running certain tests
     userpic-resize-gm: 1    # resizes userpics
+    delete-old-faqs: 1      # deletes old FAQ entries from the faqhist table
```
alierak: (this list preferably alphabetized)

rahaeli:
I am 100% fine with removing all alternate-language support for the FAQ system to make it easier here -- we are never going to use it -- or with only making it possible to load previous revisions of the default language.

me-and:
Huh, fair enough! I expect it's objecting to the rebase I did to keep the history simple. I'll open a new PR once I've addressed @rahaeli and @alierak's comments.

PR #1521, opened 2015-08-07
Add poster name to Edit Tags page

### LINKED ISSUE #1509: Show name of poster on "Edit tags" page
https://dw-suggestions.dreamwidth.org/1375578.html

Was bug 4962 in the days of 'zilla.

Text of suggestion:

Description:
The "Edit Tags on an Entry" page currently has the fields:

* subject
* current tags
* [journal]'s tags
* entry text

This is fine if you're tagging within your own journal. However, when I'm tagging in a community, and especially tagging dw_suggestion's posts, I sometimes get curious about Who Submitted This, & it would be lovely to be able to see that information at the same time as tagging, rather than having to click about (which disrupts my concentration).

I can also see this being useful for batch-tagging in communities where posts are routinely tagged with the name of the poster.

Possible solutions include:

(1) always showing the username of the poster on the Edit Tags on an Entry page
(2) show the username of the poster only when editing tags on an entry posted to a community

Downsides: you end up with one extra uneditable line on the Edit Tags on an Entry page. Personally I can't get too worked up about that. ;)

kaberett:
Fixes #1509.

Provides name of poster on dedicated Edit Tags page - useful/
of interest when managing community tags.

This is still BML and I should learn to fix it.

### DIFF
--- a/htdocs/edittags.bml
+++ b/htdocs/edittags.bml
@@ -103,6 +103,9 @@ body<=
     $ret .= "<script type='text/javascript'> var cur_taglist = '$logtagstr'; </script>";
 
     $ret .= '<table summary="" class="edittbl" cellpadding="0" cellspacing="0" width="50%">';
+    if ( $u->is_community ) {
+        $ret .= "<tr><td class='l'>$ML{'.poster'}</td><td>" . $ent->poster->ljuser_display . "</td></tr>";
+    }
     $ret .= "<tr><td class='l'>$ML{'.subject'}</td><td>$subj</td></tr>" if $subj;
 
     $ret .= '<form method="POST" action="/edittags" id="edit_tagform">';
--- a/htdocs/edittags.bml.text
+++ b/htdocs/edittags.bml.text
@@ -25,6 +25,8 @@
 
 .permissions.none=(You don't have permission to edit tags on this entry.)
 
+.poster=Poster:
+
 .readonly.journal=This journal is read-only.  Its entry tags can't be edited.
 
 .readonly.poster=This poster is read-only; the tags for this entry can't be edited.


kaberett:
... there is presumably a tidier way to get the user name to display? I was just utterly failing to find it.

rahaeli:
ljuser_display is the usual way of doing it in a BML context!

kareila:
Yes, that's the right method.  Unfortunately, it's the wrong user object. What this displays is the name of the community, not the name of the poster of the entry. You want to use `$ent->poster->ljuser_display` instead.  For bonus points, only have the line print if `$u->is_community` is true.

And please don't feel guilty about adding small features like this to BML pages without converting the entire page first. It'll get done eventually. :heart:

rahaeli:
oh man, that's bitten me more times than I can count, I probably should have noticed it. :)

kaberett:
... AND THIS IS WHY I SHOULD GET MORE SLEEP BEFORE MAKING PULL REQUESTS. Whoops. Fix'd, thank you!

htdocs/edittags.bml
```diff
@@ -103,8 +103,11 @@ body<=
     $ret .= "<script type='text/javascript'> var cur_taglist = '$logtagstr'; </script>";
 
     $ret .= '<table summary="" class="edittbl" cellpadding="0" cellspacing="0" width="50%">';
+    if ( $u->is_community ) {
+        $ret .= "<tr><td class='l'>$ML{'.poster'}</td><td>" . $ent->poster->ljuser_display . "</td></tr>"
+        };
```
kareila: This works fine now, thanks! I need to do a little style nit-pick though: the semicolon should be immediately after the `</td></tr>`, and the closing `}` should be on a line by itself, indented at the same level as the if statement.

htdocs/edittags.bml
```diff
@@ -103,8 +103,11 @@ body<=
...
     $ret .= "<tr><td class='l'>$ML{'.subject'}</td><td>$subj</td></tr>" if $subj;
-
+
```
kareila: Since I'm already requesting a change to adhere to our coding style guidelines, can also I ask you to take out the extra whitespace that was added here?

kareila:
This is really close! Next round should get it in :+1:

htdocs/edittags.bml
kaberett: Pretty sure that's actually git getting confused about whitespace moving? i.e. there was already a line of whitespace at this location, it's just that git thinks I've removed it and replaced it with a new one. If you want the line of whitespace gone altogether that's cool and I'll do it, but can I have clarification please? :-)

kareila: I think what happened was a blank line (e.g. `\n`) was replaced with a line that contains whitespace indenting the line to the level of the line above (e.g. `\s\s\s\s\n`) - it should ideally be completely unchanged.  These creep in from time to time - if it's too troublesome don't worry about it!

kaberett:
facepalm. actually fixed. I think. APPENDAGES CROSSED.

kareila:
yesssss perfect :100:

PR #1525, opened 2015-08-08
Hide subscription options for non-visible accounts

### LINKED ISSUE #1523: Possible to see tags used by deleted account
If you view the profile page of a deleted user account, there's a link to be notified when the account is purged, which points to `http://www.dreamwidth.org/manage/tracking/user?journal=<username>`. If you follow that link, you get the list of things you can subscribe to, which includes "<username> posts a new entry tagged \<dropdown\>".

Where that dropdown includes all the tags that user ever used.

I'm pretty sure that, when a user deletes their account, they don't want other people to be able to see all the tags they had used. There's potentially a hell of a lot of information there. Certainly there is in the account where I just spotted this behaviour.

me-and:
When offering tracking subscription options for an account with a status
other than "visible", only offer notification of when the account is
purged.  In particular, don't offer notification of when a new post with
a specific tag is made, since that includes the list of tags from a
deleted account.

Fixes #1523.

### DIFF
--- a/htdocs/manage/tracking/user.bml
+++ b/htdocs/manage/tracking/user.bml
@@ -69,7 +69,7 @@ body<=
             journal => $journal,
             flags   => LJ::Subscription::TRACKING,
         ) if $remote->can_track_all_community_comments( $journal )
-            and $journal->statusvis eq "V";
+            && $journal->statusvis eq "V";
 
     push @{ @$categories[0]->{$cat_title} },
         LJ::Subscription::Pending->new( $remote,
--- a/htdocs/manage/tracking/user.bml
+++ b/htdocs/manage/tracking/user.bml
@@ -70,7 +70,7 @@ body<=
             flags   => LJ::Subscription::TRACKING,
         ) if $remote->can_track_all_community_comments( $journal )
             and $journal->statusvis eq "V";
-    
+
     push @{ @$categories[0]->{$cat_title} },
         LJ::Subscription::Pending->new( $remote,
             event    => "NewUserpic",
--- a/htdocs/manage/tracking/user.bml
+++ b/htdocs/manage/tracking/user.bml
@@ -41,24 +41,20 @@ body<=
 
     # what classes to display on this page
     my $cat_title = 'Track User';
-    my $categories =
-        [
-            { $cat_title =>
-                [
-                    LJ::Subscription::Pending->new( $remote,
-                        event   => "JournalNewEntry",
-                        arg1    => '?',
-                        journal => $journal,
-                        flags   => LJ::Subscription::TRACKING,
-                    ),
-                    LJ::Subscription::Pending->new( $remote,
-                        event   => "JournalNewEntry",
-                        journal => $journal,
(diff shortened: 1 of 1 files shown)


htdocs/manage/tracking/user.bml
```diff
@@ -72,8 +68,9 @@ body<=
             event   => "JournalNewComment",
             journal => $journal,
             flags   => LJ::Subscription::TRACKING,
-        ) if $remote->can_track_all_community_comments( $journal );
-    
+        ) if $remote->can_track_all_community_comments( $journal )
+            and $journal->statusvis eq "V";
```
zorkian: Prefer to use `&&` unless you need the lower precedence operator for some reason.

me-and: Nope; that's fine -- I hadn't realised there was a difference in Perl. New version coming up imminently.

zorkian:
Although now that I've merged it, wasn't there a utility function that did the statusvis check? Instead of checking it yourself? `is_visible` or something...

zorkian:
Anyway, no biggie, I just can't remember. Thanks for doing this!

rahaeli:
Yes! We have two: $u->is_deleted checks for statusviz == D, and $u->is_visible checks for statusviz == V. (Functionally they are not exactly the same thing -- is_deleted is "is this in the deleted state, but not the visible, suspended, or expunged state", and is_visible checks "is this in the visible state, but not the suspended, expunged, or deleted state" -- but are usually close enough for government work.) They live in LJ::User, and currently can be found in cgi-bin/LJ/User/Account.pm, along with many other helpful shortcuts.

(I don't think the difference is a huge deal -- IIRC, we just use the utility functions so the code reads more Englishy -- since it's not like we'd fuck around with changing various statusvizes lest we introduce subtle and exciting bugs that might make the whole site turn into a turnip.)

me-and:
Okay, evidently I missed those. For the sake of cleanliness, I'll create
another pull request using those functions soon.

rahaeli:
Don't worry too much about it! It's mostly for code readability and what you have is pretty readable :)

me-and:
Eh, it's not a difficult fix, and I reckon `is_visible` is more usefully self documenting than `statusvis eq "V"`. Plus doing the same thing in different bits of code is a pet irritation of mine :)

me-and:
Also, as a point of interest, `is_deleted` and `is_visible` do both just check for `"D"` and `"V"` respectively, so using the functions is absolutely identical to what my code was already doing.

Of that group of functions, the only ones that don't just check `statusvis` against a single letter status are:

*   `is_expunged`: true iff `statusvis` is `"X"` or `clusterid` is `0`.
*   `is_inactive`: true iff`statusvis` is any of `"D"`, `"X"` or `"S"`
*   `is_readonly`: true iff `statusvis` is `"R"` or `get_cap('readonly')` is true

rahaeli:
Ah, the number of ways in which our code occasionally does something completely unexpected. 

(IIRC, the clusterid/get_cap checks are there because of old bugs on LJ where somebody's account got into a weird state. :) )

PR #1526, opened 2015-08-08
[#1374] allow rename swaps to use tokens from either account

### LINKED ISSUE #1374: make two-token rename swap possible independent of which account the tokens are on
In order for the two-token username swap to work, right now you need to have two rename tokens available on the account you're logged into when you initiate the swap, not one token on each account. This is confusing to people. Instead, the system should check both accounts for available tokens, and if it finds one token on each account (or even 0 tokens on the logged-in account and 2 tokens on the other account) it should use those tokens and succeed.

kareila:
Updates DW::User::Rename::swap_usernames to validate tokens from any
of the (up to three) accounts involved in a username swap, instead
of only using the available tokens for the logged in account.

Also redesigns the /rename/swap page to present the form immediately,
and treat the "too few tokens" error like any other form processing
error.  This was necessary in order to check the requested target
account(s) for available tokens.

Fixes #1374.

### DIFF
--- a/cgi-bin/DW/User/Rename.pm
+++ b/cgi-bin/DW/User/Rename.pm
@@ -229,10 +229,14 @@ sub swap_usernames {
     my $admin = LJ::isu( $opts{user} ) ? $opts{user} : $u1;
     my @tokens = @{ $opts{tokens} || [] };
 
+    # tokens can be owned by $admin (remote user),
+    # $u1 (authas user), or $u2 (intended target)
+    my $check_uids = { $admin->id => 1, $u1->id => 1, $u2->id => 1 };
+
     foreach my $token ( @tokens ) {
         $errors->add( 'token', 'rename.error.tokeninvalid' )
             unless $token && $token->isa( "DW::RenameToken" )
-                && $token->ownerid == $admin->userid;
+                && $check_uids->{$token->ownerid};
 
         $errors->add( 'token', 'rename.error.tokenapplied' )
             if $token->applied || $token->revoked;
--- a/t/rename.t
+++ b/t/rename.t
@@ -15,7 +15,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 153;
+use Test::More tests => 154;
 
 BEGIN { $LJ::_T_CONFIG = 1; require "$ENV{LJHOME}/cgi-bin/ljlib.pl"; }
 use LJ::Test qw( temp_user temp_comm );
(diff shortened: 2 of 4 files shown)


cgi-bin/DW/User/Rename.pm
```diff
@@ -229,10 +229,14 @@ sub swap_usernames {
     my $admin = LJ::isu( $opts{user} ) ? $opts{user} : $u1;
     my @tokens = @{ $opts{tokens} || [] };
 
+    # tokens can be owned by $admin (remote user),
+    # $u1 (authas user), or $u2 (intended target)
+    my $check_uids = { $admin->id => 1, $u1->id => 1, $u2->id => 1 };
```
zorkian: Style-wise, I'd prefer hashes over hashrefs. Less syntactic sugar involved in the usage, and morally, not dereferencing a pointer should be faster. (Although quite irrelevant in this usage.)

cgi-bin/DW/User/Rename.pm
```diff
@@ -229,10 +229,14 @@ sub swap_usernames {
...
+
     foreach my $token ( @tokens ) {
         $errors->add( 'token', 'rename.error.tokeninvalid' )
             unless $token && $token->isa( "DW::RenameToken" )
-                && $token->ownerid == $admin->userid;
+                && $check_uids->{$token->ownerid};
```
zorkian: Since you explicitly set the values to be `1` above, maybe check `== 1`? Else, consider `exists ...`.

t/rename.t
```diff
@@ -532,6 +532,16 @@ note( "-- two username swap (personal to personal)" );
     is( $u2->userid, $u2id, "Id of u2 remains the same after rename." );
 }
 
+note( "-- two username swap (personal to personal), one token owned by each user" );
```
zorkian: A test! YAY! :yum:

zorkian:
This is LGTM and can be merged, but I made some comments if you want style feedback.

cgi-bin/DW/User/Rename.pm
kareila: The problem with checking `== 1` is that it will log a warning if the requested hash value is undefined.  I'm just using 1 as a convenient truthy value here; I think it's fine as is. :wink: Thanks for the feedback though!

zorkian: In that case, since the thing you care about is whether or not it's in the hash, you should definitely use `exists` so that's clear.

PR #1530, opened 2015-08-10
Improve code clarity

me-and:
Use `journal->is_visible` in preference to `journal->statusvis eq "V"` -- the former is considerably more readable.

Per @zorkian and @rahaeli in #1525 pointing out to me that `is_visible` is a thing that exists.


kareila:
Yay readability!

PR #1531, opened 2015-08-10
Rewrite README into Markdown

me-and:
The repository README was already almost in Markdown format.  Polish it
up a bit further into that format, and rename it to use an extension
that will cause GitHub to parse and display the file as Markdown.


kareila:
Sweet. Thanks!

PR #1533, opened 2015-08-11
[#1356] remove "identity_posting" userprop and related checks

### LINKED ISSUE #1356: OpenID community maintainers are unable to edit tags
http://www.dreamwidth.org/support/see_request?id=30652

I dug into the code and found this in LJ::Tags::can_add_tags:
```
# we don't allow identity users to add tags, even when tag permissions would otherwise allow any user on the site                                                                                                          
# exception are communities that explicitly allow identity users to post in them                            
# FIXME: perhaps we should restrict on all users, but allow for more restrictive settings such as members?  
return undef unless $remote->is_personal || $remote->is_identity && $u->prop( 'identity_posting' );         
```

So I looked to see where in the settings we can turn on `identity_posting` for a community, and... it doesn't seem to be available.  Maybe our treatment of OpenID users changed somewhere along the line, and this got missed?

Either we should update cgi-bin/DW/Setting/CommunityPostLevel.pm to allow this setting to be turned on, or we should remove the check for it from `can_add_tags`, `can_add_entry_tags`, and `can_control_tags` in LJ/Tags.pm.  There is also a check for it in LJ/Protocol.pm.

kareila:
The legacy userprop "identity_posting" was intended to allow community admins to opt into allowing OpenID users to post entries and manage tags.  Since we decided to allow this for all OpenID users across the site, this set of changes removes the userprop and all the permission checks that were failing as a result of it not being set.

The relevant tests in protocol.t have been updated, but that test suite is still set to be skipped because of inconsistent results in other parts of the test suite.

Fixes #1356.

### DIFF
--- a/bin/upgrading/proplists.dat
+++ b/bin/upgrading/proplists.dat
@@ -221,7 +221,7 @@ userproplist.disable_comm_promo:
   indexed: 0
   multihomed: 0
   prettyname: Administratively disable community promo features for a user
-  
+
 userproplist.displaydate_check:
   cldversion: 4
   datatype: bool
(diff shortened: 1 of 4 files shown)


zorkian:
Oh yes yes yes, deleting things make me so happy :heart:

kareila:
KARZILLA SMASH

rahaeli:
But does it involve an earth-shattering kaboom? I was promised an earth-shattering kaboom.

PR #1536, opened 2015-08-11
Fix uninitialized value warnings

kareila:
Latest round of fixes for various uninitialized value warnings from the Apache log on my devhack.  See commit message for the full list.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -740,12 +740,13 @@ sub _form_to_backend {
     $props->{taglist} = "" unless $props->{taglist} && $props->{taglist} =~ /\S/;
 
     if ( LJ::is_enabled( 'adult_content' ) ) {
+        my $restriction_key = $post->{age_restriction} || '';
         $props->{adult_content} = {
             ''              => '',
             'none'          => 'none',
             'discretion'    => 'concepts',
             'restricted'    => 'explicit',
-        }->{$post->{age_restriction}} || "";
+        }->{$restriction_key} || "";
 
         $props->{adult_content_reason} = $post->{age_restriction_reason} || "";
     }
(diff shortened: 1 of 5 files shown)


cgi-bin/DW/Controller/Entry.pm
```diff
@@ -740,12 +740,13 @@ sub _form_to_backend {
     $props->{taglist} = "" unless $props->{taglist} && $props->{taglist} =~ /\S/;
 
     if ( LJ::is_enabled( 'adult_content' ) ) {
+        my $restriction_key = $post->{age_restriction} || '';
```
zorkian: For this kind of usage, where you've got `undef` on the left side, you need to use `//` instead of `||`. The former operator checks for defined-ness, which means you don't accidentally squash false values.

cgi-bin/DW/Controller/Support/History.pm
```diff
@@ -107,7 +107,7 @@ sub history_handler {
                 $reqs{$row->[0]} = {
                     spid => $row->[0],
                     winner => $points{$row->[0]}->[0],
-                    points => $points{$row->[0]}->[1]+0,
+                    points => $points{$row->[0]}->[1] || 0,
```
zorkian: The point of `+0` is to coerce to an integer, so your change here makes it so that a non-integer field doesn't get coerced like it was. So, really, this should be,

```
                    points => ($points{$row->[0]}->[1] // 0)+0,
```

cgi-bin/DW/Controller/Entry.pm
kareila: Yeah the reason I didn't do that here is all the valid values are enumerated below, they're all either text strings or the empty string.

cgi-bin/DW/Controller/Support/History.pm
kareila: Yeah the reason I didn't do that here was because the value from the database is typed to be an integer, and I believe the person who wrote the original statement used `+0` when they meant `|| 0`, just to guard against undefs.

So yeah, I had my reasons in the wider context of things! Thanks for giving me the benefit of the doubt :wink:  But if you think there's a stronger argument to be made in the other direction, I'm convincible.

zorkian:
Fair enough! Thanks for elaborating. This seems fine. :)

PR #1537, opened 2015-08-12
Closes #1517, adds TOC to faqlist when browsing by cat

### LINKED ISSUE #1517: include table of contents on faqcat view
Users can see all FAQs in a single category all at once by using the faqcat view, eg:

  https://www.dreamwidth.org/support/faqbrowse?faqcat=paidaccounts

It would be a nice touch if we included a table-of-contents at the top of that view, so each FAQ title is displayed up top, with the FAQ title in the ToC serving as an inside-the-page link to the FAQ in the list. (Because FAQ titles can change and we don't want to have to define a shortcode for each FAQ, permalink by number is probably best, so -- for instance -- ?faqcat=paidaccounts would have a ToC whose first element was "What is a paid/premium paid/seed account?" and the link goes to a #faq233 permalink, linking to the display of faq233 further down the page.)

sgsabbage:
No styling on the ol to keep the text size the same. Am happy to make any styling changes required, though!

### DIFF
--- a/htdocs/support/faqbrowse.bml
+++ b/htdocs/support/faqbrowse.bml
@@ -112,7 +112,7 @@ body<=
         LJ::Faq->render_in_place( { lang => $lang, user => $user,
                                     url => $user_url }, @faqs );
 
-        $$body .= "<ol name='toc'>";
+        $$body .= "<ol>";
         foreach my $f (@faqs) {
             my $faqid = $f->faqid;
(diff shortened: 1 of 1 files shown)


sgsabbage:
I may or may not have messed up the pull request message again because of multiple commits. Just in case, fixes #1517.

kareila:
This looks good to me! Additional styling can be requested separately if desired.

PR #1540, opened 2015-08-12
[#1539] Displays 'Default' filter as selected correctly

### LINKED ISSUE #1539: Display "Default" instead of "All" on navstrip when viewing default reading page
From https://dw-suggestions.dreamwidth.org/578825.html

sgsabbage:
Fixes #1539.

Have tried to make as few changes as possible to Web.pm because it's a big file, but may have missed something.


kareila:
Brilliant! Way smaller change than I expected.  If you missed anything, I can't spot it.

PR #1549, opened 2015-08-21
antispam sorting

### LINKED ISSUE #1522: antispam: make spam report sortable
Currently, spam reports sort by date reported. 
I would like spam reports to also be sortable by date left, for troubleshooting purposes. 

Specific use case: say someone's email is compromised and a spammer gets into it. A site admin resets the secret address and notifies the user. Meanwhile, only one of the communities that the user accidentally spammed has reported in. Two months later, other community admins notice and spamdelete. I want to be able to sort by time of date left to make sure that all of it was left before the most recent secret address reset.

kareila:
Converted spamreports.bml to TT + Foundation for great justice, then added requested sorting feature.

Fixes #1522 and then some.


zorkian:
It's a lot of code -- I read through and things seem reasonable, but I didn't go over it with a fine-toothed comb. We should definitely be watchful when we push.

kareila:
I tested it as comprehensively as I could! But I'm sure Azz will let us know if I missed anything.

PR #1553, opened 2015-08-24
fix failing test in wtf.t

kareila:
My bad - a previous undefined variable warning fix led to the key
not always being deleted from the hash, which caused an error.
This should fix it.


zorkian:
Using `defined` instead of `exists` strikes again! Darn Perl...

PR #1554, opened 2015-08-25
[#1527] fix issues with access-only images

### LINKED ISSUE #1527: DW-hosted image permissions interact poorly with custom access filters
Images set as visible to "Access List", when included in a post to a custom access filter, are visible only to the post owner and not to people in the access filter. Images set viewable to Everyone (Public) <i>are</i> visible when included in a post to a custom access filter. ("Access List" images included in posts visible to the entire access list are visible; I haven't tested whether the same applies given a default access filter, because I haven't got one.)

... as far as I can tell, having poked at it a little. I think it is fairly obviously the case that an image not being viewable in a custom-access-filtered post if it's set visible to Access List, but that it *is* visible if it's set to Everyone, is not ideal? In that custom access filters are a subset of the entire access list (in much the same way that the access list is a subset of "everyone"), so...

kareila:
Does the right thing with 'friends' setting, which is still used as an option for minimum entry security.  The /file/new path uses that setting when uploading images.

Makes sure if we request 'usemask' security, that the 'allowmask' defaults to 1 (all trusted).

New test file t/media-security.t to make sure things are working as we expect.

Fixes #1527.

### DIFF
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -110,7 +110,10 @@ sub upload_media {
 
     # set the security
     my $sec = $opts{security} || 'public';
-    $sec = 'usemask' if $sec =~ /^(?:friends|access)$/;
+    if ( $sec =~ /^(?:friends|access)$/ ) {
+        $sec = 'usemask' ;
+        $opts{allowmask} = 1;
+    }
     croak 'Invalid security for uploaded file.'
         unless $sec =~ /^(?:public|private|usemask)$/;
     if ( $sec eq 'usemask' ) {
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -117,8 +117,8 @@ sub upload_media {
     croak 'Invalid security for uploaded file.'
         unless $sec =~ /^(?:public|private|usemask)$/;
     if ( $sec eq 'usemask' ) {
-        # default allowmask of 1 unless defined otherwise
-        $opts{allowmask} //= 1;
+        # default allowmask of 0 unless defined otherwise
+        $opts{allowmask} //= 0;
     } else {
         $opts{allowmask} = 0;
     }
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -110,9 +110,15 @@ sub upload_media {
 
     # set the security
     my $sec = $opts{security} || 'public';
+    $sec = 'usemask' if $sec =~ /^(?:friends|access)$/;
     croak 'Invalid security for uploaded file.'
         unless $sec =~ /^(?:public|private|usemask)$/;
-    $opts{allowmask} = 0 unless defined $opts{allowmask} && $sec eq 'usemask';
+    if ( $sec eq 'usemask' ) {
+        # default allowmask of 1 unless defined otherwise
+        $opts{allowmask} //= 1;
+    } else {
+        $opts{allowmask} = 0;
+    }
 
     # now we can cook -- allocate an id and upload
     my $id = LJ::alloc_user_counter( $opts{user}, 'A' )
--- a/cgi-bin/DW/Media/Base.pm
+++ b/cgi-bin/DW/Media/Base.pm
@@ -173,16 +173,19 @@ sub set_security {
     confess 'Invalid argument hash passed to set_security.'
         unless defined $security;
 
-    $security = 'usemask' if $security =~ /^(?:friends|access)$/;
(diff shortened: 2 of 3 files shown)


cgi-bin/DW/Media.pm
```diff
@@ -110,9 +110,15 @@ sub upload_media {
 
     # set the security
     my $sec = $opts{security} || 'public';
+    $sec = 'usemask' if $sec =~ /^(?:friends|access)$/;
```
zorkian: The standard way we handle this access is,

```perl
if ( $sec =~ /^(?:friends|access)$/ ) {
    $sec = 'usemask';
    $allowmask = 1;
}
```

I know that your code _should_ be safe because it uses defined-or, but I'd much rather have the code be extremely explicit about the access being set and not touch it if it's already `usemask`.

cgi-bin/DW/Media/Base.pm
```diff
@@ -170,11 +170,17 @@ sub set_security {
     return 0 if $self->is_deleted;
 
     my $security = $opts{security};
+    confess 'Invalid argument hash passed to set_security.'
+        unless defined $security;
+
+    $security = 'usemask' if $security =~ /^(?:friends|access)$/;
```
zorkian: Same comments as above.

cgi-bin/DW/Media.pm
kareila: I'm pretty sure I saw at least one place where "usemask" was being used in lieu of "access" and no allowmask was passed in. Do I need to find it again and change it, or can I do this to protect against it on general principle?

zorkian: Yeah, we're not good at following standards or best practices, especially as code has come from many sources over 15 years and many different reviewers. :)

You definitely don't need to change every other occurrence (unless you want to go on a code dive, which is always welcome!).

kareila: In that case I strongly feel it best to allow the defined or, since usemask security + no allowmask is easy to do by accident and will cause problems.  That can be in addition to the explicit setting paradigm you noted above.

kareila: (If I really wanted to go on a rampage I'd figure out a way to abstract this out so we don't keep writing the same code wrongly everywhere, but I think there are other things that would be a better use of my time.)

zorkian: Okay. I was trying to make it easier.

I won't accept this patch as-is. It needs to be explicit. If you feel that means you have to go change every other occurrence, that's your choice. It doesn't seem like a good idea to block this improvement on those fixes, but that's your clal.

kareila: I did just change it to be explicit - all I was saying was that I thought it better to keep the other check in as well.

kareila: Okay, I changed it to a development warning instead of a default value.  Can we agree on that?

zorkian:
I'm in agreement about defined-or setting of `$allowmask`, but the value should be `0`. Setting it to `1` has a possible side-effect of upgrading access to "access only" when someone meant for it to be "nobody" and the code submitted `undef` instead of `0`.

cgi-bin/DW/Media.pm
```diff
@@ -110,9 +110,18 @@ sub upload_media {
...
+    if ( $sec =~ /^(?:friends|access)$/ ) {
+        $sec = 'usemask' ;
+        $opts{allowmask} = 1;
+    }
     croak 'Invalid security for uploaded file.'
         unless $sec =~ /^(?:public|private|usemask)$/;
-    $opts{allowmask} = 0 unless defined $opts{allowmask} && $sec eq 'usemask';
+    if ( $sec eq 'usemask' ) {
+        warn "Must define allowmask when security is usemask"
+            if ! defined $opts{allowmask} && $LJ::IS_DEV_SERVER;
```
zorkian: The prior code would always set it to `0` if it was undefined, in this new code, it lets through `undef`. I think we need to set it to `0` explicitly if it's undefined.

kareila:
Okay, I think we're finally on the same page with this.  Thanks for your patience!

(I should probably add another test for that particular case...)

kareila:
Test case added!

rahaeli:
Is it worth opening a bug for creating a helper class for that kind of check, as discussed above, or do you think that sort of refactoring would run the risk of breaking something too badly to justify doing it?

zorkian:
Thank you! And sorry for the confusion .. running between meetings tanks my reading comprehension sometimes. :'(

And yes, +1 to having a helper that does this logic.

PR #1555, opened 2015-08-25
Switch to SSL fonts

### LINKED ISSUE #1551: Embedded fonts from Google blocked for being non-secure
Report: https://www.dreamwidth.org/support/see_request?id=31469

"Since Dreamwidth is now on HTTPS and Google font serves from HTTP, Chrome and Firefox show Mixed Content warnings and will sometimes block the font from showing."

This is an issue in particular for a few of our official styles which use Google fonts (e.g. Database). My FF always prevents the font from being used.

zorkian:
This switches us to use the Google fonts SSL version, which will fix
mixed content warnings when viewing a Dreamwidth page securely.

Fixes #1551.

### DIFF
--- a/styles/ciel/themes.s2
+++ b/styles/ciel/themes.s2
@@ -753,7 +753,7 @@ set color_navlinks_link_hover_background = "#ffb2d0";
 ## Fonts
 ##===============================
 
-set font_sources= "http://fonts.googleapis.com/css?family=Bonbon";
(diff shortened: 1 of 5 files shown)


kareila:
I've verified that this compiles and properly loads https fonts, but I can't test that it gets rid of mixed content warnings since the development server doesn't support SSL.  I'm sure this is fine, though.

zorkian:
Thanks for the review! :)

PR #1557, opened 2015-08-27
Notifications should provide HTTPS links

### LINKED ISSUE #1354: Update links in automated emails to point to HTTPS
Right now, when I receive an email with, for instance, notification of a new comment, all the links -- links to "Reply at the webpage", "View all comments to this entry", "Go to the top of the thread this comment is part of", my notification settings, and so on -- are HTTP rather than HTTPS.

I presume that we'd prefer to just turn those links into HTTPS links rather than offer an HTTP vs. HTTPS option in https://www.dreamwidth.org/manage/settings/?cat=privacy .

kaberett:
Fixes #1354.

Changes the test condition for when to use https:// in
journal_base in preference to http://, in an extremely naive
fashion; presumably I should in fact be nesting conditions
because we want to preserve LJ::IS_SSL as a test?

Perhaps by analogy to line 35 of https://github.com/dreamwidth/dw-free/blob/29dadd1d2281129247b0e5cc696ea570004c4d0b/cgi-bin/LJ/URI.pm ?

### DIFF
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -641,6 +641,11 @@ sub is_syndicated {
 }
 
 
+sub journal_base {
+    return LJ::journal_base( @_ );
+}
+
+
 sub journaltype {
     return $_[0]->{journaltype};
 }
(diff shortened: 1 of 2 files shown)


kaberett:
nb journal_base is in an absolutely ridiculous place and should probably be moved over to LJ::Account; @Sophira has offered to do this, or I'm happy to handle it if given the go-ahead!

Sophira:
To be precise, I think `LJ::User::journal_base()` and `LJ::journal_base()` should both be moved to `cgi-bin/LJ/User/Account.pm` rather than the file they're currently in, which is `cgi-bin/LJ/User/Styles.pm`.

kareila:
I think what you want here is to return an https link if either of the variables is true, which we would most commonly express using 'or' notation like so: `( $LJ::USE_HTTPS_EVERYWHERE || $LJ::IS_SSL )`

The first says print an https link if we've turned on HTTPS Everywhere (which we haven't yet), and the second says print an https link if the page load requested one (which isn't possible in the context of sending an email notification, but will work in other contexts on the site itself).  If neither of these things is true, an http link will be printed as before.

And yes, go ahead and open an issue to move the method wherever you think it best fits once this is finished. :zap:

kaberett:
Okay, I've correct the if-statement check and I've moved everything over to LJ::User::Account (as two separate commits). Let me know if you'd like me to submit an entire new pull request for the move?

kareila:
Nope, this looks perfect.  Thanks!

PR #1558, opened 2015-08-29
fixes for atom-post.t and post.t

kareila:
When I enabled $LJ::USE_HTTPS_EVERYWHERE in my test config, atom-post.t started failing dramatically, so I investigated and found a few small changes that needed to be made to the test suite to allow it to function as expected.

Simultaneously but unrelatedly, I started noticing occasional non-deterministic failures in post.t caused by the random selection of a mood to test. I think this wasn't always failing before because the order of the array returned by the keys function is more randomized now than it used to be, starting with Perl 5.18.

### DIFF
--- a/t/atom-post.t
+++ b/t/atom-post.t
@@ -38,6 +38,8 @@ use DW::Routing;
 # so that entries can be posted to community journals
 $LJ::EVERYONE_VALID = 1;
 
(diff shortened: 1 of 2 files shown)


rahaeli:
yay fixed tests!

PR #1566, opened 2015-09-02
Fixes #1564. Correctly displays revtime on entryprops page

### LINKED ISSUE #1564: entry props not having revtime set when entry is edited
If a post is edited after posting, the post's entryprops are supposed to get a revtime -- the time the post was edited. Every post that was edited that I've spot-checked just now (as part of trying to debug a separate problem) has a revtime of 1970-01-01 00:00:00, which is what happens when a zero value is converted to time. So, something's not setting the revtime properly.

When trying to reproduce/debug this on your 'hack: to check an entry's revtime, grant your account canview:entryprops and then go to admin/entryprops and input the entry's URL.

sgsabbage:
Fixes #1564.

### DIFF
--- a/cgi-bin/DW/Controller/Admin/Props.pm
+++ b/cgi-bin/DW/Controller/Admin/Props.pm
@@ -106,7 +106,7 @@ sub entryprops_handler {
                     $prop{description} = $prop_meta->{des};
 
                     # an ugly hack, i know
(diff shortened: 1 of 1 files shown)


kareila:
Sorry for the delay in reviewing this, and thanks again for diagnosing the problem!

PR #1567, opened 2015-09-03
(Issue 1529) Makes Tag Merging Case Insensitive

### LINKED ISSUE #1529: tag merging is case-sensitive, tags are not
From the suggestions queue, better just moved over to an issue since it's a bug:

In the Merge Tags page, strange things happen with comparisons to determine whether two tags can be merged into a tag with the same name as one of the user-supplied new tag name has capital letters (as iOS keyboards tend to capitalize words on the fly, this is easy to do there).

For example, I want to merge "beer" and "stout" and supply "Beer" as the new name. The merge facility tells me that this is not allowed as the new tag already exists but is not one of the tags to be merged — evidently the check against the main list is case-insensitive but the one against the merge list isn't.

hotlevel4:
Fixes #1529.


kareila:
This appears to do the job!  Thanks :smile:

PR #1575, opened 2015-09-09
[#1573] don't include suspended users in database query

### LINKED ISSUE #1573: we keep running out of new accounts on /stats
Turns out when there's a flood of spammers and you're camping http://www.dreamwidth.org/stats you can burn through the 20 most recently created accounts pretty quickly when they're all spammers. 

Not sure what to do about it -- fetch more? idk.

kareila:
Fixes #1573.


zorkian:
Huh, so that's a warning now. Neat.

PR #1576, opened 2015-09-10
Update config files

### LINKED ISSUE #983: Merge doc/config-*.pl.txt with etc/config*.pl
Since all site-specific configuration is going to be in etc/local/ext anyway, the files in etc/config*.pl serve more as examples than anything.

When we have a new feature flag, I always forget which one I should put the sample code in...

### LINKED ISSUE #1448: update sample configs in dw-free repo to be current && be more clear they are samples
The configs that we ship with dw-free are samples, not the actual configs that are live on dreamwidth.org (for several very good reasons!) However, the samples are kind of out of date (missing a few new features) and they aren't very clear that they are samples.

So, we need to do two things:

a) Make sure every flag/feature/option/etc that's in the production config files has a corresponding entry in the sample configs in the repo;

b) Put a comment/line in the sample configs in dw-free to make it more clear that these are samples and you should probably change them.

kareila:
The usual Karzilla Smash.

Fixes #1448 and also fixes #983.


Sophira:
This will affect the Dreamhack install process as it works by copying the example configs and modifying them, so I'll take a look into that tomorrow.

PR #1579, opened 2015-09-27
Updated wiki location

azurelunatic:
Official wiki location has moved from wiki.dwscoalition.org to wiki.dreamwidth.net.

### DIFF
--- a/htdocs/site/opensource.bml.text
+++ b/htdocs/site/opensource.bml.text
@@ -9,7 +9,7 @@
 
 .development.header=Development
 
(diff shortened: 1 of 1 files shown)


kareila:
It looks like the translation string will need a new key in order for the change to be reflected on the site.  See https://github.com/dreamwidth/dw-free/pull/1494#issuecomment-122273486 for detailed instructions, and ping one of us if you have questions.

kareila:
Related to #1620

kareila:
It appears that this change was made in another commit elsewhere, so I'm closing this.  Thanks for working on it!

PR #1580, opened 2015-09-27
Updated wiki location, again

azurelunatic:
dreamwidth.net all over the place

### DIFF
--- a/htdocs/doc/s2/index.bml
+++ b/htdocs/doc/s2/index.bml
@@ -9,7 +9,7 @@
 #      Ninety Degrees <ninetyd.bis@gmail.com>
 #      Azure Jane Lunatic <azurelunatic@gmail.com>
 #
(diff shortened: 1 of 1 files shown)


kareila:
This is an extremely minor thing, but instead of changing the copyright date from 2010 to 2015, you should use the range 2010-2015. Otherwise people might think we are creating new BML pages. :wink: 

The good news is that the links don't appear to go through the translation system, so apart from that tiny copyright tweak I think this is good to go.

kareila:
Related to #1620

kareila:
Thanks for updating this. I would love to see this squished down into one commit - it's not difficult to do and I'd be happy to walk you through the process at some point.  It's outlined here: https://github.com/dreamwidth/dw-free/pull/1506#issuecomment-128172320

kareila:
I've merged this one manually as well.  Thanks!

PR #1581, opened 2015-09-27
Update CONTRIBUTING.md with current wiki location

azurelunatic:
wiki.dwscoalition.org now canonically wiki.dreamwidth.net


kareila:
Do we have an issue open for this? There are also a few wiki links in the bin/ directory and one in Build.PL that need to be updated as well.

azurelunatic:
Filed #1620 just now.

PR #1584, opened 2015-09-27
Fixes #1482. Clarify header options for Heads Up and Core2

### LINKED ISSUE #1482: Clarify header image options in Heads Up style
Related to: https://www.dreamwidth.org/support/see_request?id=31182
If I got things right, Heads Up has two options for header images: the standard one for a background picture and a custom one for an image which would go on top of the background. The way the two options are described is a bit confusing though as it isn't clear which applies to which image.
For example compare the two height options:

"The height of your header, in pixels. Use 0 for default."
"The height of your header image, in pixels."

If you're not familiar with how the style is code you can think you have to set the two options even if you only have one background pic. That's where things go wrong and title and subtitle disappears.

I think the Core2 option should state explicitly that it's for the *background* picture --this would also prevent people from thinking this can always work as a banner-- and the Heads Up options should make it clearer that this is for an additional picture.

ninetyd:
Fixes #1482 
Better property descriptions in Heads Up and Core2
Don't set height if there's no foreground image
Switch option order for better clarity
Sort width option into Header  w/ other options
Move noted to note field
Remove unneeded note (there's no reason to expect pics to be resized)

### DIFF
--- a/cgi-bin/LJ/S2Theme/headsup.pm
+++ b/cgi-bin/LJ/S2Theme/headsup.pm
@@ -17,9 +17,10 @@ sub header_props {
     my $self = shift;
     my @props = qw(
         image_foreground_header_url
-        int image_foreground_header_height
-        string image_foreground_header_alignment
+        image_foreground_header_height
+        image_foreground_header_width
         image_foreground_header_position
+        string image_foreground_header_alignment
         image_foreground_header_alt
     );
     return $self->_append_props( "header_props", @props );
--- a/styles/headsup/layout.s2
+++ b/styles/headsup/layout.s2
@@ -47,33 +47,35 @@ propgroup colors_child {
 propgroup images_child {
 
 property string image_foreground_header_url {
-   des = "The URL to the header image (this style does not automatically resize images)";
+    des = "The URL to the additional header image";
+    note = "This image goes on top of the background one if it exists.";
 }
 
 property int image_foreground_header_height {
-    des = "The height of your header image, in pixels.";
+    des = "The height of your additional header image, in pixels.";
     example = "100";
     size = 6;
 }
 
 property int image_foreground_header_width {
-    des = "The width of your header image, in pixels.";
+    des = "The width of your additional header image, in pixels.";
     example = "600";
     size = 6;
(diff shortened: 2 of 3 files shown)


cgi-bin/LJ/S2Theme/headsup.pm
```diff
@@ -17,9 +17,10 @@ sub header_props {
     my $self = shift;
     my @props = qw(
         image_foreground_header_url
-        int image_foreground_header_height
-        string image_foreground_header_alignment
+        image_foreground_header_height
```
kareila: Forgive me for taking so long to review this - I'm not as familiar with the conventions here.  Why was the "int" removed from the list here - does it make a difference at all?

styles/headsup/layout.s2
```diff
@@ -225,12 +227,14 @@ var string header_background_height;
     }
 
 var string headerimage_area_height;
```
kareila: Does the indentation on this "var string" need to be corrected?

cgi-bin/LJ/S2Theme/headsup.pm
ninetyd: I should have explained, sorry. We don't mention variable type (int, string, whatever) when sorting new properties. This is just clean up and making sure this file is consistent with other similar files.
And now I've just noticed I left the 'string' for image_foreground_header_alignment. :(

styles/headsup/layout.s2
ninetyd: I think it's the same indentation as the other variables set in function print_stylesheet () so no.

cgi-bin/LJ/S2Theme/headsup.pm
kareila: Do you want to go ahead and remove it before I commit this?

ninetyd: Will do. Sorry for the late answer.

kareila:
Let me know when this is ready for another look. Thanks!

zorkian:
(Form letter incoming.)

PR #1594, opened 2015-10-14
[#1593] let email sysban_check match email_domain sysbans with 3 elements

### LINKED ISSUE #1593: three-element domains are not being checked against email address sysbans
The code for sysban_check only looks at the last two elements of the domain name when given an email address.  If we've defined a three-element domain for sysban, that should be matched as well.

kareila:
Also update t/console-sysban.t to test expected behavior.

Fixes #1593.

### DIFF
--- a/cgi-bin/LJ/Sysban.pm
+++ b/cgi-bin/LJ/Sysban.pm
@@ -209,6 +209,12 @@ sub sysban_check {
         return 0 unless scalar @domains >= 2;
         my $domain = "$domains[-2].$domains[-1]";
         return 1 if $check->('email_domain', $domain);
+        # also check for three-element domain bans
+        if ( defined $domains[-3] ) {
+            # don't overwrite $domain here - two-element domain checked below
+            my $triple = "$domains[-3].$domains[-2].$domains[-1]";
+            return 1 if $check->( 'email_domain', $triple );
+        }
 
         # account for GMail troll tricks
         if ( $domain eq "gmail.com" ) {
--- a/cgi-bin/LJ/Sysban.pm
+++ b/cgi-bin/LJ/Sysban.pm
@@ -198,6 +198,42 @@ sub sysban_check {
             }, undef, $wh, $vl);
     };
 
+    # helper for iteratively checking all subdomains of a hostname
+    my $check_all_subdomains = sub {
+        my ( $host ) = @_;
+
+        # break the hostname into domain components
+        my @domains = split( /\./, $host );
+        my $length = scalar @domains;
+        return 0 unless $length >= 2;
+
+        # build up the list of subdomains to check for bans
+        # for example, if the full hostname is www.hack.dreamwidth.net,
+        # this will also check hack.dreamwidth.net and dreamwidth.net
+        my @check;
+        for ( my $i = 0; $i < $length; $i++ ) {
+            my $part = $domains[$i];
+            @check = map { "$_.$part" } @check;
+            push @check, $part unless $i == $length - 1;  # don't check TLD
+        }
+
+        # save database overhead by hashing all the email_domain sysbans in one query
+        my $banned = $dbr->selectall_arrayref( qq{
+                SELECT value FROM sysban
+                WHERE status = 'active'
+                  AND what = ?
+                  AND NOW() > bandate
+                  AND (NOW() < banuntil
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/Sysban.pm
```diff
@@ -209,6 +209,12 @@ sub sysban_check {
         return 0 unless scalar @domains >= 2;
         my $domain = "$domains[-2].$domains[-1]";
         return 1 if $check->('email_domain', $domain);
+        # also check for three-element domain bans
+        if ( defined $domains[-3] ) {
```
zorkian: This should be a length check as the previous check. `scalar @domains >= 3` kind of thing.

cgi-bin/LJ/Sysban.pm
```diff
@@ -209,6 +209,12 @@ sub sysban_check {
         return 0 unless scalar @domains >= 2;
         my $domain = "$domains[-2].$domains[-1]";
         return 1 if $check->('email_domain', $domain);
+        # also check for three-element domain bans
```
zorkian: This whole section has gotten messy. Wouldn't this work:

```
# $1 is the whole domain, $2 is the last two parts
return 0 unless $value =~ /@((?:.+?)?([^.]+\.[^.]+))$/;
return 1 if $check->( email_domain => $2 ) || $check->( email_domain => $1 );
```

alierak: I don't think this will work for cases where you have a 3-element domain in the list, and you want to match a subdomain of it. "The last two parts" or "the last three parts" are both special cases of "longest matching suffix" in my opinion.

kareila: I agree that there is probably a more elegant solution to be found here, but I don't think replacing a simple split expression with a complicated regular expression is a good idea for comprehensibility and maintainability.  Also, this won't work for all cases - if a user gives us a four element subdomain, and we've implemented a sysban based on the last three elements, I believe this check will still fail.

kareila:
Here's a new solution that should work in all cases.

cgi-bin/LJ/Sysban.pm
zorkian: Yes, the rules as I understood them (possibly wrong):

* If "foo.com" is specified, ban it and any/all subdomains of "foo.com" (three elements, ten elements, whatever)
* If "a.foo.com" is specified, ban it specifically

I.e. there is a short form (broad ban) and a long form (specific ban). After reading @alierak's comment I agree that probably the better rule is simply "a ban of a domain includes that domain and all subdomains", which is clearer. Although slightly more annoying to implement, and my solution certainly won't work there.

@kareila Let's not to go down a path of writing 50 lines of code because each line is "simple" and thinking that 50 lines of simple code is going to be "more maintainable" than 5 lines of more complex code. I would strongly disagree on that matter.

cgi-bin/LJ/Sysban.pm
```diff
@@ -198,6 +198,42 @@ sub sysban_check {
             }, undef, $wh, $vl);
     };
 
+    # helper for iteratively checking all subdomains of a hostname
+    my $check_all_subdomains = sub {
```
zorkian: Since we're going to hit the database and load every matching element, and we're doing this just for `email_domain` types, then we should instead of doing all this magic just:

```
SELECT value FROM sysban
WHERE status = 'active'
     AND what = ?
     AND (value = ? OR ? LIKE CONCAT("%.", value))
     AND NOW() > bandate
     AND (NOW() < banuntil
              OR banuntil = 0
              OR banuntil IS NULL)
```

And pass `('email_domain', $domain, $domain)` (yes twice, to allow exact matches, and to prevent "foobar.com" matching against "bar.com").

This gets rid of nearly all of the Perl code and makes the database do it for us. Thoughts?

kareila: It took me most of an hour to wrap my brain around what this was doing, but I think it is essentially correct - it takes every active, unexpired domain ban in the database and looks for either an exact match to the given host, or a wildcard match for a larger domain of which the given host is a subdomain.  I can use this, but I should probably add a comment or two... :blush:

kareila:
Updated... hope this does it.

zorkian:
100%! Thank you for your patience :)

Issue #1544, opened 2015-08-16
RTE sorts external sites alphabetically by character, capitals first


azurelunatic:
The [insert] Account control in the RTE offers a list of External Accounts. Since everyone but me managed to capitalize the external site names, lower-case "github" is at the bottom of the list. 

a) we could just find out where it's pulling from and capitalize Github. 
b) it's the RTE on the old update page. Is it really worth going after with pitchforks and making it sort the list differently?

azurelunatic:
It's probably in dw-free/cgi-bin/DW/External/Site/Github.pm 

line 19 is:

package DW::External::Site::github;

Would it break anything if I just capitalized it??

kaberett:
nope, that's where I'd go to kick it, and capitalising it should make shit Just Work provided you check that it's capitalised consistently anywhere that's referred to (i.e. External::Site)

fastbean-au:
I'll claim this.

azurelunatic:
Woohoo! Go for it!

kareila:
Fixed by #1598.

PR #1598, opened 2015-10-27
[#1544] fix package name and site name to allow to display and be pla…

fastbean-au:
…ced correctly.

The name github changed to GitHub, including the file name, package name, and the service name as defined in DW::External::Site.


zorkian:
Thanks for doing this!

Issue #1592, opened 2015-10-10
Timestamp sticky ticky unset globally if you edit a post


kaberett:
The "use the time this entry was posted" checkbox was turned into a sticky ticky back around June, i.e. the beta Create Entries page remembered whether its last state was checked or unchecked, so you didn't have to tick it every time if you always wanted posts to be stamped with the time of posting.

Issue #1480 was opened because that checkbox should *not* persist as checked when editing entries (otherwise you end up editing entries from six months ago and having them marked as the present).

However, I've just noticed that the PR I made to prevent the checkbox from autofilling when editing entries results in the following:
- you create a new entry, and check the "use time at posting"
- if you create another new entry, the "use time at posting" box is checked
- then edit an entry; "use time at posting" is unchecked
- create a new entry; "use time at posting" is unchecked

The sticky ticky for creating new entries should persist regardless of whether you've edited entries (without checking the box) in the interim. Sorry for introducing this one!

chrisboyle:
claim

chrisboyle:
claim again? (now I'm in the org maybe it will assign to me?)

PR #1600, opened 2015-11-07
Only persist the ticky when posting, not editing.

chrisboyle:
Fixes: #1592

I haven't yet submitted a CLA but will do so today.


kareila:
Excellent, thank you!

PR #1601, opened 2015-11-07
Hijack the proxy HTTP response and close it when we're done (fixes #1381 maybe)

### LINKED ISSUE #1381: Proxy causes system to run out of file descriptors when left running too long

sgsabbage:
Potentially fixes #1381, more notes in the ticket.

### DIFF
--- a/src/proxy/main.go
+++ b/src/proxy/main.go
@@ -101,6 +101,19 @@ func defaultHandler(w http.ResponseWriter, req *http.Request) {
 	// SOURCE is ignored programmatically; it's only for admins
 	parts := strings.SplitN(req.URL.RequestURI(), "/", 5)
 
+	// Close the connection when we're done
+	defer func() {
+		hj, ok := w.(http.Hijacker)
+		if !ok {
+			return
+		}
+		conn, _, err := hj.Hijack()
+		if err != nil {
+			return
+		}
+		conn.Close()
+	} ()
+
 	if len(parts) != 5 || parts[0] != "" {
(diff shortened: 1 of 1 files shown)


kareila:
Thanks for working on this!  I will need @zorkian or @afuna to review it since I haven't learned Go syntax.

zorkian:
Hi,

So basically the problem is actually that we just never configured anything other than the default Linux file descriptor limit (1024). Since it's so trivial to have more, we should just raise it to 100,000 or something ridiculous.

We don't want to just arbitrarily close connections like this because, without modifying the HTTP response of the prior response, you'll cause errors to the browser in the case where they're expecting the connection to stay open and they send a second request. Meanwhile you've closed the connection and sadness happens.

At best, this just leads to the browser opening a new connection and retrying (which can be quite slow from remote areas) but at worst this will cause the user to see images failing to load.

I'll make a change to our production configuration to raise the ulimit for the process and am, regretfully, going to close this PR. I do appreciate you taking this on and working at it, though. Thank you.

Issue #1333, opened 2015-04-02
tracking dialog does not fit on screen in Firefox 33.0 for Android


rahaeli:
http://www.dreamwidth.org/support/see_request?id=29743

From @swaldman3 in the request:

Problem: When viewed on a mobile phone screen, the dialog to track a thread does not fit on the screen. 

Tested on Firefox 33.0 for Android, on a Nexus 4. 

To reproduce: 
1. Visit an entry 
2. At the bottom of the page, click "track this". 
3. Note that the dialog that pops up is wider than the screen 
(see screenshot at https://www.dropbox.com/s/86bzy2ll4fpugj0/Screenshot_2014-11-02-20-58-03.png?dl=0) 
4. Note that the browser does not allow you to zoom out to view the whole dialog - presumably because it thinks the width of the text column is as far out as anybody would want to go. 

From someone in an IC:

I tried reproducing this on my iphone 4 and in Safari on a few different layouts. The results varied. 

Plain for Tabula Rasa: http://i1363.photobucket.com/albums/r707/trollanort/Misc/track4.png~original 
This got me the same result as the user. I was unable to zoom out at all, and most of the text was cut off and aligned to the far left. 

Velvet Steel for Practicality: http://i1363.photobucket.com/albums/r707/trollanort/Misc/track3.png~original 
Worked as it should. It resized the pop-up to fix exactly within the size of the screen automatically. 

Carried Away for Trifecta: http://i1363.photobucket.com/albums/r707/trollanort/Misc/track2.png~original 
I can zoom out on this one to see the entire box, but it keeps aligning itself to the right for some reason. 

Basic for Transmogrified: http://i1363.photobucket.com/albums/r707/trollanort/Misc/track1.png~original 
It allows me to zoom out and the box gets smaller, but still remains half off-screen and aligned far to the left.

chrisboyle:
claim

chrisboyle:
In Firefox 42 on Android and Chrome 46 on Android (those being current), I'm not seeing as serious a problem as the original report: the dialog still extends further right than the rest of the document but the scrollable area now extends to include it. Perhaps those browsers got smarter?

I have a change that makes max-width successfully apply so the dialog is sized to fit on the screen (whereas the current attempt in jquery.esn.js does not actually result in any such CSS).

chrisboyle:
claim again (to get assigned now I'm in the org)

PR #1605, opened 2015-11-07
Fix max-width of track dialog.

chrisboyle:
jQuery dialog's maxWidth only takes an int as pixels per their docs; I
don't see how percentage would ever have suceeded there.

Fixes: #1333

I haven't yet submitted a CLA but will do so today.


rahaeli:
Just confirming that I do have Chris's CLA now!

zorkian:
I have no idea. But thank you for fixing!

Issue #1465, opened 2015-06-29
console reference should say what priv is required for each command


rahaeli:
https://www.dreamwidth.org/admin/console/reference explains the commands, what they do, and what their args are, but doesn't say anything about what priv is required to use them. (They do grey-out if you don't have access to the command, but it doesn't tell you what priv you need.)

The main admin page at https://www.dreamwidth.org/admin tells you what privs you need for each admin page, which is super helpful. So, let's extend that to the console reference.

For each command in cgi-bin/LJ/Console/Command, the text in 'desc' should be extended to also include:

`Requires priv: <priv>`

So, for cgi-bin/LJ/Console/Command/ResetEmail.pm, one would change

`sub desc { "Resets the email address of a given account." }`

to:

`sub desc { "Resets the email address of a given account. Requires priv: reset_email." }`

Doing it that way will mean that people with the priv to use the command will see what priv it requires, but that's a feature, not a bug. For commands such as ban_set and ban_unset, the message should be "Requires priv: none."

cesy:
Completed :)

kareila:
Not quite finished - the following still need to have their privs noted:
- [ ] `cgi-bin/LJ/Console/Command/MoodthemeList.pm`
- [ ] `cgi-bin/DW/Console/Command/BonusIcons.pm`
- [ ] `cgi-bin/DW/Console/Command/ManageCircle.pm`
- [ ] `cgi-bin/DW/Console/Command/Note.pm`
- [ ] `cgi-bin/DW/Console/Command/RenameOpts.pm`
- [ ] `cgi-bin/DW/Console/Command/ResetPBEMToken.pm`
- [ ] `cgi-bin/DW/Console/Command/RevokeRenameToken.pm`

Once we've rounded up the last few stragglers, we can close this.  Thanks for all your effort so far! :+1:

PR #1606, opened 2015-11-07
#1465 privs listed for console commands

cesy:
Labels the privs required for admin console commands. Now completed.


kareila:
Thanks for doing these!

PR #1610, opened 2015-11-07
1568 consistent username length limit

swaldman3:
Partially fixes bug 1568.
- The specific cases that Karelia mentioned now use $LJ::USERNAME_MAXLENGTH instead of a hardcoded value, except for tools/textmessage where I've just hardcoded the new value - I haven't spent the time to do this one properly because I think this should probably have the code machete applied anyway!
- There are other places where the max length is still hardcoded (although at 25 or greater, so it isn't an immediate problem). There's definitely the login dialog, and possibly also when commenting as a user, when searching for a user, when renaming, and when doing other things that I didn't think of. I have not attacked these because I encountered LJ::Widgets and lost the will to live, so you may want to leave the Issue open in case somebody else wants to investigate.

**NB To push to production, you also need to add a line `$USERNAME_MAXLENGTH = 25;` to config.pl.**

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -843,6 +843,8 @@ error.usernameinvalid=This account name contains invalid characters.
 
 error.usernamelong=This account name is too long.  Account names can't be longer than 25 characters.
 
+error.usernamelong1=This account name is too long. Account names can't be longer than [[maxlength]] characters.
+
 error.username_notfound=This account name wasn't found.
 
 error.vhost.noalias=This user's account type doesn't permit domain aliasing.
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -841,8 +841,6 @@ error.unknownmode=Unknown mode.
 
 error.usernameinvalid=This account name contains invalid characters.
 
-error.usernamelong=This account name is too long.  Account names can't be longer than 25 characters.
-
 error.usernamelong1=This account name is too long. Account names can't be longer than [[maxlength]] characters.
 
 error.username_notfound=This account name wasn't found.
--- a/cgi-bin/LJ/CreatePage.pm
+++ b/cgi-bin/LJ/CreatePage.pm
@@ -33,8 +33,9 @@ sub verify_username {
     if ($given_username && !$user) {
         $error = LJ::Lang::ml('error.usernameinvalid');
     }
-    if (length $given_username > 25) {
-        $error = LJ::Lang::ml('error.usernamelong');
+    if (length $given_username > $LJ::USERNAME_MAXLENGTH) {
+        $error = LJ::Lang::ml('error.usernamelong1', 
+            { maxlength => $LJ::USERNAME_MAXLENGTH });
     }
 
     my $u = LJ::load_user($user);
--- a/cgi-bin/LJ/CreatePage.pm
+++ b/cgi-bin/LJ/CreatePage.pm
@@ -33,7 +33,7 @@ sub verify_username {
     if ($given_username && !$user) {
         $error = LJ::Lang::ml('error.usernameinvalid');
     }
-    if (length $given_username > $LJ::USERNAME_MAXLENGTH) {
+    if ( length $given_username > $LJ::USERNAME_MAXLENGTH ) {
         $error = LJ::Lang::ml('error.usernamelong1', 
             { maxlength => $LJ::USERNAME_MAXLENGTH });
     }
--- a/cgi-bin/LJ/Session.pm
+++ b/cgi-bin/LJ/Session.pm
@@ -950,8 +950,8 @@ sub get_cookie_path {
 
 sub get_path_user {
     my ( $path ) = @_;
-    return unless $path =~ m!^/(\w{1,25})\b!;
-    # FIXME: username length should be a named constant (#1568)
+    my $exp = '^/(\w{1,' . $LJ::USERNAME_MAXLENGTH . '})\b';
+    return unless $path =~ /$exp/;
 
     return lc $1;
 }
--- a/cgi-bin/LJ/Session.pm
+++ b/cgi-bin/LJ/Session.pm
@@ -950,8 +950,7 @@ sub get_cookie_path {
 
 sub get_path_user {
     my ( $path ) = @_;
-    my $exp = '^/(\w{1,' . $LJ::USERNAME_MAXLENGTH . '})\b';
-    return unless $path =~ /$exp/;
+    return unless $path =~ m!^/(\w{1,$LJ::USERNAME_MAXLENGTH})\b!;
 
     return lc $1;
 }
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -1242,7 +1242,7 @@ sub canonical_username {
     my $input = lc( $_[0] );
     my $user = "";
     my $exp = '^\s*([a-z0-9_\-]{1,' . $LJ::USERNAME_MAXLENGTH . '})\s*$';
-    if ( $input =~ /$exp/ ) {  # good username
+    if ( $input =~ m/^\s*([a-z0-9_z\-]{1,$LJ::USERNAME_MAXLENGTH})\s*$/ ) {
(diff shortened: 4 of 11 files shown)


kareila:
Don't have time to look over this now, but I can tell you that the USERNAME_LENGTH shouldn't be required in config.pl; it should go in LJ/Global/Defaults.pm, which the server admin can then override in config.pl if desired.

swaldman3:
AH! Thank you. There was a conversation on IRC this morning along the lines of "config.pl feels like the wrong place, but is there another place for globals?", and nobody knew :)

swaldman3:
@kareila Fixed! :)

azurelunatic:
(put at least a mention of it into the wiki, on http://wiki.dreamwidth.net/wiki/index.php/Directory_Structure#etc -- feel free to move around as makes sense.)

bin/upgrading/en.dat
```diff
@@ -835,6 +835,8 @@ error.usernameinvalid=This account name contains invalid characters.
 
 error.usernamelong=This account name is too long.  Account names can't be longer than 25 characters.
```
kareila: If you're going to replace a translation string and add the old version to deadphrases.dat, it also needs to be removed from en.dat.

cgi-bin/LJ/Session.pm
```diff
@@ -950,8 +950,8 @@ sub get_cookie_path {
 
 sub get_path_user {
     my ( $path ) = @_;
-    return unless $path =~ m!^/(\w{1,25})\b!;
-    # FIXME: username length should be a named constant (#1568)
+    my $exp = '^/(\w{1,' . $LJ::USERNAME_MAXLENGTH . '})\b';
+    return unless $path =~ /$exp/;
```
kareila: You don't need to break out the variable from the regular expression. It can be interpreted inline by Perl just fine:
`m!^/(\w{1,$LJ::USERNAME_MAXLENGTH})\b!`

cgi-bin/LJ/User/Account.pm
```diff
@@ -1241,7 +1241,8 @@ use Carp;
 sub canonical_username {
     my $input = lc( $_[0] );
     my $user = "";
-    if ( $input =~ /^\s*([a-z0-9_\-]{1,25})\s*$/ ) {  # good username
+    my $exp = '^\s*([a-z0-9_\-]{1,' . $LJ::USERNAME_MAXLENGTH . '})\s*$';
+    if ( $input =~ /$exp/ ) {  # good username
         $user = $1;
```
kareila: Same as above.

htdocs/tools/tellafriend.bml
```diff
@@ -187,7 +187,8 @@ _c?>
             : BML::ml( ".email.subject.entrynosubject", { sitenameshort => $LJ::SITENAMESHORT } );
  }
 
- if ($GET{'user'} =~ /^\w{1,15}$/) {
+ my $exp = '^\w{1,' . $LJ::USERNAME_MAXLENGTH . '}$';
+ if ($GET{'user'} =~ /$exp/) {
      my $user = $GET{'user'};
```
kareila: Same as above.

htdocs/tools/textmessage.bml
```diff
@@ -39,7 +39,7 @@ body<=
     unless ($user) {
         $ret .= "<?h1 $ML{'.enter.user.head'} h1?>".
             "<?p " . BML::ml( '.enter.user.text3', { sitenameshort => $LJ::SITENAMESHORT } ) . " p?>".
-            "<div style='margin-left: 40px'><form method='get' action='textmessage'>$ML{'.enter.user.input'} <input type='text' size='15' maxlength='15' name='user' /> <input type='submit' value=\"$ML{'.enter.user.submit'}\" /></form></div>".
+            "<div style='margin-left: 40px'><form method='get' action='textmessage'>$ML{'.enter.user.input'} <input type='text' size='25' maxlength='25' name='user' /> <input type='submit' value=\"$ML{'.enter.user.submit'}\" /></form></div>".
             "<p>".BML::ml('.setup.text', { aopts => "href='$LJ::SITEROOT/manage/profile/'" })."</p>";
```
kareila: If you decide not to add the variable here in this pass, please at least add a comment (similar to the one I used in cgi-bin/LJ/Session.pm) to help people remember to fix it when it gets converted to TT.

kareila:
This looks like a good first pass.  If you'll take care of the requested syntax tweaks etc., I can try to get this in and then see if I can find any more adjustments that need to be made. The conflicts reported by Github should go away once deadphrases.dat is updated to include recent additions on the develop branch.

cgi-bin/LJ/CreatePage.pm
```diff
@@ -33,8 +33,9 @@ sub verify_username {
     if ($given_username && !$user) {
         $error = LJ::Lang::ml('error.usernameinvalid');
     }
-    if (length $given_username > 25) {
-        $error = LJ::Lang::ml('error.usernamelong');
+    if (length $given_username > $LJ::USERNAME_MAXLENGTH) {
+        $error = LJ::Lang::ml('error.usernamelong1',
```
kareila: Don't forget that the DW code convention preference is for spaces within parentheses like these:
 `( length $given_username > $LJ::USERNAME_MAXLENGTH )`

I realize they weren't there before, but if you're editing a line you should add them. It aids readability.

swaldman3: Hmm, I had been working on the basis of "leave things as they are in the surrounding code". But, fine :)

cgi-bin/LJ/Session.pm
swaldman3: I'll take your word for it, but I'm afraid that line is totally illegible to me! A hazard of not using Perl very often ;)

swaldman3:
Rebased off current develop, and made changes as requested.

swaldman3:
@kareila *ping* not sure if you'll have seen that I made the requested changes.

cgi-bin/LJ/User/Account.pm
```diff
@@ -1241,7 +1241,8 @@ use Carp;
 sub canonical_username {
     my $input = lc( $_[0] );
     my $user = "";
-    if ( $input =~ /^\s*([a-z0-9_\-]{1,25})\s*$/ ) {  # good username
+    my $exp = '^\s*([a-z0-9_\-]{1,' . $LJ::USERNAME_MAXLENGTH . '})\s*$';
```
zorkian: What is `$exp` for here?

etc/config.pl
```diff
@@ -683,6 +683,8 @@
         comment_html_anon => 1,      # HTML comments from anon commenters?
         comment_html_auth => 0,      # HTML comments from non-anon commenters?
     );
+
+
```
zorkian: These lines seem spurious?

htdocs/tools/textmessage.bml
```diff
@@ -39,7 +39,9 @@ body<=
     unless ($user) {
         $ret .= "<?h1 $ML{'.enter.user.head'} h1?>".
             "<?p " . BML::ml( '.enter.user.text3', { sitenameshort => $LJ::SITENAMESHORT } ) . " p?>".
-            "<div style='margin-left: 40px'><form method='get' action='textmessage'>$ML{'.enter.user.input'} <input type='text' size='15' maxlength='15' name='user' /> <input type='submit' value=\"$ML{'.enter.user.submit'}\" /></form></div>".
+            "<div style='margin-left: 40px'><form method='get' action='textmessage'>$ML{'.enter.user.input'} <input type='text' size='25' maxlength='25' name='user' /> <input type='submit' value=\"$ML{'.enter.user.submit'}\" /></form></div>".
+            #NB the maxlength of the username text box should be based on
+            #LJ::USERNAME_MAXLENGTH, not hardcoded here.
```
zorkian: You can use `$LJ::USERNAME_MAXLENGTH` here. This is in a BML code block, so you can just reference it.

zorkian:
Almost ready -- just some comments and then I think we're good to go!

zorkian:
(Form letter incoming.)

PR #1611, opened 2015-11-07
Fixes #1306, adds quickreply to day, recent and network

### LINKED ISSUE #1306: add quickerreply to other views (lastn, day, etc)
People like it on the reading page, so let's go ahead and extend it to the other views.

sgsabbage:
Fixes #1306.

Added a convenience method for checking if a page should have quickreply added, added the needed bits to the s2 core layout. Hopefully that's everything!

Note that I've used minimal everywhere except entry (so existing behaviour is unchanged) as it's styled better for inline replies in lists of entries.

It may well be that we don't need to check which view we're on anymore, but I've taken a cautious approach as it's likely there are scenarios I haven't considered.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2670,6 +2670,15 @@ sub sitescheme_secs_to_iso {
 sub current_box_type {}
 sub curr_page_supports_ebox { 0 }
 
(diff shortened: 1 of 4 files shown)


kareila:
:100: :+1:

PR #1614, opened 2015-11-07
Removing redundant colon from mobile post page

azurelunatic:
this has been getting up my nose for years

### DIFF
--- a/htdocs/mobile/post.bml
+++ b/htdocs/mobile/post.bml
@@ -90,7 +90,7 @@ _code?>
 
 <form method='post' action='/mobile/post'>
 <?_code { return LJ::form_auth(); } _code?>
-<?_ml .form.subject _ml?>:<br />
+<?_ml .form.subject2 _ml?>:<br />
 <input name='subject' size='50' style='width: 90%' /><br />
 <?_ml .form.post _ml?>:<br />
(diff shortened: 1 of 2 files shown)


kareila:
The proper ritual sacrifices need to be made for the translation system: you need to change the key to `.form.subject2` and so on. :bee: :bee:

rahaeli:
Almost there! When you increment a string like that, you also need to change the reference in the file itself to invoke .form.subject2 instead of .form.subject.

azurelunatic:
Whee! (oh god this is terrifying)

rahaeli:
You're doing great! It's not your fault the translation system is an eldritch horror!

azurelunatic:
Tonight's Storytime with Azz in -dev was explaining to @sgsabbage Why We Change The Code String Via Patch. It was all very baffling ("Because the steam-radiator still sits there angrily waiting for us to bark our shins on it in the middle of the night.") up until the moment he beheld the Translation UI, at which point he achieved enlightenment.

azurelunatic:
(it's possible that I could *also* have removed the *other* colon from the file itself, but that would have been simple and easy. And also incorrect.)

kareila:
I merged this manually so that I could squash down the commits. Thanks for fixing this!

PR #1615, opened 2015-11-07
Use user display name in circle page

### LINKED ISSUE #1608: Page title "ext_1 in your circle" when circling an OpenID user
The title of `/circle/ext_1/edit` is "ext_1 in your circle". There is a correct title in the body of the page: "joeblogs.example.org in your circle".

me-and:
Use user's display name, rather than their user name, on the page for
editing circle membership.  This isn't the display name according to the
profile page; it means that for Dreamwidth users, their username appears
in the title, but for OpenID users, rather than "ext_{number}"
appearing, their OpenID URL appears.

Fixes #1608.


zorkian:
LGTM

PR #1616, opened 2015-11-07
Don't suggest editing deleted entries

### LINKED ISSUE #1613: Entries deleted via http://www.dreamwidth.org/editjournal offer the option to "Edit this entry again" on Success page
They shouldn't.

kaberett:
Fixes #1613.

(There was never anything incorrect here and @me-and definitely didn't go "... wait, what?" at me.)


zorkian:
Hah. Yeah. It happens.

PR #1388, opened 2015-05-17
Additional information on entry splash page

woggy:
Pulls in the posted security level and subject for the entry and displays them to the user on the splash page. Functional for both old and new-style update pages.

Fixes #1231

PR #1617, opened 2015-11-07
More detailed subject on Success landing page

kaberett:
Adds "(no subject)" to extradata line for subject if entry
has no subject, instead of leaving blank.

Prevents overescaping HTML entities (e.g. \&lt;3), by analogy
with #1453 and fixes therefore, for TT only.

(No issue for this specifically, but it came out of conversations around #1388.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -1038,7 +1038,7 @@ sub _do_post {
         my $extradata = {
             security => $form_req->{security},
             security_ml => "",
-            subject => LJ::ehtml( $form_req->{subject} ),
+            subject => $form_req->{subject},
         };
         if ( $extradata->{security} eq "usemask" ) {
             if ( $form_req -> {allowmask} == 1 ) {
--- a/htdocs/update.bml.text
+++ b/htdocs/update.bml.text
@@ -147,5 +147,9 @@
 
 .extradata.sec.custom=The entry was posted with custom access.
 
-.extradata.subject=The entry was posted with the following subject: [[subject]]
+.extradata.subj=The entry was posted with the following subject: 
+
+.extradata.subject.no_subject=(no subject)
+
+.extradata.subject.subject=[[subject]]
 
--- a/views/entry/success.tt
+++ b/views/entry/success.tt
@@ -24,8 +24,14 @@ the same terms as Perl itself.  For a copy of the license, please reference
 [%- END -%]
 
 [%- IF poststatus.ml_string != ".edit.delete" -%]
+
 <p>[% extradata.security_ml | ml %]</p>
-<p>[% ".extradata.subject" | ml( subject => extradata.subject) %]</p>
+<p>[% ".extradata.subj" | ml %]
+    [% IF extradata.subject.length > 0 %]
+        [% ".extradata.subject.subject" | ml( subject => extradata.subject ) %]
+    [% ELSE %]
+        [% ".extradata.subject.no_subject" | ml %]
+    [% END %]</p>
 [%- END -%]
 
 [%- IF warnings.exist -%]
--- a/views/entry/success.tt.text
+++ b/views/entry/success.tt.text
@@ -48,4 +48,8 @@
 
 .extradata.sec.custom=The entry was posted with custom access.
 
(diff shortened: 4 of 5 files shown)


kareila:
Thanks for working on this and for keeping the behavior of the BML and TT versions consistent with each other!

I see one minor issue with this approach which I will point out in line comments, but otherwise looks good to me.

htdocs/update.bml.text
```diff
@@ -147,5 +147,9 @@
 
 .extradata.sec.custom=The entry was posted with custom access.
 
-.extradata.subject=The entry was posted with the following subject: [[subject]]
+.extradata.subj=The entry was posted with the following subject: 
+
+.extradata.subject.no_subject=(no subject)
+
+.extradata.subject.subject=[[subject]]
```
kareila: I see how you ended up here by refactoring the existing string, but: there is no reason to use `.extradata.subject.subject` as a translation string if its only purpose is to reprint the value of the variable. You should just print the value of the variable instead of making a translation call.

views/entry/success.tt.text
```diff
@@ -48,4 +48,8 @@
 
 .extradata.sec.custom=The entry was posted with custom access.
 
-.extradata.subject=The entry was posted with the following subject: [[subject]]
\ No newline at end of file
+.extradata.subj=The entry was posted with the following subject: 
+
+.extradata.subject.subject=[[subject]]
```
kareila: Same as above.

cgi-bin/DW/Controller/Entry.pm
```diff
@@ -1038,7 +1038,7 @@ sub _do_post {
         my $extradata = {
             security => $form_req->{security},
             security_ml => "",
-            subject => LJ::ehtml( $form_req->{subject} ),
+            subject => $form_req->{subject},
```
kareila: This is a bit misleading: the problem in #1453 is that the text there was passed as an argument to `form.textbox` and `form.hidden`, both of which escape their arguments when printing their form fields.  You're not using those below, so either this should be preserved, or (better) you should move the escaping into the TT view.  I'll point out what I think that should look like below.

views/entry/success.tt
```diff
@@ -24,8 +24,14 @@ the same terms as Perl itself.  For a copy of the license, please reference
 [%- END -%]
 
 [%- IF poststatus.ml_string != ".edit.delete" -%]
+
 <p>[% extradata.security_ml | ml %]</p>
-<p>[% ".extradata.subject" | ml( subject => extradata.subject) %]</p>
+<p>[% ".extradata.subj" | ml %]
+    [% IF extradata.subject.length > 0 %]
+        [% ".extradata.subject.subject" | ml( subject => extradata.subject ) %]
```
kareila: Here is what I recommend instead for your line 31:
```
[% extradata.subject | html %]
```
The `| html` does the same thing LJ::ehtml would do in the controller, and as I said elsewhere, there's no need to pass this through the translation system at all.

rahaeli: Shouldn't we run things through the translation system so they can be changed on-the-fly instead of with a patch & a push? Or am I misreading things?

kareila: What was happening before was the subject was being passed to a single translation string as input.  But now we have split up the header ("The entry was posted with the following subject:") and the subject into separate strings, and the subject is either the entry subject text, or the text "(no subject)" - in the first case, passing it through the translation system gains us nothing.

If we want to go back to having everything combined into one translation string, though, so that we can reorder things on the fly, we'll need a slightly different approach from the one used here.  Something like:
```
[% IF extradata.subject.length > 0 %]
[% subject = extradata.subject | html %]
[% ELSE %]
[% subject = ".extradata.subject.no_subject" | ml %]
[% END %]
<p>[% ".extradata.subject" | ml( subject => subject) %]</p>
```

rahaeli: Aha! Got it. No, we don't need to have it combined into one string, I just want to make sure that I can change text on the fly without having to patch & push. :)

kaberett:
... dammit I have screwed up git. I think I've made the requested changes, but they've ended up over at #1641 for reasons I am not entirely clear on. Sorry!

kareila:
No worries, I'll just close this one then.

PR #1618, opened 2015-11-08
adding .form.subject

azurelunatic:
getting rid of the extra colon on the mobile post page, now with more translation.


kareila:
The only change I see in this commit is your changes to deadphrases.dat.  Can you also commit the other related changes to this branch?

azurelunatic:
I can try. (I'm using the github web interface, which is kind of not the expected way.)

rahaeli:
I will confess that I have occasionally deleted the branch and started over. BECAUSE GIT.

kareila:
Okay, I committed the related changes from the other branch. I don't think adding this to deadphrases.dat is necessary.  Presumably the existing mobile page will DIAF at some point.

Issue #1577, opened 2015-09-24
QuickerReply + access-locked post + OpenID = inaccurate error message


kaberett:
From a support request left by an OpenID user (http://www.dreamwidth.org/support/see_request?id=31619): 

> I was on my reading page and clicked on Reply on a friend's locked entry. The quicker reply box opened up (surprising me because I typically open entries in new tabs before hitting reply). After typing in my comment I clicked on Post Comment and the following error popped up:
>
> Comment error: OpenID users aren't authorized to reply to this entry.
>
> I then opened up the entry in a new browser tab, pasted my comment in the reply box and was able to post my comment successfully.

chrisboyle:
claim

A similar error happens when QuickReplying on the individual entry page. You get bounced to `www/talkpost_do` with a heading "Comment Not Posted" and a bullet point immediately below of the same "OpenID users aren't authorized to reply to this entry."

chrisboyle:
Actually QuickReply on the entry page worked once I validated my email. There's still a bug there; the message should have said go validate your email. So to summarise, there are bugs in all three places

* Reading page: incorrect message, and should have allowed commenting once I validated my email.
* Entry page: incorrect message only (behaviour is correct).
* Reply page: incorrectly **allowed** a comment without validated email.

This is all with the default `opt_whocanreply=reg` ("registered accounts") - I assume the intent there is that OpenID users must validate their email to be allowed under that setting.

chrisboyle:
* Reading page is AJAX-y via `/username/__rpc_addcomment` -> `RPC::Misc` -> `LJ::Protocol` -> `LJ::Comment->create` -> `LJ::Talk::Post::init` ("handles all the error checking") and `LJ::Talk::Post::post_comment` -> `enter_comment` which INSERTs.
* Entry page: Full-page post to `talkpost_do` -> `LJ::Talk::Post::post_comment`
* Reply page: as entry page.

chrisboyle:
Having checked with Kareila about whether access-grant without verification is supposed to be sufficient, it is. This gives:

Page | -V-A | -V+A | +V-A | +V+A | Change?
--- | --- | --- | --- | --- | ---
Reading page (AJAX) | link | link | fail & wrong msg | fail & wrong msg | **Make both +V cases work.** Could also enable QR for -V+A?
Entry page (talkpost_do w/o explicit ID mode) | wrong msg | fail & wrong msg | works | works | **Fix message.**
[...]

PR #1619, opened 2015-11-08
Fix bogus rejections of OpenID comments.

chrisboyle:
This makes a difference from the reading page (with a theme with
non-broken quick reply support) and from the entry page without a
verified email; see issue for details.

Fixes: #1577 - see copious details and tables of cases on that issue.


kareila:
Thanks again for your work on this!  Much appreciated.

PR #1629, opened 2015-11-09
[#1628] check for userids, not IP addresses

### LINKED ISSUE #1628: Last N Hours reports do not do the right thing with IPv6 addresses
http://www.dreamwidth.org/admin/spamreports?mode=last06hr_a will illustrate this for a little while, and http://www.dreamwidth.org/admin/spamreports?mode=last24hr_a for a while longer. 

These are accessed from http://www.dreamwidth.org/admin/spamreports and you will need the antispam privs to see these pages. http://wiki.dreamwidth.net/wiki/index.php/Developing_Antispam

a) it has an IPv6 address and displays the correct count for that address
b) it displays an OpenID user in the listing and the link reflects a user not an IP
c) when you get to the report, it's not the IPv6 results, it's by=posterid rather than by=ip and there's nothing there because the OpenID user hasn't had any spam reports generated against it http://www.dreamwidth.org/admin/spamreports?mode=view&by=posterid&what=2601:282:301:17&state=open / http://www.dreamwidth.org/admin/spamreports?mode=view&by=posterid&what=2601:282:301:17&state=closed
d) when you change the by= to by=ip, it displays the IPv6's reports http://www.dreamwidth.org/admin/spamreports?mode=view&by=ip&what=2601:282:301:17&state=open / http://www.dreamwidth.org/admin/spamreports?mode=view&by=ip&what=2601:282:301:17&state=closed

My guess is that the lastn hours page sees something that "doesn't look" like an IP in there, and then treats it as a user even though it should not do so; if you also want to poke at the individual spam report page to make sure that it's not possible to have a mistaken identity situation, that would also be good. 

Spamwhackers have been instructed that if they see an OpenID in the "last N hours (anonymous)" spam roundup, to not do anything to that OpenID (if they have the powers to) as it has done nothing wrong and is the victim of a mistaken identity situation.

kareila:
This changes the conditional for deciding whether an antispam report
is for a user or an IP address.  The old regexp only looked for
the existence of a period, which worked for IPv4 but not IPv6.

Since userids are only digits, use a regexp that checks for those
instead and assume anything else should be treated as an IP address.

Fixes #1628.

### DIFF
--- a/cgi-bin/DW/Controller/Admin/SpamReports.pm
+++ b/cgi-bin/DW/Controller/Admin/SpamReports.pm
@@ -186,14 +186,14 @@ sub main_controller {
         foreach ( sort { $b <=> $a } keys %revhits ) {
             my $r = $revhits{$_};
             foreach ( @$r ) {
-                if ( /\./ ) {  # ip report
-                    push @rows, [ $hits{$_}, $_,
-                                  $viewlink->( 'ip', $_, 'open', $times{$_} )
-                                ];
-                } else {  # user report
+                if ( /^\d+$/ ) {  # userid
                     my $u = LJ::load_userid( $_ );
                     push @rows, [ $hits{$_}, LJ::ljuser($u),
                                   $viewlink->( 'posterid', $_, 'open', $times{$_} )
+                                ];
+                } else {  # assumed to be IP address
+                    push @rows, [ $hits{$_}, $_,
+                                  $viewlink->( 'ip', $_, 'open', $times{$_} )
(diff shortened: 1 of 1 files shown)


zorkian:
That's pretty funny. Thank you for fixing. :)

kareila:
It was calling load_userid on an IPv6 address *and succeeding*. I'm pretty horrified that is even possible.

rahaeli:
I think it's funny AND horrifying!

PR #1633, opened 2015-11-21
[#1368] more detailed explanation of options when renaming an account

### LINKED ISSUE #1368: add clearer warning when choosing rename redirecting
See #1367 for the admin end of this, but: the process for renaming should make people more aware that redirecting a username will mean that you can't then rename another account to the old username.

(Ideally via usernames in the message, since renames are confusing and having it written out will help.)

Something like: "You will be renaming "oldusername" to "newusername". You have chosen to redirect "oldusername" to "newusername" so that anyone with a link to your old username will be forwarded to your new username. If you do this, "oldusername" will not become available for you or anyone else to register or rename to in the future, and if you want to rename another account to "oldusername" you won't be able to."

(Phrasing crappy, I'm on like 27h uptime, feel free to tweak. Heh.)

kareila:
Thought about adding an interstitial page or pop-up, but this was easiest.

Fixes #1368.

### DIFF
--- a/cgi-bin/DW/Controller/Rename.pm
+++ b/cgi-bin/DW/Controller/Rename.pm
@@ -113,6 +113,7 @@ sub rename_handler {
             $vars->{formdata} = {
                 authas      => $authas,
                 journaltype => $get_args->{type},
+                journalname => $get_args->{type} eq "P" ? $remote->user : "communityname",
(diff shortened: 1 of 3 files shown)


rahaeli:
Thank you for making my crappy copy so much better!

PR #1635, opened 2015-11-23
[#1330] add admin form for viewing other users' notification settings

### LINKED ISSUE #1330: admin tool to view user's notification subscriptions
Sometimes it would help for troubleshooting to see a user's notifications, the way they'd appear on http://www.dreamwidth.org/manage/settings/?cat=notifications -- it's possible to do so, but only by impersonating them and loading that page, and I feel dirty every time I impersonate someone and would like to minimize the usage as much as possible! Minimum necessary data access, etc.

So, I would like an /admin/subscriptions -- hinged off (a new sub-priv) canview:subscriptions, returns the user's subscriptions in some fashion, including both entries/threads they've subscribed to (the bottom half) and their settings for the checkboxes on the top half. It doesn't have to be an exact duplicate of http://www.dreamwidth.org/manage/settings/?cat=notifications the way the user would see it, but it should contain all the info that page contains.

kareila:
Replaces the top row of action buttons with a simple input form if
the logged-in user has the canview:subscriptions priv.

Fixes #1330.

### DIFF
--- a/htdocs/manage/settings/index.bml
+++ b/htdocs/manage/settings/index.bml
@@ -32,8 +32,16 @@ body<=
     my $authas;
     my $u;
 
-    if ($remote) {
-        $authas = $GET{authas} || $remote->user;
+    $authas = $GET{authas} || $remote->user if $remote;
+
+    my $can_view_other = ( $remote && $remote->has_priv( "canview", "subscriptions" ) );
+
+    if ( $can_view_other && $GET{cat} && $GET{cat} eq 'notifications' ) {
+        # impersonation mode - check for $u->user ne $authas
+        $u = LJ::load_user( $GET{user} ) if $GET{user};
+    }
+
+    if ( $remote && ! defined $u ) {
         $u = LJ::get_authas_user($authas);
         return LJ::bad_input($ML{'error.invalidauth'})
             unless $u;
@@ -203,6 +211,15 @@ body<=
         $given_cat = "display";
     }
 
+    if ( $u->user ne $authas ) {
+        # can only impersonate notification settings -
+        # if invisible or disabled, print this error
+        return LJ::bad_input( $ML{'error.invalidauth'} )
+            if $given_cat ne 'notifications';
+        # don't allow editing
+        $cats_with_settings{notifications}->{form} = 0;
+    }
+
     my @settings = @{$cats_with_settings{$given_cat}->{settings}};
 
     # remove any settings that don't exist for this category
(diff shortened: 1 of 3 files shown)


htdocs/manage/settings/index.bml
```diff
@@ -32,8 +32,16 @@ body<=
     my $authas;
     my $u;
 
-    if ($remote) {
-        $authas = $GET{authas} || $remote->user;
+    $authas = $GET{authas} || $remote->user if $remote;
+
+    my $can_view_other = ( $remote && $remote->has_priv( "canview", "subscriptions" ) );
```
zorkian: Useless use of parenthesis sighted!

kareila: I went back and forth on whether to use the `? 1 : 0` construction at the end of that, and ultimately left it off.

htdocs/manage/settings/index.bml
```diff
@@ -32,8 +32,16 @@ body<=
...
+
+    if ( $can_view_other && $GET{cat} && $GET{cat} eq 'notifications' ) {
+        # impersonation mode - check for $u->user ne $authas
+        $u = LJ::load_user( $GET{user} ) if $GET{user};
```
zorkian: Shouldn't this be promoted into the if statement?

htdocs/manage/settings/index.bml
```diff
@@ -445,6 +466,23 @@ body<=
     $ret .= $cats_with_settings{$given_cat}->{desc};
     $ret .= "</p></div>";
 
+    $ret .= "<div class='settings_content'>";
+    $ret .= "<div class='$given_cat'>";
+
+    if ( $given_cat eq "notifications" && $can_view_other ) {
+        # form for viewing another user's notifications
```
zorkian: I know this is necessary here, but every time I see BML I always think "wow, I can't believe this used to be how we made a web site". And still is in this case, but modern Foundation/TT is so much nicer to look at...

And then I wonder what I will think in another 10 years.

htdocs/manage/settings/index.bml
kareila: It could be - it's syntactically the same either way, and I thought it was visually clearer to have the two $GET{user} references adjacent to each other.

zorkian:
Just minor comments on style and then it can merge.

htdocs/manage/settings/index.bml
kareila: In another ten years we'll probably be building sites to be projected directly into people's eyeballs. :)

I decided quite firmly that converting this page to TT was beyond the scope of what I wanted to do for this bug.

htdocs/manage/settings/index.bml
zorkian: For code comprehension ability, I think it's nicer to see "if they want to view another user, and they can view another user... then load another user" rather than "if they can view another user, load another user... if they specified one".

htdocs/manage/settings/index.bml
zorkian: :+1: No sense delaying an immediate need for the "grand amazingness of the future".

kareila:
Updated as requested.

zorkian:
Thank you!

rahaeli:
You know that in ten years we'll still be finding bits of BML hidden away in the code somewhere.

PR #1641, opened 2015-12-02
More detailed subject on Success landing page

kaberett:
Adds "(no subject)" to extradata line for subject if entry
has no subject, instead of leaving blank.


kareila:
Looks great! Thanks for all your hard work.

PR #1647, opened 2015-12-07
[#1645] parse embedded content before displaying edit entry form (beta)

### LINKED ISSUE #1645: inconsistent display of HTML code for embedded content when editing entries
From http://www.dreamwidth.org/support/see_request?id=32273 (1/2)

When using the legacy (HTML) editor, the edited code looks like (for example):
```
<site-embed id="1"><iframe src="https://books.google.com/ngrams/interactive_chart?content=bussard+ramjet&amp;case_insensitive=on&amp;year_start=1970&amp;year_end=2008&amp;corpus=15&amp;smoothing=3&amp;share=&amp;direct_url=t1%3B%2CBussard%20ramjet%3B%2Cc0" width="900" height="500" marginwidth="0" marginheight="0" hspace="0" vspace="0" frameborder="0" scrolling="no"></iframe></site-embed>
```

When editing the same entry with the beta update page, all you see is:
```
<site-embed id="1"/>
```

In other words, people using the beta update page can't access the HTML code of embedded content to edit it.

kareila:
By analogy with editjournal.bml - does the right thing in my testing.

Fixes #1645.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -756,7 +756,7 @@ sub _form_to_backend {
 
     # Check if this is a community.
     $props->{admin_post} = $post->{flags_adminpost} || 0;
-        
+
     # entry security
     my $sec = "public";
     my $amask = 0;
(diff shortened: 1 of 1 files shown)


zorkian:
"By analogy" does this mean you are making this do the same thing that editjournal does?

kareila:
That was my intention yes :)

zorkian:
Great -- if that's the assertion, I'll merge it. I don't recall what the behavior is so normally I'd be pretty worried about this kind of change having unintended consequences, but changing to match the existing system is usually reasonable.

Issue #1646, opened 2015-12-07
add Google Calendar to embed whitelist


kareila:
From http://www.dreamwidth.org/support/see_request?id=32273 (2/2)

Old Google Calendar embeds that use www.google.com still work, but new ones use calendar.google.com and need whitelisting.

Example content:
```
<iframe src="https://calendar.google.com/calendar/embed?src=en.usa%23holiday%40group.v.calendar.google.com&ctz=America/Chicago" style="border: 0" width="800" height="600" frameborder="0" scrolling="no"></iframe>
```

srukle:
I can claim this.

srukle:
pull request made: #1648

PR #1648, opened 2015-12-08
(#1646) add g calendar

srukle:
tested on hackspace:
1) removed www.google.com
2) added calendar.google.com
3) no conflicts found with embedded content books.google.com
4) no error messages reported on start up


kareila:
Please don't delete the www.google.com/calendar whitelist. That will break the older embeds.

srukle:
Updated to include www.google.com
tested on hackspace, did not find any conflicts

PR #1655, opened 2015-12-16
[#1653] comment out use of missing DW::Stats::timing subroutine

### LINKED ISSUE #1653: problematic DW::Stats code in DW::Worker::XPostWorker
The DW::Stats::timing subroutine is undefined.  I saw warnings when I was testing crossposts in my dev environment.

It looks like the relevant changeset is 2579e39489ba8a2fd6eef26c9c9c163e7648d660.

Should we simply remove the statements, or figure out what was intended to happen there?

kareila:
I propose that we comment out the affected lines until we
decide whether/how to implement the subroutine.

Fixes #1653.


zorkian:
Yeah I think at one point I had an implementation of this method and lost it when I was doing the dogstatsd integration

PR #1666, opened 2016-01-07
(Issue 1046) Convert Support Index to TT and Foundation

hotlevel4:
As stated above.

A few questions/concerns:
(1) Was unable to test Known Issues as I'm not sure where the edit functionality for that is.
(2) The URLs in the ask and search links in the Ask a Question box no longer contain site.root as part of the URL.
(3) I did not put in the CSS for green/blue boxes. Should that be done? If so, should it be done through SCSS or in the page itself?


rahaeli:
The Known Issues box is pulled in through the on-site file insertion. You want to give yourself the priv "fileedit:*" and then visit /admin/fileedit and edit support-currentproblems.

The green/blue boxes are holdovers from LJ, I believe -- the colors don't mesh with our site skins -- but do serve nicely to visually separate the page. Let's add some CSS classes to the page named something *other* than 'blue box' and 'green box' (support-highight-darker and support-highlight-lighter?) and then pick some colors and put them in each site skin? I'll play with some colors for each site skin if you don't feel comfortable doing it yourself!

cgi-bin/DW/Controller/Support/Index.pm
```diff
@@ -0,0 +1,77 @@
...
+package DW::Controller::Support::Index;
+
+use strict;
+use warnings;
+
+use DW::Controller;
+use DW::Routing;
+use DW::Template;
+
+DW::Routing->register_string( '/support/', \&index_handler, app => 1 );
```
kareila: This line should be removed.  Registering the handler for `/support/index` automatically registers `/support/` as well.

cgi-bin/DW/Controller/Support/Index.pm
```diff
@@ -0,0 +1,77 @@
...
+DW::Routing->register_string( '/support/index', \&index_handler, app => 1 );
+
+sub index_handler {
+    my $r = DW::Request->get;
+
+    my ( $ok, $rv ) = controller( anonymous => 1, form_auth => 1 );
```
kareila: I don't see anywhere on this page where we are processing user input, so the `form_auth` argument should be removed.

cgi-bin/DW/Controller/Support/Index.pm
```diff
@@ -0,0 +1,77 @@
...
+    my $user_url;
+
+    my $vars = {};
+    
+    my $currentproblems = LJ::load_include( "support-currentproblems" );
+    LJ::CleanHTML::clean_event( \$currentproblems, {} );    
+    $vars->{currentproblems} = $currentproblems;
+    
+    my $dbr = LJ::get_db_reader(); 
+    my $sth;
```
kareila: I would move these two lines down to where `$sth` is assigned below; they are out of context here. I would also do `my $sth = $dbr->prepare(...);` -- I see no reason to separate the scope declaration from the assignment in this case.

cgi-bin/DW/Controller/Support/Index.pm
```diff
@@ -0,0 +1,77 @@
...
+        my $u = LJ::load_user( $LJ::EXAMPLE_USER_ACCOUNT );
+        $user = $u ? $u->user : "<b>[Unknown or undefined example username]</b>";
+        $user_url = $u ? $u->journal_base : "<b>[Unknown or undefined example username]</b>";
+    }    
+    
+    $sth = $dbr->prepare( "SELECT statkey FROM stats WHERE statcat='pop_faq' ORDER BY statval DESC LIMIT 10" );
+    $sth->execute;    
+
+    while ( my $f = $sth->fetchrow_hashref ) {
+        $f = LJ::Faq->load( $f->{statkey}, lang => BML::get_language() );
```
kareila: Part of the conversion process is replacing any function calls within the BML package, so this call to `BML::get_language()` needs to be replaced.  It looks to me like the non-BML equivalent here would be `LJ::Lang::get_effective_lang()`.

kareila:
I noted a few minor issues in the controller module, but overall this looks good.  I think it's OK to leave `site.root` out of the links you mentioned, since they link to other pages in the support hierarchy.

I would like @rahaeli's opinion on whether we should keep the default styling on the boxes or if we want to return them to green and blue.  I think the term "Big Blue Box" is so engrained in our jargon that having it be non-blue would be confusing to our support volunteers.  I would also like a second opinion on the one-column layout of the conversion versus recreating the existing two-column layout.

If we do decide to go back to those colors, I think just defining them on that one page would be OK.

rahaeli:
* We should keep the two-column, please!

* Yes, please keep the existing colors, trying to find new ones that work with each of the site skins would be a pain.

hotlevel4:
Fixed issues above and hard coded CSS colors for boxes.

kareila:
I really like how this is looking.  Two requests:

1. The breakpoint between one-column and two-columns is still wider than I'm comfortable with - it takes effect on my laptop screen with anything much less than full screen-width of the browser window.  I'd suggest reducing the amount of horizontal padding between the two columns (the vertical spacing between boxes is fine) and possibly reducing the size of the body text slightly.

2. The box colors look good, but instead of using inline style attributes, it would be better to use CSS classes that can be easily altered.  You can use inline CSS in the header of the template file - see https://github.com/dreamwidth/dw-free/blob/develop/views/search.tt#L16 for an example.

Thanks again for working on this!

hotlevel4:
Just wanted to follow up on this? Thanks!

kareila:
Sorry, I missed that you had updated this!  Thanks for following up.

kareila:
Two remaining issues and then I think this can go live on the site.  Thanks for your patience!

Issue 1 (minor): I think the h4 tags should be h3 tags, so that the header text stands out from the body text.

Issue 2 (major): The page is currently unreadable in the Gradation site schemes.  I'm so sorry I didn't think to test the other site schemes sooner!

But there is good news!  The existing green and blue hues for the various site schemes are already defined for non-Foundation pages.  We just need to define them for use in Foundation pages, and then we can apply them here.  Once that's done, you can remove the inline CSS on the page - I'm sorry I misled you earlier!

Since you've already spent so much time on this, I'm going to take care of the necessary CSS definitions in a separate pull request, and then you can use the new classes when they're available.

kareila:
Once the above PRs are merged, your existing "greenbox" and "bluebox" classes will have the correct coloring in all site schemes.  All that remains is to remove the page's definition for both classes and to update the h4 tags to h3 tags.  Thanks again!

kareila:
CSS PRs are merged now!

hotlevel4:
Updated h4 to h3 and removed inline CSS. Sorry for the delay!

zorkian:
@kareila is this on your plate or should I look?

kareila:
I've got it, was just waiting for the support queue to calm down 😑

kareila:
I think this looks good! Thanks for seeing it through.  ☺️

PR #1668, opened 2016-01-14
Add transfer subcommand to bonus_icons

### LINKED ISSUE #1546: bonus_icons console command could use an 'xfer' param
bonus_icons right now just has:

bonus_icons username: lists # of bonus icons
bonus_icons username add #: adds # of bonus icons
bonus_icons username remove #: removes # of bonus icons

I would also like "bonus_icons username xfer otherusername", which should:

a) Check to make sure that username and otherusername have the same confirmed email address (and fail if they don't with a "Email addresses do not match" error)

b) Remove all the bonus icons from username

c) Add all the bonus icons to otherusername

It will save me time, typing, and trouble. :)

kaberett:
Fixes #1546.

New syntax "bonus_icons \<user1\> xfer \<user2\>" transfers all of
user1's bonus icon slots to user2.

(In light of http://www.dreamwidth.org/support/see_request?id=32564 -- @rahaeli, it'd be super-easy to tweak this so the syntax was instead "bonus_icons \<user1\> xfer \<"all"|integer\> \<user2\>" -- please say if you'd prefer that?)

### DIFF
--- a/cgi-bin/DW/Console/Command/BonusIcons.pm
+++ b/cgi-bin/DW/Console/Command/BonusIcons.pm
@@ -6,8 +6,9 @@
 #
 # Authors:
 #      Mark Smith <mark@dreamwidth.org>
+#      Alex Brett <kaberett@dreamwidth.org>
 #
-# Copyright (c) 2012 by Dreamwidth Studios, LLC.
+# Copyright (c) 2012-2016 by Dreamwidth Studios, LLC.
 #
 # This program is free software; you may redistribute it and/or modify it under
 # the same terms as Perl itself.  For a copy of the license, please reference
@@ -25,19 +26,21 @@ sub cmd { 'bonus_icons' }
 sub desc { 'Manage bonus icons for an account. Requires priv: payments:bonus_icons.' }
 sub args_desc {
     [
-        'command' => 'Subcommand: add, remove.',
+        'command' => 'Subcommand: add, remove, xfer.',
         'username' => 'Username to act on.',
-        'count' => 'How many icons to add or remove.',
+        'commandargs' => "'add' and 'remove' take one argument: count (the number
+                          of icons to add or remove). 'xfer' takes one argument:
+                          the target username (for icons to be transferred to)."
     ]
 }
-sub usage { '<username> [<subcommand> <count>]' }
+sub usage { '<username> [<subcommand> <commandargs>]' }
 sub can_execute {
     my $remote = LJ::get_remote();
     return $remote && $remote->has_priv( 'payments' => 'bonus_icons' );
 }
 
 sub execute {
-    my ( $self, $user, $cmd, $count ) = @_;
+    my ( $self, $user, $cmd, $cmdarg ) = @_;
 
     my $remote = LJ::get_remote();
     return $self->error( 'You must be logged in!' )
@@ -56,25 +59,54 @@ sub execute {
     }
 
     return $self->error( 'Invalid subcommand.' )
-        if $cmd && $cmd !~ /^(?:add|remove)$/;
+        if $cmd && $cmd !~ /^(?:add|remove|xfer)$/;
 
-    return $self->error( 'Count must be a positive integer.' )
-        unless $count =~ /^\d+$/;
-    $count += 0;
+    if ( $cmd eq 'add' || $cmd eq 'remove' ) {
 
-    if ( $cmd eq 'add' ) {
-        my $new = max( $to_u->prop( 'bonus_icons' ) + $count, 0 );
-        $to_u->set_prop( bonus_icons => $new );
+        my $count = $cmdarg;
+
+        return $self->error( 'Count must be a positive integer.' )
+            unless $count =~ /^\d+$/;
+        $count += 0;
+
+        if ( $cmd eq 'add' ) {
+            my $new = max( $to_u->prop( 'bonus_icons' ) + $count, 0 );
+            $to_u->set_prop( bonus_icons => $new );
+            LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
+                    sprintf( 'Added %d icons, new total: %d.', $count, $new ) );
+            $self->print( sprintf( 'User now has %d icons.', $new ) );
+
+        } elsif ( $cmd eq 'remove' ) {
+            my $new = max( $to_u->prop( 'bonus_icons' ) - $count, 0 );
+            $to_u->set_prop( bonus_icons => $new );
+            LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
+                    sprintf( 'Removed %d icons, new total: %d.', $count, $new ) );
+            $self->print( sprintf( 'User now has %d icons.', $new ) );
+
+        }
+
+    } elsif ( $cmd eq 'xfer' ) {
+        my $destination_u = LJ::load_user( $cmdarg );
+
+        return $self->error( 'Invalid target user.' )
+            unless $destination_u;
+        return $self->error( 'E-mail addresses do not match.' )
+            unless $to_u->email_raw eq $destination_u->email_raw; 
+        return $self->error( 'One or more email address(es) not confirmed.' )
+            unless $to_u->is_validated && $destination_u->is_validated;
+
+        my $xfer_count = $to_u->prop( 'bonus_icons' );
+        $to_u->set_prop( bonus_icons => 0 );
         LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
-                sprintf( 'Added %d icons, new total: %d.', $count, $new ) );
-        $self->print( sprintf( 'User now has %d icons.', $new ) );
-
-    } elsif ( $cmd eq 'remove' ) {
-        my $new = max( $to_u->prop( 'bonus_icons' ) - $count, 0 );
-        $to_u->set_prop( bonus_icons => $new );
-        LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
-                sprintf( 'Removed %d icons, new total: %d.', $count, $new ) );
-        $self->print( sprintf( 'User now has %d icons.', $new ) );
+                sprintf( 'Transferred %d icons to %s.', $xfer_count,
+                    $destination_u->user ) );
+        my $new_total = $destination_u->prop( 'bonus_icons' ) + $xfer_count;
+        $destination_u->set_prop( bonus_icons => $new_total );
+        LJ::statushistory_add( $destination_u, $remote, 'bonus_icons',
+                sprintf( 'Received %d icons from %s, new total: %d.',
+                    $xfer_count, $to_u->user, $new_total ) );
+        $self->print( sprintf( '%s now has %d icons.', $destination_u->user,
+            $new_total ) );
 
     }
 


rahaeli:
Nah, requests to only transfer a few icons or transfer to multiple accounts are rare enough that I can just do those manually!

kaberett:
Fab :-)

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -56,25 +58,54 @@ sub execute {
...
-    if ( $cmd eq 'add' ) {
-        my $new = max( $to_u->prop( 'bonus_icons' ) + $count, 0 );
-        $to_u->set_prop( bonus_icons => $new );
-        LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
-                sprintf( 'Added %d icons, new total: %d.', $count, $new ) );
-        $self->print( sprintf( 'User now has %d icons.', $new ) );
+        my $count = $cmdarg;
+
+        return $self->error( 'Count must be a positive integer.' )
+            unless $count =~ /^\d+$/ || $cmd eq 'xfer';
```
kareila: You shouldn't need the `$cmd eq 'xfer'` check here because you are in a block that only executes if $cmd is add or remove.  It's harmless but potentially confusing.

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -56,25 +58,54 @@ sub execute {
...
+
+        }
+
+    } elsif ( $cmd eq 'xfer' ) {
+        my $target_u = LJ::load_user( $cmdarg );
+
+        return $self->error( 'Invalid target user.' )
+            unless $target_u;
+        return $self->error( 'E-mail addresses do not match.' )
+            unless $to_u->prop( 'email' ) eq $target_u->prop( 'email' );
```
kareila: We check that the email addresses match here, but does this make sure the addresses were confirmed on both accounts?

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -56,25 +58,54 @@ sub execute {
...
+        return $self->error( 'Invalid target user.' )
+            unless $target_u;
+        return $self->error( 'E-mail addresses do not match.' )
+            unless $to_u->prop( 'email' ) eq $target_u->prop( 'email' ); 
 
-    } elsif ( $cmd eq 'remove' ) {
-        my $new = max( $to_u->prop( 'bonus_icons' ) - $count, 0 );
-        $to_u->set_prop( bonus_icons => $new );
+        my $xfer_count = $to_u->prop( 'bonus_icons' );
+        my $new_to_u = max( $to_u->prop( 'bonus_icons' ) - $xfer_count, 0 );
```
kareila: Shouldn't one of these $to_u variables be $target_u?  Otherwise you're subtracting the thing from itself.

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -56,25 +58,54 @@ sub execute {
...
+        $to_u->set_prop( bonus_icons => $new_to_u );
         LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
-                sprintf( 'Removed %d icons, new total: %d.', $count, $new ) );
-        $self->print( sprintf( 'User now has %d icons.', $new ) );
+                sprintf( 'Transferred %d icons to %s.', $xfer_count,
+                    $target_u->user ) );
+        my $new_target_u = max( $target_u->prop( 'bonus_icons' ) 
+            + $xfer_count, 0 );
```
kareila: You shouldn't need to use max here, since the result of the addition will always be a positive number.

kareila:
Thanks for working on this. I don't see any major issues apart from making sure the email addresses on both accounts are confirmed, but I'd appreciate some clarification/simplification when doing the math on the number of icons - I think it's more complicated than it needs to be.  Let me know if you have any questions!

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: So I drew myself a pretty detailed sketch while writing this, and having tested I am pretty sure it's working as intended -- $to_u is the source account for bonus icon slots and $target_u is the sink (or receiver, or whatever). I've maintained the terminology of "$to_u" by analogy with the use of $to_u in "remove". I agree it's confusing (not... helped by the fact that GHI seems to be skipping 20 lines out of the diff, so giving completely the wrong impression of the context AFAICT) but I'm not sure how to make it clearer?

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: ... good point; done. (Comment for my benefit for easy tracking of what-still-needs-doing before the next pull req!)

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: okay!

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: So in fact this is also the point where I have a question about how to simplify the maths -- I think I had a chat about how Perl handles objects and things that point at them with geekosaur in IRC when I was working on this, but what I'm tripping up on here is:

$xfer_count is the full number of bonus icons that $to_u has. $to_u->prop( 'bonus_icons' ) needs setting to 0 (which... i could actually just do rather than faffing with the subtraction, okay, well done me) without making $xfer_count also 0. If I set $xfer_count, then reset $to_u->prop( 'bonus_icons' ) to 0, in Perl does $xfer_count stay the number I first thought of (i.e. the number I want to add to $target_u)? (I tried reading docs and got myself confused on this point. :-/ Happy to be pointed at what I should be reading, though!)

alierak: Equivalent code: my $new_to_u = 0; which is probably correct since you're transferring away all their bonus icons.

kaberett: Yep. Do Perl objects work in such a way that I do need the $new_to_u level of abstraction, is the point I'm currently a bit stuck on, I think? :-)

alierak: No - once you have the result of a calculation it's just a number, and it won't change due a change in the objects you looked at when setting up the calculation.

cgi-bin/DW/Console/Command/BonusIcons.pm
rahaeli: Huh. I went looking to see if we had helper functions for checking that two accounts have the same email address and both accounts have their address confirmed, and we don't that I can find. We not only don't have that, we don't have ANY functions for email checks. It's done directly each time. Whoa.

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: Fab, thanks!

rahaeli: $to_u is convention across console commands for "account that is the target of a console command", but for the recipient, how about $destination_u?

kaberett: Editing. :-)

cgi-bin/DW/Console/Command/BonusIcons.pm
kaberett: Done. I think.

@rahaeli -- I am probably not the best person to file this bug but can do so if no-one else gets to it!

rahaeli: yeah I was going to file the bug, I just got distracted!

kaberett:
Changes made, tested, and pushed!

cgi-bin/DW/Console/Command/BonusIcons.pm
rahaeli: oh wait a second: we do have is_validated, which is part of what I was thinking. But yeah, too much in-code comparison of $u->email and $u->email_raw being done...

kaberett: yep, just used it! :-)

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -85,27 +85,26 @@ sub execute {
...
+        my $destination_u = LJ::load_user( $cmdarg );
 
         return $self->error( 'Invalid target user.' )
-            unless $target_u;
+            unless $destination_u;
         return $self->error( 'E-mail addresses do not match.' )
-            unless $to_u->prop( 'email' ) eq $target_u->prop( 'email' ); 
+            unless $to_u->prop( 'email' ) eq $destination_u->prop( 'email' ); 
+        return $self->error( 'E-mail address is not confirmed.' )
+            unless $to_u->is_validated && $destination_u->is_validated;
```
kaberett: @rahaeli, are you happy with this, or would you like the error message to say which account(s) do(es)n't have the e-mail address confirmed?

rahaeli: Nah, go for easy and just make it "one or more email addresses is not confirmed" (psst: email, not e-mail, by site style handbook) -- any error in this process means I'm going to have to dig into it more, so it's ok if it's just a "something is broken" abort.

Jen can confirm, but: in looking through the code for opening up that other bug (and thus refamiliarizing myself with things), I think the 'unless' comparisons should maybe be done as "$to_u->email_raw" instead of "$to_u->prop( 'email' )".

kareila: Correct, email_raw is what is used elsewhere.

kaberett:
... accidentally deleted some line comments, sorry. email_raw now used!

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -85,27 +85,26 @@ sub execute {
...
+        $to_u->set_prop( bonus_icons => 0 );
         LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
                 sprintf( 'Transferred %d icons to %s.', $xfer_count,
-                    $target_u->user ) );
-        my $new_target_u = max( $target_u->prop( 'bonus_icons' ) 
-            + $xfer_count, 0 );
-        $target_u->set_prop( bonus_icons => $new_target_u );
-        LJ::statushistory_add( $target_u, $remote, 'bonus_icons',
+                    $destination_u->user ) );
+        $destination_u->set_prop( bonus_icons => $destination_u->prop( 'bonus_icons' ) + $xfer_count );
```
kareila: Because you're referencing this number so often, and because the `prop` method has a certain amount of computational overhead, I would still save this to a variable here:
```
my $new_total = $destination_u->prop( 'bonus_icons' ) + $xfer_count;
$destination_u->set_prop( bonus_icons => $new_total );
```
And then in the two places below where you call `$destination_u->prop( 'bonus_icons' )` you can change those to `$new_total` as well.  Or whatever you want to call it; it was $new_target_u before but you renamed $target_u, so... :grin: 

The rest is much easier to read now, thanks! :+1:

rahaeli: lol, poor @kaberett, @kareila and I are clearly feeling our oats this morning. <3

kaberett: we're cool, I took the opportunity of fixing this up to add myself to the authors list and update the copyright :-p

@kareila, perfectly happy to rebase per previous if you're good to lose all the line comments...

cgi-bin/DW/Console/Command/BonusIcons.pm
```diff
@@ -56,25 +59,54 @@ sub execute {
...
-        my $new = max( $to_u->prop( 'bonus_icons' ) - $count, 0 );
-        $to_u->set_prop( bonus_icons => $new );
+        my $xfer_count = $to_u->prop( 'bonus_icons' );
+        $to_u->set_prop( bonus_icons => 0 );
         LJ::statushistory_add( $to_u, $remote, 'bonus_icons',
-                sprintf( 'Removed %d icons, new total: %d.', $count, $new ) );
-        $self->print( sprintf( 'User now has %d icons.', $new ) );
+                sprintf( 'Transferred %d icons to %s.', $xfer_count,
+                    $destination_u->user ) );
+        my $new_total = $destination_u->prop( 'bonus_ions' ) + $xfer_count;
```
kareila: As amusing as the idea is of catalyzing a chemical reaction, I think you should probably correct this typo. :wink: 

And yeah, go ahead and squish the commits if you like.

kaberett: #preventchemistswritingcode

kaberett:
DONE. HOPEFULLY.

kareila:
YES YOU ARE DONE.

rahaeli:
\o/

PR #1675, opened 2016-02-14
Remove duplicate "about" from Support views

kaberett:
Explanations about what happens to HTML entered in Support requests
does not need to have "about" repeated. This removes it from TT and
replaces all BML strings, just in case.


kareila:
This is good, but you should also remove the .no.html.allowed lines from the text files. You don't need to deadphrase them.

kaberett:
Done!

PR #1677, opened 2016-02-14
Disable logged-out support requests by default

### LINKED ISSUE #1630: Dev suggestion: Allow DW-based sites to disable anonymous (logged-out) support requests
Apologies for using GitHub for a suggestion, but it doesn't really seem appropriate for **dw_suggestions** (as it's not something users would benefit from) and **dw_dev** seems a little bit too open for what could be a potentially sensitive request.

I believe that DW-based sites should be able to restrict support requests to logged-in users only if they choose. My reasoning is that I noticed that today the 'Daily Snapshot' Dreamhack was being bombarded with spammy support requests. Normally, I wouldn't care about this because the Daily Snapshot resets each day anyway.

In this case, however, there's a very important reason to be concerned - the support board automatically fires off a confirmation email when a support request is opened, and very often the email addresses used by spammers for the From and Return-Path email headers are real, working addresses of people who had nothing to do with it and who are thus very likely to mark the email as spam (because, technically speaking, it kind of <em>is</em>).

I could set `$LJ::CAPTCHA_FOR{support_submit_anon}` to 1 in `config.pl`, but that only works with an API key - and that's not something which I think would be wise to deploy over every Dreamhack. Apparently there's a demo key available, but the sort of usage I was seeing before I shut the Dreamhack down (571 support requests in 3 hours) would most likely be far beyond what would be allowed for a demo key.

Basically, I'd like to be able to have the default Dreamhack configuration disallow these anonymous support requests by default, as they end up sending spam unwittingly and I'm not okay with that.

I may come up with a pull request for this myself if people agree that this is worth doing, but I didn't want to solve a problem if people don't think it needs solving!

Sophira:
Although important for production sites, allowing logged-out support requests on development servers can lead to a lot of spam. This change allows logged-out support requests to be disabled, and disables them in the default configuration.

Fixes #1630.

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -809,6 +809,8 @@ error.security.s1.2=Error: Sorry, security filtering is not supported within S1 
 
 error.support.invalid_category=Invalid support category
 
+error.support.mustbeloggedin=New support requests may only be submitted while logged in.
+
 error.support.nonuser=You cannot submit support requests from non-user accounts.
 
 error.support.norequest=You did not enter a support request.
--- a/cgi-bin/DW/Controller/Support/Submit.pm
+++ b/cgi-bin/DW/Controller/Support/Submit.pm
@@ -104,6 +104,14 @@ sub submit_handler {
             if LJ::is_enabled( 'support_request_language' );
     }
 
+    # If not logged in and logged-out requests are disabled, give a fatal error
+    unless ( $remote || LJ::is_enabled( 'loggedout_support_requests' ) ) {
+        my $errors = DW::FormErrors->new;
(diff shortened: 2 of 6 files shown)


Sophira:
Sorry that this took so long for me to do!

kaberett:
I've applied this set of changes to my hack; it's working as desired! Ditto several other folk who were at the contributor weekend -- I can round them up if useful.

zorkian:
Haha, the number of dev environments I've had get spambombed... Thank you :)

kaberett:
\o/

rahaeli:
Fucking spammers.

Issue #1669, opened 2016-01-14
Page title for viewing custom access filters uses filter number not name


kaberett:
i.e. "kaberett | (XX)" rather than "kaberett | (groupname)".

The equivalent issue for reading pages/subscription filters was bug 5320 (and subsequently fixed post-transition to GHI though I can't find any trace of it on Github).

kaberett:
claim

kareila:
I found the line you need to change! It's in LJ/S2/FriendsPage.pm, line 155:
`$p->{filter_name} = $opts->{securityfilter};`

To change that to the name of the group, you'll need to call the trust_groups method:
`my $filter = $u->trust_groups( id => $opts->{securityfilter} );`

Then once you have the filter object, `$filter->{groupname}` will get you what you need.

kaberett:
Ah, excellent! Thank you so so much for digging that out for me.

PR #1678, opened 2016-02-16
Allow viewing of access group filters on your reading page

kareila:
Fixes an issue when attempting to view your reading page filtered
by a custom access group - would fail to validate the group name
and display an error instead of loading the page.

Not related to any existing issue but discovered when attempting to debug #1669.


kaberett:
... that is fantastic and I applaud you. (I also wave from somewhere in the depths of ESN.)

zorkian:
THE MERGE HAMMER

PR #1679, opened 2016-02-16
Indicate entry security in notification e-mails

### LINKED ISSUE #717: Indicate entry security status in comment notifs for locked/filtered entries
I'm pretty sure this already exists in HTML mail, but I get text mail from DW and will until the end of days, so this would be a nice thing to have ;)

When you get a comment notif spawned from a comment on a locked (or filtered) entry, it looks like:

> Username (username) replied to a comment you left in a Dreamwidth entry
> "Entry title" ( entry_url ). The comment
> they replied to was:

There's nothing indicating that the entry is locked, so not only does this wind up with me inevitably winding up staring at the "you do not have permission to view this entry" screen if I click through, because I'm inevitably logged in as the wrong account, there's also the (slim, but still) chance that someone could get a notif spawned from a locked entry, enough after-the-fact for them to forget it was a locked entry, and forward the comment to somebody else without realizing.

Let's change that to explicitly indicate it's locked:

> Username (username) replied to a comment you left in a Dreamwidth entry
> "Entry title" (entry_url) [locked]. The comment
> they replied to was:

I think that placement is probably the best option to work well with any variation of comment type ("replied to a Dreamwidth entry", "replied to your Dreamwidth entry", "replied to a comment you left in a Dreamwidth entry", etc) and stands the best chance of getting noticed, since it's right up front next to the URL and near the buttons you'd have to hit to forward the email in most mail clients.

kaberett:
Fixes #717.

Adds an indication that an entry is locked or filtered to notification
e-mails fired for comment and new entry subscriptions.

Things I am aware of that probably need changing but I'd like some input:

- <s>putting dead phrases into deadphrases.dat</s> working on this right now, commit to come
- <s>updating all the comments in LJ::Comment and LJ::Event::JournalNewEntry to include [[postsecurity]]</s> done
- <s>I am not including status indicators on officialpost notifications</s> (Denise has said this is correct)
- I *think* I've caught everywhere this needs to happen?
- is there an obviously better variable name than "postsecurity"?
- I'm not very happy with new line 1125 of en.dat, but I can't immediately think of a better way to do it given the changes I'm making in LJ::Event::JournalNewEntry


zorkian:
Thanks for doing this!

PR #1682, opened 2016-02-20
Fixes #1671. Needs max and interval set at time of creation

### LINKED ISSUE #1671: Beta update time weirdness: 23:30
One of my folks observed that when you create an entry using the new beta update page between 23:30 and 00:00 (the nearest she could figure), the timestamp sticks at 23:30 and cannot be changed. I'm duplicating the same in GMT -8. 

Things I have tried so far during the affected time range: 
* posting without stickyticky tickied (23:30)
* posting with stickyticky tickied (23:30)
* posting with stickyticky tickied, and whacking the "hey change the time y0!" button (23:30) 
* trying to edit the time post-posting to the correct value of 23:53 or whatever (allowed to type in, but upon loss of focus jumped back to 23:30) 
* trying to edit the time in the box manually before posting (allowed to type in, but upon loss of focus it immediately jumps back to 23:30) 

Of interest, it displays the time correctly at first load of the page, then jumps to 23:30. 
After midnight hit, I thought of editing the time to something that was not 23:30; values before 23:30 worked (such as 23:19, and any value for any other hour I tried). 

Is this the same thing as #1502?

sgsabbage:
Does not stop interval working correctly when selecting time through menu

(Fixes #1671 - adding this everywhere because I can never remember where it needs to actually be)

### DIFF
--- a/htdocs/js/pages/entry/new.js
+++ b/htdocs/js/pages/entry/new.js
@@ -548,6 +548,8 @@ var postForm = (function($) {
         $("#js-entrytime-time").pickatime({
             editable: true,
             format: "HH:i",
+            interval: 1,
+            max: 1439,
 
             trigger: document.getElementById("js-entrytime-time-button"),
             container: '.displaydate-component .picker-output'


kareila:
... I'm going to write this off as a workaround to a weird problem with pickadate.js and not try too hard to understand it.  WFM.

PR #1684, opened 2016-02-20
There is only one United Kingdom

### LINKED ISSUE #1623: Country lists include two United Kingdoms with identical text
For example at `/manage/profile/` -> Location. One has value `GB`, the other `UK`.

kaberett:
Fixes #1623.

We don't need to offer users two United Kingdoms for location. The
superfluous one is removed from the presented list of options. (I...
*think* it's the correct one I'm removing, at any time.)

### DIFF
--- a/cgi-bin/LJ/Widget/Location.pm
+++ b/cgi-bin/LJ/Widget/Location.pm
@@ -265,6 +265,7 @@ sub country_options {
     my %countries;
     # load country codes
     DW::Countries->load( \%countries );
+    delete $countries{UK};
 
     my $options = ['' => $class->ml('widget.location.country.choose'), 'US' => 'United States',
                    map { $_, $countries{$_} } sort { $countries{$a} cmp $countries{$b} } keys %countries];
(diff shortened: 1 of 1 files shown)


cgi-bin/LJ/Widget/Location.pm
```diff
@@ -265,6 +265,7 @@ sub country_options {
     my %countries;
     # load country codes
     DW::Countries->load( \%countries );
+    delete %countries->{UK};
```
kareila: The syntax for this reference needs to be `$countries{UK}` instead of `%countries->{UK}`, which produces the following warning: `Using a hash as a reference is deprecated at cgi-bin/LJ/Widget/Location.pm line 268.`  But apart from that, it is generally doing the right thing in the right place. :heart:

kaberett:
... *facepalm* done!

kaberett:
(Now with slightly better understanding of Perl and something that Actually Works. Test your 1am commits, Alex, it's good for the codebase and it's good for you.)

kareila:
If I might make a suggestion - this is an ideal example for adding a comment in the code explaining why we need to do this, so that whoever comes through here next doesn't end up scratching their head and asking around or digging through commit logs to figure out why we needed to delete UK from the list.

kaberett:
aaaaand done. (I might not be wording well!) Holler if you want me to rebase it as well!

kareila:
Brilliant, thanks!

PR #1685, opened 2016-02-20
Fixes #1626 by storing the different plural phrases on the comment link

### LINKED ISSUE #1626: Quick reply pluralisation is incorrect when going from 0 to 1 and 1 to 2 comments
When going from 0 to 1 comments, the comments link says '1 comments'

When going from 1 to 2 comments, the comments link says '2 comment'.

sgsabbage:
Fixes #1626 

Both core and database use print_readlink, thus changes to both S2 styles.

I would have liked to have use get_plural_phrase() in DW::Controller::RPC::Misc::addcomment_handler - but I couldn't figure that one out, so it seemed easiest to just store the plural phrases on the link.

### DIFF
--- a/htdocs/js/jquery.quickreply.js
+++ b/htdocs/js/jquery.quickreply.js
@@ -119,8 +119,20 @@ jQuery(document).ready(function($) {
 
                                     $readLink
                                         .ajaxtip() // init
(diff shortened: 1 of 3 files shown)


zorkian:
Nice, thank you for doing this correctly (passing all three forms) rather than some regex bananas.

PR #1686, opened 2016-02-20
Fixes #1651 - show hidden readlink on all pages

### LINKED ISSUE #1651: Quick*Reply on non-reading-page views needs to update comment count
Related to #1626 (magical jQuery pluralisation): implementation of Quick*Reply for views other than the reading page as implemented for #1306 does not (but should!) include updating the comment count link, for consistency.

Specifically, when using Quick*Reply from kaberett.dreamwidth.org (or any other journal) on a post with no comments, the comment count indicator does not appear. If one comment exists, the second comment triggers an update to "2 comment" per #1626.

sgsabbage:
Fixes #1651 - as all pages have quickenedinger reply all pages should show the hidden readlink so that it can be updated by the JS.

Removed the specific setter from the FriendsPage as it'll already be set now.

### DIFF
--- a/cgi-bin/LJ/Entry.pm
+++ b/cgi-bin/LJ/Entry.pm
@@ -718,6 +718,7 @@ sub comment_info {
         screened => $has_screened,
         screened_count => $screenedcount,
         show_readlink => $comments_enabled && ( $replycount || $has_screened ),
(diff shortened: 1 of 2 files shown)


zorkian:
When I read the bug I expected the code to be much more complicated. Yay!

PR #1690, opened 2016-02-20
Fixes #1657, add required JS files for icon browsing to Day and Recent pages

### LINKED ISSUE #1657: Quick Reply: clicking on Browse icons posts comments (Recent page)
http://www.dreamwidth.org/support/see_request?id=32462

Clicking on browse icons in the QR form immediately posts the comment. This seems to be true on recent pages but not on reading pages. Haven't tested it on day pages or network pages.

sgsabbage:
Fixes #1657

### DIFF
--- a/cgi-bin/LJ/S2/RecentPage.pm
+++ b/cgi-bin/LJ/S2/RecentPage.pm
@@ -90,6 +90,15 @@ sub RecentPage
         return 1;
     }
 
+    # libs for userpicselect
+    # if we're using the site skin, don't override the jquery-ui theme, as that's already included
+    my @iconbrowser_extra_stylesheet;
+    @iconbrowser_extra_stylesheet = ( 'stc/jquery/jquery.ui.theme.smoothness.css' )
+            unless $opts->{handle_with_siteviews_ref} && ${$opts->{handle_with_siteviews_ref}};
+
+    LJ::need_res( LJ::Talk::init_iconbrowser_js( 1, @iconbrowser_extra_stylesheet ) )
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/S2/RecentPage.pm
```diff
@@ -90,6 +90,15 @@ sub RecentPage
...
 
+    # libs for userpicselect
+    # if we're using the site skin, don't override the jquery-ui theme, as that's already included
+    my @iconbrowser_extra_stylesheet;
+    @iconbrowser_extra_stylesheet = ( 'stc/jquery/jquery.ui.theme.smoothness.css' )
+            unless $opts->{handle_with_siteviews_ref} && ${$opts->{handle_with_siteviews_ref}};
+
+    LJ::need_res( LJ::Talk::init_iconbrowser_js( 1, @iconbrowser_extra_stylesheet ) )
+        if $remote && $remote->can_use_userpic_select;
+
```
kareila: I suspect this works just fine where it is, but for maintenance purposes, can you move this declaration up to be just after the related jquery includes, similar to the placement you used in DayPage.pm?

Duplicated code is my least favorite, but the pain is somewhat less if we can keep the related things together. :disappointed:

kareila: Never mind, this makes me sad and I will just plan on refactoring it.

rahaeli: KARZILLA SMASH

PR #1693, opened 2016-02-20
1661 banned behaviour

### LINKED ISSUE #1661: extend banning functionality to cover contact in other journals/communities
Currently, when you ban someone, that person cannot:

* leave a comment in your journal
* send you a PM
* buy you paid time or points in the shop
* send you points 

They can, however:

* reply to your comments in someone else's journal
* reply to your comments in a community
* reply to your entries in a community

In other words, banning applies to the journal, not to the person behind it. This has historically caused problems: a not-unknown tactic when someone has banned you is to find a comment they made in a mutual friend's journal or a comment or entry they made in an unrelated community, and reply to them there. It was originally done that way so banning someone couldn't affect their ability to comment elsewhere on the site, but it's hard to explain to people and leads to issues.

Banning should be changed to apply everywhere on the site. If Alice has banned Bob, Bob should not be able to:

* post a comment to a post Alice has made in a community
* post a reply to a comment Alice has made in a community or to another journal
* post a reply to a comment Alice has left in Bob's journal (whether it was left before the ban or after)

Some details:

* If Carol has posted an entry in a community and Alice has replied to that entry, Bob should not be able to reply to Alice's comment. Bob should still be able to post a separate top-level reply to Carol's community post.

* If Carol has posted an entry in a community, Alice has replied to that entry, and David has replied to Alice, Bob should not be able to reply to Alice, but should be able to reply to David.

* If Alice has posted a community entry, Bob should not be able to reply to that entry at all, whether it's a top-level reply or in reply to Carol or David.

* If Alice and Bob were formerly friends and Alice had left comments in Bob's journal, Bob should not be able to go back to those comments and reply to them.

I'm sure there are other use cases I'm not thinking of off the top of my head, but basically, if it's something that would fire a comment notification ("Bob replied to your comment in..."), the comment shouldn't be allowed.

The error message can be the same as it is if Bob is trying to comment in Alice's journal after being banned. (I don't know what it is off the top of my head, but something like "You cannot reply to this [comment|community entry], because Alice has banned you from replying to them" or some such is fine.)

swaldman3:
Fixes #1661

### DIFF
--- a/htdocs/talkpost_do.bml
+++ b/htdocs/talkpost_do.bml
@@ -176,6 +176,15 @@ body<=
     return LJ::bad_input($ML{'/talkpost.bml.error.noreply_readonly_journal'})
         if $journalu && $journalu->is_readonly;
 
+    # don't allow user A to reply to a comment by user B if user B has banned user A
+    if ( defined $remote ) {    #skip this if not logged in
+        my $parentu = LJ::load_userid( $parent->{posterid} );
+        return LJ::bad_input( $ML{'.error.banned.reply'} )
+            if ( defined $parentu && $parentu->has_banned( $remote ) );
+            #we check whether parentu is defined because it won't be
+            #if the parent is an anonymous comment.
+    }
+
     ## insertion or editing
     my $wasscreened = ($parent->{state} eq 'S');
     my $err;
--- a/htdocs/talkpost_do.bml
+++ b/htdocs/talkpost_do.bml
@@ -176,6 +176,10 @@ body<=
     return LJ::bad_input($ML{'/talkpost.bml.error.noreply_readonly_journal'})
         if $journalu && $journalu->is_readonly;
 
+    # is the current user banned by the author of the original post?
+    return LJ::bad_input( $ML{'.error.banned.entryowner'} )
+        if ( defined $remote && $entryu->has_banned( $remote ) );  
+
     # don't allow user A to reply to a comment by user B if user B has banned user A
     if ( defined $remote ) {    #skip this if not logged in
         my $parentu = LJ::load_userid( $parent->{posterid} );
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -3301,6 +3301,7 @@ sub init {
     my $parent = {
         state     => $parpost->{state},
         talkid    => $partid,
+        posterid  => $parpost->{posterid},
     };
     my $comment = {
         u               => $up,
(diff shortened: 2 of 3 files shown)


swaldman3:
Boh. Realised that this doesn't work for quick reply at present (which seems to use entirely separate code for checking this stuff). Will revisit as time permits.

htdocs/talkpost_do.bml
```diff
@@ -176,6 +176,19 @@ body<=
     return LJ::bad_input($ML{'/talkpost.bml.error.noreply_readonly_journal'})
         if $journalu && $journalu->is_readonly;
 
+    # is the current user banned by the author of the original post?
+    return LJ::bad_input( $ML{'.error.banned.entryowner'} )
+        if ( defined $remote && $entryu->has_banned( $remote ) );
```
zorkian: Please no parenthesis around postfix conditionals: http://wiki.dreamwidth.net/wiki/index.php/Dev_Programming_Guidelines#Don.27t_use_extraneous_parentheses

htdocs/talkpost_do.bml
```diff
@@ -176,6 +176,19 @@ body<=
...
 
+    # is the current user banned by the author of the original post?
+    return LJ::bad_input( $ML{'.error.banned.entryowner'} )
+        if ( defined $remote && $entryu->has_banned( $remote ) );  
+
+    # don't allow user A to reply to a comment by user B if user B has banned user A
+    if ( defined $remote ) {    #skip this if not logged in
+        my $parentu = LJ::load_userid( $parent->{posterid} );
+        return LJ::bad_input( $ML{'.error.banned.reply'} )
+            if ( defined $parentu && $parentu->has_banned( $remote ) );
```
zorkian: Extraneous parentheses.

htdocs/talkpost_do.bml
```diff
@@ -176,6 +176,19 @@ body<=
...
+            #we check whether parentu is defined because it won't be
```
zorkian: For this comment, please just put the words in the top of the block and use capitalization and spacing. I.e., since this entire block is just doing a single check, it should all be in one comment.

```
# Comment comment
if ( ... ) {
    my $parentu = ...
    return ...
}
```

The idea here is that it means you can read through the comment first, know what the entire block of code does, and then look at the code. (This is because the block is so short, we don't want to spread comment text around as that makes your brain have to switch back and forth. Also, since the code is so short, it is basically self-documenting and if you do need to call something out it should just be on top.)

zorkian:
(Form letter incoming.)

PR #1694, opened 2016-02-20
Display access filter name in journal page title

### LINKED ISSUE #1669: Page title for viewing custom access filters uses filter number not name

kaberett:
Fixes #1669.

The Recent Entries page, when displaying posts to a specific access
filter, now has a title of the form "kaberett | (test)" rather than
"kaberett | (1)".

### DIFF
--- a/cgi-bin/LJ/S2/RecentPage.pm
+++ b/cgi-bin/LJ/S2/RecentPage.pm
@@ -79,8 +79,14 @@ sub RecentPage
     }
 
     if ( $opts->{securityfilter} ) {
+        my $filter = $u->trust_groups( id => $opts->{securityfilter} );
         $p->{filter_active} = 1;
-        $p->{filter_name} = $opts->{securityfilter};
+        if ( defined $filter ) {
+            $p->{filter_name} = $filter->{groupname};
+        } else {
+            # something went wrong; just use the group number
+            $p->{filter_name} = $opts->{securityfilter};
+        }
     } 
 
     my $get = $opts->{'getargs'};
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/S2/RecentPage.pm
```diff
@@ -79,8 +79,9 @@ sub RecentPage
     }
 
     if ( $opts->{securityfilter} ) {
+        my $filter = $u->trust_groups( id => $opts->{securityfilter} );
         $p->{filter_active} = 1;
-        $p->{filter_name} = $opts->{securityfilter};
+        $p->{filter_name} = $filter->{groupname};
```
kareila: This is right and good (and exactly what I suggested you do :wink:) but it occurs to me to wonder, is it possible for trust_groups to fail to load the filter? It would be best to make sure you call `$filter->{groupname}` only if `$filter` is defined:
```
         if ( defined $filter ) {
                $p->{filter_name} = $filter->{groupname};
         } else {
                # something went wrong, just use the group number
                $p->{filter_name} = $opts->{securityfilter};
         }
```
Granted, this will probably never be a problem because we have robust error checking in LJ::Styles, but it's good to develop defensive programming habits to guard against future bugs.

Also I just realized this change only affects RecentPage, but FriendsPage has the same behavior, although you might not be able to reproduce it without #1678 applied.  I'd prefer to apply the same fix in both places before merging this in.

kaberett:
Done! ... er. Will double-check it's plausible in the morning.

PR #1695, opened 2016-02-20
[#1692] fix user method error when not logged in

### LINKED ISSUE #1692: Logging out from community settings gives error, but only on Dreamhacks
- Log in as the owner of a community
- Go to the community's settings page (Organise -> Manage Communities -> Settings)
- Hit the "log out" button in the top right

* the following error message is produced:
`Can't call method "user" on an undefined value at /dreamhack/home/8374-swaldman/dw/htdocs/manage/settings/index.bml line 198. @ newhack.dreamwidth.net`

This is reproducible on up-to-date dreamhacks, but does not occur on production.

kareila:
Make sure $u is defined when doing permission checks for
displaying notification settings for other users. Otherwise
the page doesn't display for logged-out viewers.

Fixes #1692.

### DIFF
--- a/htdocs/manage/settings/index.bml
+++ b/htdocs/manage/settings/index.bml
@@ -211,7 +211,7 @@ body<=
         $given_cat = "display";
     }
 
-    if ( $u->user ne $authas ) {
+    if ( $u && $u->user ne $authas ) {
         # can only impersonate notification settings -
         # if invisible or disabled, print this error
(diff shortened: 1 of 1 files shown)


zorkian:
Although I will say your PR says 'make sure it's defined' but the code doesn't make sure it's defined, it makes sure it's true. (In case this is a bug, I'm mentioning it.)

Issue #1622, opened 2015-11-08
AJAX quick* reply when you can't comment is unsubmittable


chrisboyle:
Spun out of #1577, this relates to any Quick{,er,est,ish,ly,ening,sand,...}Reply where the form would fold back up on success instead of navigating to the entry page, but you're not allowed to reply. Example scenario:
* You're replying from the Reading page.
* You're an OpenID without verified email.
* The AJAX form flashes up an error underneath about how you should verify your email address.
* There is no way on this page to fix the problem - if the problem is even the fixable one where you need to verify email, and you click through to do so, then you've lost your comment.

## Solution 1
Stop you before you waste time typing a comment, e.g. by navigating to `?mode=reply` (which includes the explanation and other login options) instead of opening an unsubmittable form.

## Solution 2
Make submission of such a Quick*Reply navigate to the full-page comment form populated with your comment and the error (you have login options there to fix it). JS can simply do this on receiving an error response instead of flashing the error briefly underneath the form. This would probably involve a second attempt to submit the comment but that should always be safe as the first one has failed. Take care not to do that if the error is a network error rather than receiving an error response, as the submission may in fact have gone through in that case.

## Debate

* Advantage of 1: clues you in before you write your masterpiece.
* Advantage of 1: avoids caring about details of jQuery's error presentation as above.
* Advantage of 2: avoids creating inconsistency where some-but-not-all Reply links on a Reading page will navigate away from it, losing your place.
* Advantage of 2: avoids caring about ia ia S2 fth'agn in making QR thus conditional.

## Solution 3
The quick-reply form opens as usual, but has a prominent notice on it:
> [icon]your_username is not allowed to comment on this post. You can choose a different identity on the next page.

or:

> Only registered users can comment on this post. You can log in on the next page.

* Submitting it goes directly to the full-page form with the login options, which is hopefully not a surprise given the notice.
* There is no prior AJAX attempt, thus removing any worries about double posting.
* ~~This submission includes a parameter e.g. `select_identity=1` meaning don't attempt to post yet, which ensures we don't post when the notice implied it wouldn't, if meanwhile they logged in in another tab.~~
* ~~The top of the full-page form then says: Comment Not Posted / Choose an identity below~~
* All this is achieved simply by greying out the Post Comment button in favour of the More Options button (and including the appropriate notice described above).

## Solution 4
Like 3 but without the form. Clicking Reply shows just the notice and a link to `?mode=reply`, like this:

[...]

PR #1696, opened 2016-02-21
QR: Warn of comment blocking in advance

chrisboyle:
Make QuickReply forms tell you clearly before you write a comment that
you won't be able to post with your current identity.

Fixes: #1622


cgi-bin/LJ/Talk.pm
```diff
@@ -3695,7 +3695,13 @@ sub make_preview {
             unless $_ eq 'body' || $_ eq 'subject' || $_ eq 'prop_opt_preformatted' || $_ eq 'editreason';
     }
 
-    $ret .= "<br /><input type='submit' value='$BML::ML{'/talkpost_do.bml.preview.postcomment'}' />\n";
+    my $post_disabled = $u->does_not_allow_comments_from($remote)
+        || ($u->{'opt_whocanreply'} eq 'reg' && $remote->is_identity && ! ( $remote->is_validated || $u->trusts( $remote )));
```
zorkian: This logic only depends on `$u` and `$remote`. Instead of having it in two places, can we move it into a method on `LJ::User`? Could we just add it to `does_not_allow_comments_from` or does that break some other code? That method _sounds_ like it should just have this behavior, since `opt_whocanreply` is just specifying who is allowed to comment...

chrisboyle: @zorkian Folding it into that method would disable Quick Reply entirely in this situation due to S2.pm:3677 and I think removing that line would require more comprehensive testing and more thought in the area of permission to comment generally.... which @swaldman3 seems to be planning under #1704 and #1661, so perhaps it's best to wait for those to land before worrying about this PR.

zorkian: Okay, so then we can't fold it into that one method, but we could at least abstract this set of logic into a new method? That shouldn't break anything, and at least keeps this code simpler (and ensures that anybody who touches either of these places in the future understands they're related).

chrisboyle:
@zorkian Extracted logic and fixed style in commits to fb4453d.

zorkian:
Thank you so much! Looks good :)

Issue #1688, opened 2016-02-20
Posting to a moderated comm gives landing page with missing subject


cesy:
If you post an entry to a moderated comm, you get to a landing page with the message "Your entry is in the moderation queue. The community administrators will accept or reject it." The next line says "The entry was posted with the following subject:" and doesn't display the subject even if there was one.

kareila:
This sounds vaguely familiar. Can you check and see if it still happens with #1658 applied?

cesy:
It still happens when the subject doesn't have any HTML in.

kareila:
Sorry I wasn't clear. The changes I made to editjournal.bml in #1658 fix a case where the "no subject" text wasn't being printed.  I can test it later if it's not convenient for you to do so right now.

kaberett:
The entries in question have a subject (rather than no subject). There is no additional information at all provided after the colon in the "This entry was posted with the subject:" string. @cesy, could you maybe grab a screenshot if you've still got the page open? <3

cesy:
![dw error 1](https://cloud.githubusercontent.com/assets/570999/13198314/a16a177a-d7fd-11e5-9c07-21b90fc90e49.png)
![dw error 2](https://cloud.githubusercontent.com/assets/570999/13198315/a16d4300-d7fd-11e5-82a8-510681604c3d.png)

kareila:
Does it only happen with the beta entry form?

cesy:
Yes: 
![dw error 3](https://cloud.githubusercontent.com/assets/570999/13199160/7efbeaac-d815-11e5-8f51-b9ce78c74dc7.png)

kareila:
So it looks like the "following subject" text is being printed unintentionally in this case, since it isn't displayed on the older form's success page.

me-and:
Claim. Fix is almost done, too.

PR #1699, opened 2016-02-21
Don't attempt to give subject &c. for posts to moderated communities

me-and:
Per #1688, posting to a moderated community from the beta update page, unlike the old update page, attempts and fails to display the entry subject. Don't do that.

On the grounds that it's another pending change in this area, I've built this on @kareila's branch for #1658. If that gets changed or rejected, it's fairly simple to adapt this fix accordingly.


kareila:
I merged this in manually since Github was reporting a conflict.  Thanks!

PR #1708, opened 2016-03-15
Wow that is not okay language in comments

me-and:


kaberett:
Thank you for sweeping these all up into one PR! I'd only been going to edit LJ::CleanHTML, it hadn't actually occurred to me to grep the codebase for other instances...

srukle:
Yikes. Thank you for the diligence!

rahaeli:
Eeeesh I thought we'd gotten rid of all that a long time ago. Thanks for fixing that!

Issue #1711, opened 2016-03-16
Some JS being served over HTTP not HTTPS on Create Entries page?


kaberett:
https://www.dreamwidth.org/support/see_request?id=32926 -- pne provides a more thorough bug report of this than I'm going to manage, so herewith their words!

> The Internet as a whole is moving towards "All SSL All The Time" (well, TLS).
>
> So I decided to try changing the "http://" to "https://" on the update page. That gave me a green padlock next to the location bar but with a grey "alert" triangle; clicking on that told me that "Firefox has blocked parts of this page that are not secure".
>
> Clicking on AdBlockPro's list of blockable items showed that everything on its list used https:// except for one: http://s.dreamwidth.org/js/md5.js
>
> Other resources from s.dreamwidth.org were loaded over SSL, including one called
>
> https://s.dreamwidth.org/js/??6alib/inputcomplete.js,6alib/json.js,6alib/template.js,userpicselect.js,6alib/datasource.js,6alib/selectable_table.js,rte.js,settingprod.js,ljwidget.js,ljwidget_ippu.js,widget_ippu/settingprod.js,entry.js,poll.js,browserdetect.js,xpost.js,md5.js,login.js?v=1440903277
>
> which contains md5.js, so it should be a simple tweak to either remove md5.js being loaded by itself, or to have it loaded over HTTPS if the containing page is.
>
> Thank you!

rahaeli:
That's weird; I'm pretty sure we should be serving everything via HTTPS.

kareila:
I can reproduce this with the legacy update page.  There is a line in the rendered source that reads, in its entirety:
```html
<script type="text/javascript" src="http://s.dreamwidth.org/js/md5.js"></script>
```
Looks like it is being printed in cgi-bin/LJ/Web.pm in a context where `$LJ::JSPREFIX` hasn't been changed to reflect whether SSL was requested.

rahaeli:
Oho. Any other places that's happening, while we're at it?

kareila:
I searched for JSPREFIX and saw a couple of similar script tags that need checking. One was in LJ::Talk and the other was RTE-related.

kareila:
Thinking about spending some time investigating this if I can get HTTPS working on a dev server.

kareila:
The easiest fix for this was to remove 'http:' from JSPREFIX. The prod configs use a different JSPREFIX from the default and will need to be updated as well.

kareila:
[...]

PR #1714, opened 2016-03-18
Update $JSPREFIX to use $RELATIVE_SITEROOT

kareila:
Fixes issue #1711 in global defaults; configs with custom JSPREFIX will need updating.

Also adds support for a new config variable SSL_PORT without which I could not have tested this.


zorkian:
I feel vaguely like I tried putting JS on relative root sometime... but maybe not.

kareila:
I looked at the prod configs just now and it looks like we tried using relative root everywhere but had problems with resources in HTML emails not working with relative root.  We don't include JSPREFIX resources in emails so that one should be fine!

Issue #1674, opened 2016-02-04
add DeviantArt to profile list


rahaeli:
Suggestion seen on Twitter and I'm not gonna bother with making them go through dw-suggestions :) 

For profile: username should link to username.deviantart.com
For user tag: userhead should link to username.deviantart.com/gallery, username should link to username.deviantart.com

kaberett:
Already exists! And even I got around to putting it in the FAQ: http://www.dreamwidth.org/support/faqbrowse?faqid=87

Want me to tweet back @ them?

rahaeli:
Huh, I didn't see it on the profile page as an option when I checked. (I did look!) Am I just missing something?

kaberett:
... huh, no, okay. So the username tag does exist, but the profile field doesn't.

rahaeli:
right then! Title of issue edited. (The person tweeting only mentioned the profile, not the username tag, anyway.)

srukle:
I might be able to claim this. This is a quick right up of what I think _might _need to be done, but I did not look too extensively.

https://gist.github.com/srukle/08d4e5a70b05d3aadc0d

rahaeli:
You also need to add the userprop to the props list! You can see the last thing that was added to the profile here:

   https://github.com/dreamwidth/dw-free/commit/fe0000fc597cdca99eb9c52b224a4b1976d31d69

srukle:
OK -- so I'll make changes on my hack space today:
https://gist.github.com/srukle/ed1667fdd3eefada4134 -- in bin/upgrading/proplists.dat
https://gist.github.com/srukle/08d4e5a70b05d3aadc0d -- in dw-free/blob/3868ae55348e08737ad2eba430b504d90b649f06/cgi-bin/DW/Logic/ProfilePage.pm

I'll update this space with the result.

rahaeli:
Sounds excellent!

srukle:
Followed this example: https://github.com/dreamwidth/dw-free/commit/fe0000fc597cdca99eb9c52b224a4b1976d31d69

Received this error when going to edit profile:

Kinda messy to go through right now. I'll run a reinstall on Dreamhack and try again.

[...]

PR #1716, opened 2016-03-20
add deviantart #1674

srukle:
resolves issue #1674 

- adds deviantart to profile
- tested on hackspace


kareila:
This looks great and I applaud your perseverance in getting this to work!

I did notice one issue however.  The link given is http://www.deviantart.com/username, which does not appear to work correctly.  I believe it should instead be http://username.deviantart.com/

If you'll tweak that I can get this in.  Thanks!

srukle:
Thanks for looking this over thoroughly, Jen. I made the changes [here](https://github.com/dreamwidth/dw-free/pull/1716/commits/3a2c275e5a1c7835b29d9243decf9497f6c9a023).

PR #1717, opened 2016-03-22
Add referenced FAQ to volunteer support notifs

### LINKED ISSUE #1715: E-mailed Support notifications do not include referenced FAQs
e.g. my notification e-mail for my visible answer at https://www.dreamwidth.org/support/see_request?id=32907 reads:

> A follow-up to the following  support request has been submitted:
> 
> Category: Entries
> Subject: Only 14 days worth of posts?
> User: kaberett
> URL:http://www.dreamwidth.org/support/see_request?id=32907
> Type: Screened Answer
> Text:
> 
> Hi stormcloude,
> 
> As explained in the linked FAQ and as you've observed, the Reading Page
> displays only the most recent 1000 entries from the last 14 days. These
> limits are the same as those on LiveJournal
> (http://www.livejournal.com/support/faq/219.html), for the same reasons of
> issues with server load and site performance. It is possible to bypass
> these limits by filtering the Reading Page by date
> (https://www.dreamwidth.org/support/faqbrowse?faqid=76), but this is a
> feature that is only available to paid accounts on both Dreamwidth and
> LiveJournal because of the demands it places on the servers.
> 
> As such there isn't any setting you can change, other than purchasing paid
> time. Details on how much paid accounts cost are available at
> https://www.dreamwidth.org/support/faqbrowse?faqid=153, and a summary of
> additional features available to paid account holders is provided at
> https://www.dreamwidth.org/support/faqbrowse?faqid=4. Alternatively, you
> can visit individual journals or communities to read their entire archives.
> 
> 
> I'm sorry that this wasn't what you were hoping to hear, and please do let
> me know if you have any further questions.
> 
> Best,
> -kaberett.
> 
> 
> ====
> 
> You can view this request here:
> 
> http://www.dreamwidth.org/support/see_request?id=32907
> 
> You are receiving this email because you've requested notifications of
> changes to support requests. You may change this notification setting here:
> 
> http://www.dreamwidth.org/support/changenotify

There's no indication that https://www.dreamwidth.org/support/faqbrowse?faqid=25 was included as a Reference FAQ, and it would be useful for record-keeping purposes not to have to click through to see this. I suspect this would also be useful from the perspective of users. I suggest adding a line to the effect of:

> FAQ Reference: What is my reading page? (https://www.dreamwidth.org/support/faqbrowse?faqid=25&view=full)

... for plaintext e-mails, with the FAQ title linkified for HTML e-mails.

kaberett:
Fixes #1715.

If a FAQ was referenced in a Support answer, the ticket update
notification e-mail will now include the FAQ title and link.

### DIFF
--- a/cgi-bin/LJ/Support.pm
+++ b/cgi-bin/LJ/Support.pm
@@ -1130,19 +1130,20 @@ sub work {
 
     } elsif ($type eq 'update') {
         # load the response we want to stuff in the email
-        my ($resp, $rtype, $posterid) =
-            $dbr->selectrow_array("SELECT message, type, userid FROM supportlog WHERE spid = ? AND splid = ?",
+        my ($resp, $rtype, $posterid, $faqid) =
+            $dbr->selectrow_array("SELECT message, type, userid, faqid FROM supportlog WHERE spid = ? AND splid = ?",
                                   undef, $sp->{spid}, $a->{splid}+0);
 
         # set up $show_name for this environment
         my $show_name;
         if ( $posterid ) {
-            my $u = LJ::load_userid ( $posterid );
+            my $u = LJ::load_userid( $posterid );
             $show_name = $u->display_name if $u;
         }
 
         $show_name ||= $sp->{reqname};
 
+        # set up $response_type for this environment
         my $response_type = {
             req => "New Request", # not applicable here
             answer => "Answer",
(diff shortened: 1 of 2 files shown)


kaberett:
NB I've split out the generation of the body of the e-mail for updates to put sorting-out-the-faqref in its own block of code, per sending out e-mail to the ticket submitter (see line 1000 onward). I'm not convinced this is particularly tidy, though, and the code repetition makes me wince; but the only way I can think of to do it while keeping generating-the-body-of-the-email-in-one-block is to include "FAQ REFERENCE: [none]" by default if no FAQ is referenced.

Thoughts/feedback/preferences v welcome!

cgi-bin/LJ/Support.pm
```diff
@@ -1152,15 +1153,39 @@ sub work {
...
+            $lang = LJ::Support::prop( $spid, 'language' )
+                if LJ::is_enabled( 'support_request_language' );
+            $lang ||= $u->prop( 'browselang' ) if $u;
+            $lang ||= $LJ::DEFAULT_LANG;
+            
+            # now actually get the FAQ
+            my $faq = LJ::Faq->load( $faqid, lang => $lang );
+            if ( $faq ) {
+                $faq->render_in_place;
+                my $faqref = $faq->question_raw . " " . "$LJ::SITEROOT/support/faqbrowse?faqid=$faqid&view=full";
```
zorkian: Aw does `LJ::Faq` not have a `url_full` method or something? Bummer.

cgi-bin/LJ/Support.pm
```diff
@@ -1152,15 +1153,39 @@ sub work {
...
+                # now add it to the e-mail!
+                $body .= "\n" . LJ::Lang::ml( "support.email.notif.update.body.faqref", {
+                        faqref => $faqref
+                } );
+                $body .= "\n";
+            }
+        }
+        $body .= LJ::Lang::ml( "support.email.notif.update.body.text", {
+                text => $resp
+           } );
```
zorkian: This should be dedented to match up with `$body`.

cgi-bin/LJ/Support.pm
```diff
@@ -1152,15 +1153,39 @@ sub work {
...
-                type => $response_type,
                 username => LJ::trim( $show_name ),
                 url => "$LJ::SITEROOT/support/see_request?id=$spid",
-                text => $resp
+                type => $response_type
             } );
+        if ( $faqid ) {
+            # need to set up $lang
+            my $lang;
+            my $u = LJ::load_userid( $posterid ) if $posterid;
```
zorkian: So, `my $x = foo if $bar;` is a trap in some versions of Perl, so we don't allow it. See: http://wiki.dreamwidth.net/wiki/index.php/Dev_Programming_Guidelines#Postfix_conditionals

You can just do `my ( $lang, $u );` above and then `$u = LJ::load_userid( $posterid ) if $posterid;` in this line.

cgi-bin/LJ/Support.pm
kaberett: This is how it's done in 1011 of LJ::Support too and I've just had a look and there super isn't a url_full method. Want me to file the bug? :-)

cgi-bin/LJ/Support.pm
kaberett: oops.

kaberett:
aaaand done!

cgi-bin/LJ/Support.pm
kaberett: Thanks for the link!

kaberett:
oh dammit, meant to rebase. keep forgetting that step. please hold ;)

cgi-bin/LJ/Support.pm
zorkian: It's a good babydev thing, so sure! It's a nice obvious kind of cleanup that would improve the readability of the code, IMO.

zorkian:
Thanks for fixing it up! <3

PR #1701, opened 2016-02-22
Clarify extradata options on entry success

kaberett:
Related to #1641 and friends.

* Security information was misleading when posting to a community
rather than a personal account.
* When posting to custom filter(s), the names of the filter(s) in
question are now specified.
* Editing entries using the BML create entries page still provided
an empty string if no subject was provided - I overlooked this in
my earlier pull request #1641, sorry.
* This PR has conflicts with #1699 and #1658, which I am expecting
to resolve manually once they have been merged with develop.

(I note that there's a lot of duplicated code, and that how editjournal.bml and update.bml handle generating *exactly the same text* is unfortunately different -- not sure how best to go about fixing this, though, nor if I've been too verbose/could usefully rephrase how I choose whether to display filters.)

PR #1721, opened 2016-03-25
Clarify extradata options on entry success

kareila:
Refactor and simplify @kaberett's code from #1701 for displaying different success text depending on the security status of the entry, and whether it was posted to a personal journal vs. a community.


zorkian:
Nice cleanup!

PR #1724, opened 2016-03-28
[#1723] proxy URLs included in srcset attributes

### LINKED ISSUE #1723: Some http:// image URLs not being run through the image proxy
I've been noticing that a lot of photos from feeds have not been showing up correctly on my reading page when I use https://. At first I thought the proxy was broken, but today I dug into it and found out it's not quite that!

It turns out we proxy image URLs in the `src` attribute, but not those in the `srcset` attribute (which is used for responsive images). So, for example:

    

Becomes:

    

We should handle both but this is going to be just a little bit tricky because we'd have to parse out the URLs first.

(hullo)

kareila:
Fixes #1723 in my testing; although I don't have a working proxy on my dev machine, I verified that this formed correct proxy URLs for the given test case.

Also updates SSL-related defaults in config-local.pl.  We decided on USE_HTTPS_EVERYWHERE instead of USE_SSL_EVERYWHERE, and USE_SSL wasn't included.  It took me a frustratingly long time to realize that `https_url` wasn't working because I didn't have USE_SSL enabled in my config.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -136,7 +136,7 @@ sub clean
     my $remove_abs_sizes = $opts->{remove_abs_sizes} || 0;
     my $remove_fonts = $opts->{'remove_fonts'} || 0;
     my $blocked_links = (exists $opts->{'blocked_links'}) ? $opts->{'blocked_links'} : \@LJ::BLOCKED_LINKS;
-    my $blocked_link_substitute = 
+    my $blocked_link_substitute =
         (exists $opts->{'blocked_link_substitute'}) ? $opts->{'blocked_link_substitute'} :
         ($LJ::BLOCKED_LINK_SUBSTITUTE) ? $LJ::BLOCKED_LINK_SUBSTITUTE : '#';
     my $suspend_msg = $opts->{'suspend_msg'} || 0;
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -825,9 +825,20 @@ sub clean
                         ! defined $hash->{height}) { $img_bad ||= $opts->{imageplaceundef}; }
                     if ($opts->{'extractimages'}) { $img_bad = 1; }
 
-                    my $url = canonical_url($hash->{src}, 1);
-                    $url = https_url( $url, journal => $journal, ditemid => $ditemid ) if $LJ::IS_SSL;
-                    $hash->{src} = $url;
+                    my $sanitize_url = sub {
+                        my ( $url ) = @_;
+                        $url = canonical_url( $url, 1 );
```
zorkian: Nit: Isn't this section just,

```
my $url = canonical_url( $_[0], 1 );
```

kareila: Sure, those are equivalent.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -825,9 +825,20 @@ sub clean
...
+                        return $url unless $LJ::IS_SSL;
+                        return https_url( $url, journal => $journal, ditemid => $ditemid );
+                    };
+
+                    $hash->{src} = $sanitize_url->( $hash->{src} );
+
+                    # some responsive images use srcset as well as src;
+                    # both attributes should be proxied for https if requested
+                    if ( defined $hash->{srcset} ) {
+                        $hash->{srcset} =~ s!\b(http://\S+)!$sanitize_url->( $1 )!egi;
```
zorkian: TIL what `srcset` does and contains. Neat.

cgi-bin/LJ/CleanHTML.pm
kareila: Tweaked this and updated.

zorkian:
Thank you! <3

afuna:
Yay thank you I'm pretty excited about this :)

Maybe we can turn on https everywhere ;)

PR #1725, opened 2016-03-28
Update LiveJournal.pm comments

azurelunatic:
Comment language lacked that certain something.


rahaeli:
Thank you for catching that!

PR #1728, opened 2016-04-01
[#1367] make rename acct reusable when using rename_opts cmd

### LINKED ISSUE #1367: need some admin way to break rename redirect while freeing up redirected username for immediate re-use
If a user chooses the wrong options while renaming, they can wind up not being able to do what they want to do, and I can't fix it.

Scenario: User has 3 accounts, A, B, and C. They want to rename B to C, and then A to B. (And don't care what happens to A afterwards.)

If they do the B->C rename first, and accidentally select the redirect option when renaming, they are then locked out of renaming A->B because username B is still technically 'in use' (as a redirect). Even if I use the rename admin tool to break the redirect, the user still can't perform the rename.

If I use the rename admin tool to break a redirect from old->new, the 'old' username should be freed up for re-use immediately (via rename token, same as any other deleted account).

kareila:
Updates the rename_opts console command to expunge a broken
redirect account instead of merely deleting it, so that its
username can be renamed to by someone else.

Also some minor housekeeping: whitespace cleanup, module
header cleanup, and fixing a typo in the reference.

Fixes #1367.

### DIFF
--- a/cgi-bin/DW/Console/Command/RenameOpts.pm
+++ b/cgi-bin/DW/Console/Command/RenameOpts.pm
@@ -1,7 +1,8 @@
 #!/usr/bin/perl
 #
 # DW::Console::Command::RenameOpts
-# This module 
+#
+# Console command for tweaking options on renamed users.
 #
 # Authors:
 #      Afuna <coder.dw@afunamatata.com>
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Console/Command/RenameOpts.pm
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/perl
 #
 # DW::Console::Command::RenameOpts
-# This module
```
zorkian: THIS MODULE

I wonder if a linter that looks for bad grammar would be more useful or annoying (maybe not for DW, but in general).

cgi-bin/DW/Console/Command/RenameOpts.pm
```diff
@@ -65,7 +66,7 @@ sub execute {
 
         if ( $cmd eq 'break_redirect' )  {
             if ( $u->break_redirects ) {
-                $u->set_deleted;
+                $u->set_expunged;
```
zorkian: This only gets used on type R accounts right? I.e., the "fake" accounts that currently redirect to people. If so, I _think_ this is safe technically... it tickles my feelers and makes me worry slightly, but I think I've convinced myself it's OK technically.

PR #1738, opened 2016-04-07
[#1734] always set opt_preformatted to a defined value when crossposting

### LINKED ISSUE #1734: Crossposter failing to correctly handle un/checking the Disable Auto-Formatting option (old Create Entries, HTML editor)
http://www.dreamwidth.org/support/see_request?id=33047

If you make a post to DW that is cross-posted to LJ with the Disable Auto-Formatting box (in the HTML editor of the old Create Entries page) checked, and subsequently edit that post to uncheck the Disable Auto-Formatting box, the crossposter fails to handle the edit elegantly.

In my test case (private post; no comments on either the DW or LJ entry) the line-breaks haven't been inserted on the LJ copy despite attempting to crosspost the edited version, with no failure notification in my DW inbox. (I made no other changes to the post; I am set up to use the HTML editor by default on LJ.)

In the case of the user reporting this issue, there were comments on (at least) the LJ version of the post; unchecking the DA-F box and crossposting the edited version reportedly created a new post on LiveJournal with correct formatting, rather than editing the previous post.

kareila:
When the checkbox was cleared, the entry property was undefined,
so the effective value wasn't being passed along when editing a
crosspost, and the preformatting on the remote site could never
be undone.

Fixes #1734.

### DIFF
--- a/cgi-bin/DW/External/XPostProtocol/LJXMLRPC.pm
+++ b/cgi-bin/DW/External/XPostProtocol/LJXMLRPC.pm
@@ -337,9 +337,12 @@ sub entry_to_req {
     my $entryprops = $entry->props;
     $req->{props} = {};
     # only bring over these properties
-    for my $entrykey (qw ( adult_content current_coords current_location current_music opt_backdated opt_nocomments opt_noemail opt_preformatted opt_screening used_rte pingback )) {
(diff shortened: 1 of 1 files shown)


zorkian:
Excitingly weird kind of bug.

PR #1739, opened 2016-04-10
Fix #1528: failure to untag shouldn't fail rest of edit but should fail a tag-add

### LINKED ISSUE #1528: Unclear or absent error/success messages relating to tag removal permissions
Specifically, relating to behaviours when attempting to add or remove tags from posts in commnities with restricted tagging permissions. Behaviour described occurs for user with permission to add (but not create) tags but not to remove them.

_Via /edittags:_ delete tag from _Current tags_ bar. Hit "Save changes". You are returned to the entry page with no error message; the tag you attempted to delete is still in place. Any tags you added from the existing list have been added.

_Via beta Edit Entry page (if you are post-owner):_ Remove tag from tag module. Hit "Save". Edit Entry page reloads with error message at top "Tags error: You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tag]]." <i>However</i>, any additional tags <i>added</i> are now visible in all other views of the entry (and possibly edits of body text/subject similarly go through - haven't tested) - this success is not indicated in any way. Suggest error message shows up on entry success landing page instead.

_Via old update/edit entries page:_ Not tested, sorry, don't quite have brain.

chrisboyle:
See commit messages; I've gone with @kaberett's suggested behaviour. Here's what it looks like from the beta editor:

from-new

...and from editjournal.bml:

from-old

Fixes: #1528

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -3983,7 +3983,7 @@ taglib.error.access=You are not allowed to tag entries in this journal.
 
 taglib.error.add=You are not allowed to create new tags for this journal; the entry was not tagged with [[tags]].
 
-taglib.error.delete=You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tags]].
+taglib.error.delete2=You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tags]].  Any tags you tried to add have not been added.
 
 taglib.error.exists=The tag name '[[tagname]]' is already in use and cannot be renamed to. You can merge the tags instead.
 
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -3983,7 +3983,11 @@ taglib.error.access=You are not allowed to tag entries in this journal.
 
 taglib.error.add=You are not allowed to create new tags for this journal; the entry was not tagged with [[tags]].
 
-taglib.error.delete2=You are not allowed to delete tags from entries for this journal; any tags you tried to add have not been added.  The entry is still tagged with [[tags]].
+taglib.error.delete2<<
+You are not allowed to delete tags from entries for this journal; any tags you tried to add have not been added.
+
+The entry is still tagged with: [[tags]].
+.
 
 taglib.error.exists=The tag name '[[tagname]]' is already in use and cannot be renamed to. You can merge the tags instead.
 
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -3983,7 +3983,7 @@ taglib.error.access=You are not allowed to tag entries in this journal.
 
 taglib.error.add=You are not allowed to create new tags for this journal; the entry was not tagged with [[tags]].
 
-taglib.error.delete2=You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tags]].  Any tags you tried to add have not been added.
+taglib.error.delete2=You are not allowed to delete tags from entries for this journal; any tags you tried to add have not been added.  The entry is still tagged with [[tags]].
 
 taglib.error.exists=The tag name '[[tagname]]' is already in use and cannot be renamed to. You can merge the tags instead.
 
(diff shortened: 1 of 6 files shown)


kareila:
I agree this is an improvement, but wouldn't it be better to have the information on the success page appear in a consistent order?  I don't have any particular opinion on what the order should be, just that it should be the same on both pages.

bin/upgrading/en.dat
```diff
@@ -3983,7 +3983,7 @@ taglib.error.access=You are not allowed to tag entries in this journal.
 
 taglib.error.add=You are not allowed to create new tags for this journal; the entry was not tagged with [[tags]].
 
-taglib.error.delete=You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tags]].
+taglib.error.delete2=You are not allowed to delete tags from entries for this journal; the entry is still tagged with [[tags]].  Any tags you tried to add have not been added.
```
kareila: I think this information would be more easily parsed if the list of tags continued to be at the end of the line.  So something like: "You are not allowed to delete tags from entries for this journal; any tags you tried to add have not been added. The entry is still tagged with [[tags]]."

Might also be better to break it into two lines at this point.

cgi-bin/LJ/Protocol.pm
```diff
@@ -2157,8 +2166,8 @@ sub editevent
             $u->do( 'INSERT INTO logslugs (journalid, jitemid, slug) VALUES (?, ?, ?)',
                     undef, $ownerid, $itemid, $slug );
             if ( $u->err ) {
-                $res->{message} ||= 'Sorry, it looks like that slug has already been used. ' .
-                    'Your entry has been updated, but you can still edit it again to add a unique slug.';
+                $add_message->( 'Sorry, it looks like that slug has already been used. ' .
+                    'Your entry has been updated, but you can still edit it again to add a unique slug.' );
```
kareila: If you're using this add_message subroutine to potentially return multiple errors from one edit, that would be another argument in favor of breaking up the messages onto multiple lines.

kareila:
I don't see any major problems, just a few suggestions for improvements before accepting the changes.  Thanks for working on this!

bin/upgrading/en.dat
chrisboyle: I agree, but I think with that phrasing I'd expect the list of tags to be the entry's full list, where currently it's just the ones you tried to remove - maybe that's an improvement in any case. So:

> You are not allowed to delete tags from entries for this journal; any tags you tried to add have not been added.
>
> The entry is still tagged with [[all_of_its_tags]].

chrisboyle:
@kareila All done. :-)

cgi-bin/LJ/Tags.pm
```diff
@@ -820,13 +818,10 @@ sub update_logtags {
...
-    push @skippable_errors, LJ::Lang::ml( "taglib.error.add", { tags => join( ", ", @unauthorized_add ) } ) if @unauthorized_add;
-    push @skippable_errors, LJ::Lang::ml( "taglib.error.delete", { tags => join( ", ", map { $utags->{$_}->{name} } keys %delete ) } ) if %delete && ! $can_control ;
-    $err->( join " ", @skippable_errors ) if @skippable_errors;
-
-    # but delete nothing if we're not a controller
-    %delete = () unless $can_control || $opts->{force};
+    my @add_delete_errors;
+    push @add_delete_errors, LJ::Lang::ml( "taglib.error.add", { tags => join( ", ", @unauthorized_add ) } ) if @unauthorized_add;
+    push @add_delete_errors, LJ::Lang::ml( "taglib.error.delete2", { tags => join( ", ", map { $utags->{$_}->{name} } keys %{$tags} ) } ) if %delete && ! $can_control && ! $opts->{force};
+    return $err->( join " ", @add_delete_errors ) if @add_delete_errors;
```
kareila: This should be `join "\n\n"` instead of `join " "` to get the desired newlines between errors.

Apart from that, this looks ready to go!

chrisboyle:
@kareila Thanks, fixed the join.

PR #1741, opened 2016-04-11
Release 1.28

kareila:
This includes all the live fixes currently in production that are not yet committed.


zorkian:
Thanks for cleaning this up. <3

PR #1743, opened 2016-04-11
Cleanup 2015q4

kareila:
I went to start a new log warning cleanup branch and realized I had one from several months ago that never got submitted.  Shouldn't be anything too tricky in here, just the usual uninitialized variables and non-numeric comparisons.


zorkian:
Yeah nothing looks crazy here. Merged!

PR #1749, opened 2016-04-13
don't test $LJ::HTTPS_UPGRADE_REGEX unless it is defined

kareila:
Otherwise the conditional will always be true.

This made t/post.t start failing.  Yay tests!


zorkian:
Yay tests boo @zorkian! D:

PR #1750, opened 2016-04-16
Remove customview from codebase

kaberett:
Fixes Bugzilla bug 2324.

customview has always been a hack (to allow users to use their own
domains?) and has never been supported on Dreamwidth; as of 2011 no
reports had been received of customview being used or needed, and
the decision was taken to remove the feature rather than begin
supporting it. This commit applies the code machete; visiting URLs of
the form e.g.
http://kaberett.kaberett.hack.dreamwidth.net/data/customview/?s2id=148
now produces a site-skinned 404 instead of a black-on-white error
message about account permissions.

### DIFF
--- a/cgi-bin/DW/FeedCanonicalizer.pm
+++ b/cgi-bin/DW/FeedCanonicalizer.pm
@@ -29,7 +29,7 @@ my %LJISH_SITES = map { $_ => 1 } (
 );
 
 my $LJISH_URL_PART = "(data/(?:rss|atom)(?:_friends|\.xml|\.html)?|"
-    . "data/customview|rss(?:/friends|/data|\.xml|\.html)?)(/.+?)?(.+?)?";
+    . "rss(?:/friends|/data|\.xml|\.html)?)(/.+?)?(.+?)?";
 
 sub canonicalize {
     my $uri_string = $_[0];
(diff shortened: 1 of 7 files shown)


kaberett:
There are two remaining instances of customview in the codebase and I'd quite like advice/input about them, please!

1. cgi-bin/DW/FeedCanonicalizer.pm features a regex that appears to optionally match on URLs including data/customview. I don't speak regex well enough to be confident editing this.

2. A comment in LJ::CleanHTML references both S1 and customview. I've added an extra line to the comment noting that neither of these are actually relevant to Dreamwidth, but that's clumsy; I'd rather do better but I'm not entirely sure how.

Input  much appreciated! <3

cgi-bin/DW/FeedCanonicalizer.pm
```diff
@@ -318,26 +318,16 @@ sub make_ljish {
...
-    } else {
-        my $extra;
-        $extra = "/friends" if $feed =~ /friends$/;
-
-        my %query = $uri->query_form;
-
-        if ( $query{tag} ) {
-            return "ljish://$domain/$username$extra?tag=" . uri_escape( $query{tag} );
-        }
+    my $extra;
```
kareila: This should keep the ="" assignment to avoid warnings about using undefined variables in string concatenation.

cgi-bin/DW/FeedCanonicalizer.pm
```diff
@@ -318,26 +318,16 @@ sub make_ljish {
...
+    $extra = "/friends" if $feed =~ /friends$/;
+
+    my %query = $uri->query_form;
 
-        return "ljish://$domain/$username$extra";
+    if ( $query{tag} ) {
+        return "ljish://$domain/$username$extra?tag=" . uri_escape( $query{tag} );
```
kareila: Alternatively this could be `$extra .= "?tag=" . uri_escape( $query{tag} );` and then you don't have to repeat the `return "ljish://$domain/$username` bit.

cgi-bin/LJ/User/Styles.pm
```diff
@@ -616,9 +616,6 @@ sub make_journal {
     if ($opts->{'vhost'} =~ /^other:/ && ! LJ::get_cap($u, "domainmap")) {
         return $notice->( BML::ml( 'error.vhost.noalias' ) );
     }
-    if ($opts->{'vhost'} eq "customview" && ! LJ::get_cap($u, "styles")) {
-        return $notice->( BML::ml( 'error.vhost.nostyle' ) );
```
kareila: error.vhost.nostyle should be removed from en.dat and added to deadphrases.dat.

kareila:
Regarding cgi-bin/DW/FeedCanonicalizer.pm:
```
my $LJISH_URL_PART = "(data/(?:rss|atom)(?:_friends|\.xml|\.html)?|"
    . "data/customview|rss(?:/friends|/data|\.xml|\.html)?)(/.+?)?(.+?)?";
```
Each pipe character acts as an "or" so you can just remove the relevant bit:
```
my $LJISH_URL_PART = "(data/(?:rss|atom)(?:_friends|\.xml|\.html)?|"
    . "rss(?:/friends|/data|\.xml|\.html)?)(/.+?)?(.+?)?";
```
(And yes that is a hideous regexp.)

kareila:
I would leave that bit in LJ::CleanHTML alone for now apart from the added comment, since IDK if that code block is being used for any other purpose - maybe tag the comment with FIXME to indicate it needs more scrutiny.

I think that's everything for my first pass!

kaberett:
Changes made!

PR #1751, opened 2016-04-16
Move DW::BusinessRules::InviteCodes::DWS to -free

kaberett:
Fixes Bugzilla bug 3955 (with corresponding dw-nonfree pull request).

The Bugzilla bug said:

> Following discussion in http://dw-lounge.dreamwidth.org/57318.html,
> the reasons for having separate bits in dw-free and dw-nonfree no
> longer apply, and Denise and Mark agreed to have that code GPL'd.
>
> The easiest way is probably to keep both modules, DW::BusinessRules::InviteCodes
> and DW::BusinessRules::InviteCodes::DWS, in separate files, just
> moving the dw-nonfree file to dw-free. This will keep the API
> documentation intact and minimize the risk of any side-effects for
> DW. (Note that this may cause problems for other sites that have
> their own DW::BusinessRules::InviteCodes::Foowidth. They would
> need to either delete DW::BusinessRules::InviteCodes::DWS or rename
> DW::BusinessRules::InviteCodes::Foowidth to DW::BusinessRules::InviteCodes::DWS
> in their local repo, so we need a warning in dw_dev when the change
> is committed.)

This may also affect any API work.

### DIFF
--- a/cgi-bin/DW/BusinessRules/InviteCodes/DWS.pm
+++ b/cgi-bin/DW/BusinessRules/InviteCodes/DWS.pm
@@ -1,0 +1,225 @@
+#!/usr/bin/perl
+#
+# DW::BusinessRules::InviteCodes::DWS
+#
+# This module implements business rules for invite code distribution that are
+# specific to Dreamwidth Studios, LLC
+#
+# Authors:
+#      Pau Amma <pauamma@dreamwidth.org>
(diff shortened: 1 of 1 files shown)


kaberett:
https://github.com/dreamwidth/dw-nonfree/pull/110

PR #1752, opened 2016-04-16
Specify color_page_text for all Basic Boxes themes

kaberett:
Fixes Bugzilla bug 5217.

Sets color_page_text explicitly for all Basic Boxes themes.


kaberett:
Uses the colour specified by @ninetyd in the original ticket -- please holler if anything's changed and it should Be Something Else! <3

PR #1753, opened 2016-04-16
Tidy up CSS credit for Leftovers and Patsy

kaberett:
Fixes Bugzilla bug 5223.

Replaces tab characters with spaces, and removes "please do not
redistribute" note and credit to pinvoke (not used), per @ninetyd's
comments on the original bug.

### DIFF
--- a/styles/leftovers/layout.s2
+++ b/styles/leftovers/layout.s2
@@ -77,11 +77,10 @@ function print_stylesheet () {
 /*-------------------------------------
 
 
-	layout name: leftovers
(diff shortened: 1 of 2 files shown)


kaberett:
(I note that the parts of the file I've edited don't show up in e.g. basicboxes/layout.s2, but I'm going to need someone who knows More About Styles Than Me to tell me if the entire section just wants removing for standardisation's sake!)

kareila:
Looks like these are just file comments - I don't see a problem, but it can always be revisited later.

PR #1760, opened 2016-04-30
List "Other Services" in neatly formatted columns

### LINKED ISSUE #1702: List "Other Services" in columns on profile pages
Currently the "Other Services" list on profile pages (e.g. [mine](http://me-and.dreamwidth.org/profile)) is a single column with a lot of whitespace on the right hand side. I think it'd look neater formatted into multiple columns.

Currently the list is formatted using a table; I think the sensible way to implement this would be to switch to formatting using `<ul>`, then using CSS column width properties to do the formatting; that should mean it looks fine regardless of the screen size – you specify the column width in CSS and the browser formats the content into an appropriate number of columns for the screen – and will degrade to a vertical list on old browsers that don't support CSS column widths.

If we want example code, that sort of CSS formatting is how a lot of reference lists on Wikipedia are formatted. [The article on CN Lester](https://en.wikipedia.org/wiki/CN_Lester), for example, has 30em wide columns in its references section.

Per discussion on IRC, this didn't seem to quite warrant a suggestion in dw_suggestions, so I'm just opening an issue to see if anyone can spot any glaring problems with the plan.

me-and:
Fixes #1702.

Rather than using tables to render the list of "Other Services" on profile pages, use a `ul`. While we're at it, format the list into columns on devices that support it.

AIUI, this will:

* Render columns on any vaguely recent version of most browsers, with the exception of IE9 and earlier, which don't have support for columns.
* Render a list that looks very similar to the current table display on graphical browsers that don't support columns (looking at you, IE)
* Be considerably more accessible than the old table-based layout for users with screen readers or text-only browsers.

I've picked a width of 15em as something that looks vaguely reasonable to me, and I've made sure the behaviour is reasonable for over-sized usernames. You can currently see what it looks like on my Dreamhack at http://system.meandering.hack.dreamwidth.net/profile.

### DIFF
--- a/htdocs/profile.bml
+++ b/htdocs/profile.bml
@@ -601,16 +601,16 @@ body<=
     my $imlist;
     foreach my $service ( $profile->external_services ) {
         my $title = $ML{$service->{title_ml}};
-        $imlist .= "<tr class='im_$service->{type}'>";
-        $imlist .= "<td class='im_icon'><img src='$LJ::IMGPREFIX/profile_icons/$service->{image}' alt=\"$title\" title=\"$title\" /></td>";
+        $imlist .= "<li class=\"im_$service->{type}\">";
+        $imlist .= "<div class=\"im_icon\"><img src=\"$LJ::IMGPREFIX/profile_icons/$service->{image}\" alt=\"$title\" title=\"$title\" /></div>";
         if ( $service->{status_image} ) {
             my $status_title = $ML{$service->{status_title_ml}};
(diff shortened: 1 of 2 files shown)


kareila:
This looks quite good, thank you!  The only thing I would add is some padding between the username and the contents of the "im_status" div.  There's no whitespace separating the two elements when I view it in Chrome or Safari.

me-and:
Good point. Done that!

Issue #1742, opened 2016-04-11
directory search shenanigans


kareila:
http://www.dreamwidth.org/support/see_request?id=33110

Doing a directory search for journals located in North Carolina, updated at any time, returns no results.

https://www.dreamwidth.org/directorysearch?opt_format=simple&opt_pagesize=200&start_search=1&widget%5Bgeosearchlocation%5D_country=US&widget%5Bgeosearchlocation%5D_statedrop=NC

Doing a directory search for journals located in North Carolina, updated in the past 7 days, returns tons of results, none of which seem to be located in North Carolina.  I'm not sure any of the results have any location set at all; most are syndicated feeds.

https://www.dreamwidth.org/directorysearch?opt_format=simple&opt_pagesize=200&start_search=1&ut_days=7&widget%5Bgeosearchlocation%5D_country=US&widget%5Bgeosearchlocation%5D_statedrop=NC

If the directory search is this egregiously broken, we should either figure out how to fix it, or give up and replace it with something that's actually possible to get working in development environments.

rahaeli:
My preference is for replacement and has been for a long time; the directory's badly integrated and we have too many search options (none of which talk to each other).

(Ideally I want a replacement that can do tag search the way Tumblr does.)

kaberett:
Pedantry alert:

The way tumblr *currently* does, or the way tumblr *did* up until a couple of months ago?

rahaeli:
The way where you can search & browse the whole site for any post tagged with [whatever] and read them all together, like how the Latest Things tag feed works but for literally any tag.


On Apr 11, 2016, at 2:58 PM, kaberett <notifications@github.com> wrote:

> Pedantry alert:
> 
> The way tumblr currently does, or the way tumblr did up until a couple of months ago?
> 
> �
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> 

-- 
Denise Paolucci
denise@dreamwidth.org
Dreamwidth Studios: Open Source, open expression, open operations. Join us: www.dreamwidth.org

alierak:
The wiki instructions for setting up directory search are missing info on the directory-meta worker, and the wiki list of workers needs some help. Unfortunately I don’t seem to be able to edit the wiki…

http://wiki.dreamwidth.net/wiki/index.php/Search/UserSearch <http://wiki.dreamwidth.net/wiki/index.php/Search/UserSearch>

[...]

PR #1764, opened 2016-05-04
[#1742] don't lowercase all get_args keys in LJ::create_url

kareila:
This is the conservative version that only preserves case sensitivity for known case-sensitive arguments, and logs a warning if any other case changes are requested from clicked links. Hand-typed requests are still silently converted to lowercase as originally intended, for now.

I'd like to gather some data from the production logs and add any other needed exceptions before going on to the next step, which would be to hopefully eliminate lowercasing in create_url, and move the logging up a level into get_args.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -1096,7 +1096,13 @@ sub create_url {
         foreach my $k ( keys %$orig_args ) {
             next if $k eq lc $k;
             next if $k =~ /^Widget\b/;
-            warn "[#1742] lowercasing $k, url: $url\n";
+            # already found another unintended side effect in /endpoints/draft
+            next if $k eq 'getProperties';
+
+            # don't complain about hand-typed arguments, but do convert them (for now)
+            my $referer = DW::Request->get->header_in("Referer");
+            warn "[#1742] lowercasing $k, url: $url (referer: $referer)\n"
+                if $referer;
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/Web.pm
```diff
@@ -1088,7 +1088,25 @@ sub create_url {
...
+    unless ( $opts{cur_args} ) {
+        foreach my $k ( keys %$orig_args ) {
+            next if $k eq lc $k;
+            next if $k =~ /^Widget\b/;
+            # already found another unintended side effect in /endpoints/draft
+            next if $k eq 'getProperties';
+
+            # don't complain about hand-typed arguments, but do convert them (for now)
+            my $referer = DW::Request->get->header_in("Referer");
+            warn "[#1742] lowercasing $k, url: $url (referer: $referer)\n"
```
zorkian: I'm pretty sure this will be very, very spammy? Maybe only print this if it actually lowercased something?

kareila: ... That's what it's doing, due to all the 'nexts' just above.

zorkian: Ah oops. Yeah, I personally find code style of that form (lines almost look identical) to lead to line blindness and then I missed one. At my dayjob, we have a style guide preferring the use of Compounds Ifs like:

```next if $k eq lc $k || $k =~ /^Widget\b/ || $k eq 'getProperties';```

PR #1770, opened 2016-05-11
Fix #1768

### LINKED ISSUE #1768: dw-free shouldn't default to using en_DW for translation
... since the en_DW data is in dw-nonfree.

LJ::Global::Defaults does the right thing: `@LANGS = ("en") unless @LANGS;`

However etc/config.pl says: `@LANGS = qw( en_DW );`

One possible solution could be to put this config line in a conditional that only executes if `-d $HOME/ext/dw-nonfree` is true.

pinterface:
Fix for #1768, of a sort.


kareila:
I am content with this solution for now.  Thanks for working on it!

PR #1771, opened 2016-05-12
Add url and url_full methods to LJ::Faq

### LINKED ISSUE #1722: Add url_full method to LJ::Faq
Currently when including a FAQ link somewhere (as in #1717, for LJ::Support) you end up having to write something like

```
# now actually get the FAQ
my $faq = LJ::Faq->load( $faqid, lang => $lang );
if ( $faq ) {
    $faq->render_in_place;
    my $faqref = $faq->question_raw . " " . "$LJ::SITEROOT/support/faqbrowse?faqid=$faqid&view=full";
}

... which is a bit protracted. For readabilty & clean-up purposes it would be nice if instead one could use LJ::Faq->url_full!

pinterface:
These are callable both on FAQ instances ($faq->url), and directly
(LJ::Faq->url 123), so you can easily produce a FAQ URL regardless
of whether you have a Faq object, or just the id.

Closes #1722.

### DIFF
--- a/cgi-bin/LJ/Faq.pm
+++ b/cgi-bin/LJ/Faq.pm
@@ -231,6 +231,18 @@ sub sortorder {
     return $self->{sortorder};
 }
 
+sub url {
+    my ($class, $faqid) = @_;
+    $faqid = $class->{faqid} if ref $class;
+    return "$LJ::SITEROOT/support/faqbrowse?faqid=$faqid";
+}
+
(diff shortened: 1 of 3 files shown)


kaberett:
Thank you for getting this one! I am super excited to have it available :-)

zorkian:
Nice! Thank you for writing tests, too. Awesome :)

PR #1773, opened 2016-05-12
Adjust things to avoid Perl 5.20's precedence warnings

### LINKED ISSUE #1772: Possible precedence issues (complaints by Perl 5.20)
These appear in the apache logs on startup.

```
Possible precedence issue with control flow operator at /home/dw/dw/cgi-bin/LJ/Stats.pm line 66.
Possible precedence issue with control flow operator at /home/dw/dw/cgi-bin/LJ/Stats.pm line 113.
Possible precedence issue with control flow operator at /home/dw/dw/cgi-bin/DW/Hooks/SiteSearch.pm line 25.
Possible precedence issue with control flow operator at /home/dw/dw/cgi-bin/DW/Shop.pm line 191.
```

For instance, in Stats.pm, around line 66:
```perl
    return LJ::Stats::get_db("dbr")
        or die "Can't get db reader handle.";
```
Which, by Perl's rules of precedence, is equivalent to
`(return f) or (die "msg")`, which is pretty nonsensical.
Presumably, this should be something more like `return f || die`,
to match the ordering of `return (f or die)`.

see http://perldoc.perl.org/perldiag.html#Possible-precedence-issue-with-control-flow-operator

pinterface:
Due to Perl's precedence rules, `return x or thing` is always
interpreted as `return x`, so the `or thing` never happens.
That's not really ideal when you're trying to error out (`or die`),
so adjust these calls to ensure the dieing is not dead code.

Fixes #1772

### DIFF
--- a/cgi-bin/DW/Hooks/SiteSearch.pm
+++ b/cgi-bin/DW/Hooks/SiteSearch.pm
@@ -21,8 +21,9 @@ use LJ::Hooks;
 
 sub _sphinx_db {
     # ensure we can talk to our system
-    return LJ::get_dbh( 'sphinx_search' )
+    my $dbh = LJ::get_dbh( 'sphinx_search' )
         or die "Unable to get sphinx_search database handle.\n";
+    return $dbh;
 }
 
 LJ::Hooks::register_hook( 'setprop', sub {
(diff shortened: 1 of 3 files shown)


kareila:
Hurray first commit! 🎉

rahaeli:
DANCE PARTY

PR #1776, opened 2016-05-14
Encode plaintext e-mails with quoted-printable

### LINKED ISSUE #1501: Unicode characters in plaintext email notifications coming through as WTF-8
This may be related to switching to SES for mail sending, or it might not be. (In other words, it might not be reproducable except in production; I haven't tried.) It did just start recently, though.

Text entered in UTF-8 is coming through in my (plaintext) email notifs as WTF-8. For instance, the following sentence in Russian:

Хва́стать, не коси́ть, спина́ не боли́т.

came through as:

Ð¥Ð²Ð°ÌÑÑÐ°ÑÑ, Ð½Ðµ ÐºÐ¾ÑÐ¸ÌÑÑ, ÑÐ¿Ð¸Ð½Ð°Ì Ð½Ðµ Ð±Ð¾Ð»Ð¸ÌÑ.

Text displays properly on website and in HTML emails. 

I'm not sure if this is worth spending too much time on, since encoding errors are enough to make strong people weep and it *does* work in HTML emails, but since it used to work and now it doesn't, I figured I'd file a bug.

pinterface:
HTML e-mails apparently work okay, but plaintext e-mails are full of
mojibake, but only in production, which runs through SES (that is,
sending e-mails via my local SMTP server works fine!).  The primary
difference between the two is that the HTML e-mails encode their
contents via quoted-printable, and the plaintext e-mails send raw,
unencoded 8-bit data.  SMTP and 8-bit data do not get along, and I
strongly suspect SES is re-encoding the data into 7-bit, poorly.

If what I surmise is correct, this should fix #1501.

### DIFF
--- a/cgi-bin/LJ/Sendmail.pm
+++ b/cgi-bin/LJ/Sendmail.pm
@@ -147,7 +147,8 @@ sub send_mail
                                    'Bcc'     => $opt->{bcc} || '',
                                    'Subject' => $subject,
                                    'Type'    => 'text/plain',
-                                   'Data'    => $body);
+                                   'Data'    => $body,
+                                   'Encoding' => 'quoted-printable');
 
             $msg->attr("content-type.charset" => $charset);
         }


rahaeli:
Oh wow, if that is the fix, you get the Debug Award.

zorkian:
I can't think of a reason _not_ to do this, and if it fixes it, yay! Let's go for it.

PR #1779, opened 2016-05-14
Comment on config file to update username as well as password

cesy:
This will make it easier for the next person who makes the same
mistake as me and forgets to update it.

### DIFF
--- a/etc/config-private.pl
+++ b/etc/config-private.pl
@@ -20,7 +20,7 @@
                'master' => {  # master must be named 'master'
                    'host' => "localhost",
                    'port' => 3306,
-                   'user' => 'dw',
+                   'user' => 'dw',    #CHANGETHIS if on Dreamhack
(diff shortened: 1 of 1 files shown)


etc/config-private.pl
```diff
@@ -20,7 +20,7 @@
                'master' => {  # master must be named 'master'
                    'host' => "localhost",
                    'port' => 3306,
-                   'user' => 'dw',
+                   'user' => 'dw',    #CHANGETHIS if on Dreamhack
                    'pass' => 'password',    # CHANGETHIS
                    'dbname' => 'dw',
```
kareila: Might as well comment dbname as well, if you want to make a note to change all the things that need to be changed. Pretty sure all the database names on the Dreamhack server are of the form "dreamhack_[username]".

cesy: Good point - updated them all now.

Issue #1719, opened 2016-03-23
add email comparison function


rahaeli:
I originally thought (in the comments to issue #1668) that we didn't have ANY email check/comparison functions, but turns out we do have some (in cgi-bin/LJ/User/Message.pm), I just couldn't remember where they were. :)

Still, we don't have a function that accepts two accounts and returns yes if they have the same email address and no if they don't. For error message purposes, we should probably distinguish "do these two accounts have the same email address" and "are the email addresses for both accounts confirmed", and the latter can probably be just a call to is_validated, but I think there's value in adding a function for the comparison.

(This comparison is used in renaming, and now in the bonus icon transfer that @kaberett is doing in #1668, and I think maybe somewhere else but my brain is mush right now.)

For readability purposes I think it's probably best done as a call on a particular user, with the user being compared to as the arg? Like, using #1668 as an example of the kind of comparisons we need, would people find this immediately readable?

> return $self->error( 'Email addresses do not match') 
> if $destination_u->has_same_email_address_as( $to_u );
> return $self->error( 'One or more email addresses are not confirmed' )
> unless $to_u->is_validated && $destination_u->is_validated;

kareila:
That seems legible and easy to remember, but I would suggest just calling it "has_same_email_as" for consistency with other email-related methods.

rahaeli:
oh, sure, that was just a quick off-the-top-of-my-head name. :)

chrisboyle:
claim

chrisboyle:
The only other place I can find (having grepped for e.g. `'mail.* eq .*mail'`)  is in `LJ/Poll.pm` to prevent two users with the same email voting in the same poll.

rahaeli:
lol I didn't know that code still existed in there. (It was only used for some LJ 'official' polls like voting for the advisory board.)

PR #1782, opened 2016-05-14
Add $user->has_same_email_as( $user2 ).

chrisboyle:
Also introduce consistency in that this always compares case-insensitively where some of these (bonus icons and a community-admin test case) were previously case-sensitive for no reason that I can see.

Fixes: #1719


zorkian:
Thank you!

Issue #1287, opened 2015-03-07
make community moderated entry accept/reject message box auto-scale


rahaeli:
How hard would it be to make the text box you get when accepting or rejecting an entry from the moderation queue should auto-resize to fit the typing? (I have an extension that will let me grab-and-resize textareas, but it would be a lot nicer to have the window fit the text, much like the very box I am typing in right now does: start with a reasonable size, then expand if the person typing in it, like me, tends to run off at the keyboard.)

afuna:
Not super hard -- or at least we already have a jquery plugin that does that (on the new update page I think?) so it would be a matter of applying to the right element.

rahaeli:
Awesome, let's do that then :)

chrisboyle:
The plugin seems to be `./ext/dw-nonfree/htdocs/js/jquery/jquery.autogrow-textarea.js` and is used by `/site/suggest`. The top of the plugin's JS file says MIT licensed, so its presence in dw-nonfree confuses me. Anyone ( @kareila ) know why? Perhaps if they meant the no-promotion-clause version of MIT (not what's *usually* meant these days afaik)? Otherwise it should be GPL-compatible (but I'm not a lawyer).

I mean I guess the auto-grow could be conditional on presence of non-free, but if the plugin can be moved, that seems simpler.

The plugin was added in https://github.com/dreamwidth/dw-nonfree/commit/aab5f36137534eedd8df4b4b1d80c70c987aad7b mentioning "bug 2566" which I don't know if there's a copy of ( @kaberett ?)

kareila:
Bug 2566 was "Auto expanding text boxes on Suggest An Improvement" and I think the only reason that plugin was added to dw-nonfree is because it was only used for the Site Suggestions page, which is also in dw-nonfree.  I would be fine with having it moved to dw-free.

chrisboyle:
claim

PR #1786, opened 2016-05-15
Make entry moderation message box auto-scale.

chrisboyle:
Moves autogrow plugin in from non-free (MIT licensed, only in non-free
because only /site/suggest was using it; separate PR to follow to clean
up the other end of this move).

Fixes: #1287


zorkian:
SGTM!

PR #1789, opened 2016-05-16
remove backend support for unique-email polls

kareila:
D has expressed surprise at least once upon discovering that the backend code
for unique-email polls still exists. It was designed to prevent multiple users
with the same email address from voting in the same poll: "It was only ever used
for some LJ 'official' polls like voting for the advisory board."

If we don't want the code, we should remove it.

### DIFF
--- a/bin/upgrading/deadphrases.dat
+++ b/bin/upgrading/deadphrases.dat
@@ -1165,7 +1165,9 @@ general pingback.option.lj_only
 general pingback.option.open
 general pingback.public.comment.text
 general pingback.sourceuri.default_title
+general poll.error.alreadyvoted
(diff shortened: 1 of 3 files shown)


zorkian:
Death to code!

zorkian:
We have always been at war with deprecated code!

zorkian:
Code debt is peace!

rahaeli:
Heh, I was going to open up an issue for it myself, but got distracted!

I should probably go through and take a look at all the unused hooks to see what else we can kill.

kareila:
so efficient I fix bugs before they get reported  :nail_care:

PR #1791, opened 2016-05-18
more code machete in poll code

kareila:
Followup to #1789.

Removes more unused poll-related hooks, which has the side effect of removing everything related to the "createdate" poll property.

Now that "unique" and "createdate" have both been excised, we can also completely drop the pollprop tables and related code.  I verified that "SELECT COUNT(*) FROM pollprop2" returned zero rows in production; pollproplist2 was populated from proplists.dat.


zorkian:
boom! jen strikes again!

PR #1792, opened 2016-05-18
Embiggen ml_items.itcode

### LINKED ISSUE #1787: Some ML item codes are over-long
Seen in the apache logs:
> '/customize/options.bml.widget.navstripchooser.option.color.control_strip_bordercolor' exceeds maximum code length, truncating to 80 symbols at /home/dw/dw/cgi-bin/LJ/Lang.pm line 329.
> '/customize/options.bml.widget.navstripchooser.option.color.control_strip_linkcolor' exceeds maximum code length, truncating to 80 symbols at /home/dw/dw/cgi-bin/LJ/Lang.pm line 329.

Fortunately, they're the only such strings:
```
mysql> select itcode from ml_items where length(itcode) = 80;
+----------------------------------------------------------------------------------+
| itcode                                                                           |
+----------------------------------------------------------------------------------+
| /customize/options.bml.widget.navstripchooser.option.color.control_strip_bgcolor |
| /customize/options.bml.widget.navstripchooser.option.color.control_strip_borderc |
| /customize/options.bml.widget.navstripchooser.option.color.control_strip_fgcolor |
| /customize/options.bml.widget.navstripchooser.option.color.control_strip_linkcol |
+----------------------------------------------------------------------------------+
4 rows in set (0.00 sec)
```

bgcolor and fgcolor happen to fit exactly, but bordercolor and linkcolor overflow slightly.

Since they're still unique based on those first 80 characters it's not a practical problem, but probably something to consider.  It may make sense to widen itcode, or shorten those names.

And in case you're wondering, for the most part things are well within the size limit.  There are only a handful of items above 70 characters, the majority are under 50:
```
mysql> select ceil(length(itcode)/10)*10 len, count(*) cnt from ml_items group by len order by len;
+------+------+
| len  | cnt  |
+------+------+
|   10 |   29 |
|   20 |  524 |
|   30 | 2178 |
|   40 | 2746 |
|   50 | 1255 |
|   60 |  415 |
|   70 |  159 |
|   80 |   39 |
+------+------+
8 rows in set (0.01 sec)
```

pinterface:
And deadphrase the truncated codes, of course.

Possibly controversial: I've marked ml_items.itcode as being ASCII-only, under the theory that it's intended as an internal name and has not had due consideration for the prospect of containing anything other than ASCII.  It felt appropriate, but I'd be happy to yank it out if preferred.

Fixes #1787.

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -1203,7 +1203,7 @@ CREATE TABLE ml_items (
     dmid    TINYINT UNSIGNED NOT NULL,
     itid    MEDIUMINT UNSIGNED AUTO_INCREMENT NOT NULL,
     PRIMARY KEY (dmid, itid),
-    itcode  VARCHAR(80) NOT NULL,
+    itcode  VARCHAR(120) CHARACTER SET ascii NOT NULL,
     UNIQUE  (dmid, itcode),
     proofed TINYINT NOT NULL DEFAULT 0, -- boolean, really
     INDEX   (proofed),
@@ -4151,6 +4151,12 @@ EOF
                   "ALTER TABLE spamreports ".
                   "MODIFY ip VARCHAR(45)" );
     }
+
+    unless ( column_type( 'ml_items', 'itcode' ) =~ /120/ ) {
+        do_alter( 'ml_items',
+                  "ALTER TABLE ml_items MODIFY COLUMN itcode VARCHAR(120) CHARACTER SET ascii NOT NULL" );
+    }
+
 });
 
 
--- a/bin/upgrading/deadphrases.dat
+++ b/bin/upgrading/deadphrases.dat
@@ -1676,3 +1676,6 @@ contentflag.viewingconcepts.byposter
 contentflag.viewingexplicit.bycommunity
 contentflag.viewingexplicit.byjournal
 contentflag.viewingexplicit.byposter
+
+general /customize/options.bml.widget.navstripchooser.option.color.control_strip_borderc
+general /customize/options.bml.widget.navstripchooser.option.color.control_strip_linkcol
--- a/cgi-bin/LJ/Lang.pm
+++ b/cgi-bin/LJ/Lang.pm
@@ -16,7 +16,7 @@ use strict;
 use LJ::LangDatFile;
 
 
-use constant MAXIMUM_ITCODE_LENGTH => 80;
+use constant MAXIMUM_ITCODE_LENGTH => 120;
 
 my @day_short   = (qw[Sun Mon Tue Wed Thu Fri Sat]);
 my @day_long    = (qw[Sunday Monday Tuesday Wednesday Thursday Friday Saturday]);


rahaeli:
I don't think that should be an issue, but I'll defer to the experts!

cgi-bin/LJ/Widget/NavStripChooser.pm
```diff
@@ -36,7 +36,7 @@ sub render_body {
     $ret .= "<p>" . $class->ml('widget.navstripchooser.colors') . "</p>";
 
     # choose colors
-    my $color_selected = $u->prop( 'control_strip_color' ) ne ''
+    my $color_selected = ( $u->prop( 'control_strip_color' ) // '' ) ne ''
```
zorkian: This bit of code seems unrelated to the original fix, yeah? If so, it should be pulled out into a separate pull request so we don't cross the streams.

bin/upgrading/update-db-general.pl
```diff
@@ -1201,7 +1201,7 @@
     dmid    TINYINT UNSIGNED NOT NULL,
     itid    MEDIUMINT UNSIGNED AUTO_INCREMENT NOT NULL,
     PRIMARY KEY (dmid, itid),
-    itcode  VARCHAR(80) NOT NULL,
+    itcode  VARCHAR(120) CHARACTER SET ascii NOT NULL,
```
zorkian: I'm actually unfamiliar with what this particular syntax does. What behavior does this impact for the DB?

pinterface:
To some extent, `CHARACTER SET ascii` is more about what it won't do.

As a quick overview: MySQL keeps track of both the character set data is stored in, and the character set the client is communicating with.  If they agree, data gets shipped around as-is.  If they disagree--say, if storage is marked as latin1 and the client is communicating in utf-8--, MySQL will helpfully convert the data.

Which is great, /if/ the data is accurately marked.  If a column is marked as, say, latin1, and actually stored as, say, utf-8, data gets re-encoded and you get mojibake.

One way to see this in action would be to query for an entry known to have unicode, and observe the difference in query output after `set names 'utf8';` and `set names 'latin1';`.  (This sort of charset disagreement is one of the issues you have with Sphinx, by the way.)

So coming back around to `CHARACTER SET ascii`, that means the field can only contain ASCII characters.  Under MySQL's default mode, that means MySQL will issue a warning and replace non-ascii characters with question marks.  Under `sql_mode = 'traditional'`, MySQL will actually error (not that I'd attempt traditional mode in production, given the age of the codebase).  The intent being to break loudly and quickly, rather than subtly and later, in the event of weirdness.

pinterface:
And you're right about the undef-cleanup bit.  Removed.  And hey, kareila also fixed that, and it's merged in already, so less work for me! :)

zorkian:
Right, that makes sense. I guess my followup question then is that is this appropriate for the specific column or can/should we set this somewhere else? I.e., it seems like a lot of work to reason this through for every column, vs just setting it at some global level and ensuring that everybody follows the same plan. I.e., utf8 everywhere or something.

I think that in this particular case it makes sense to use ASCII, though, so I'll merge. I'd be interesting to hear if there's some big win we could get on the overall problem tho or if it's just intractable at that level.

pinterface:
Utf-8 everywhere would definitely be the ideal[1].

From the MySQL side, the conversion is ugly, but relatively straightforward.  See [2] or [3], but basically it involves converting CHAR/TEXT->BINARY/BLOB->CHAR/TEXT.  Which, as you might imagine, is unlikely to be pleasant in production.

From a code perspective, I'd be very hesitant.  Utf-8 works great if all you ever do is ship it around as dumb bytes (embedded into latin1 generally works well enough, too).  It also works great if the entire stack is charset-aware and knows it's utf-8.  Mixed environments, on the other hand--every transition point is a possible bug, and in an old, large codebase there can be a lot of transition points.  DW is in theory all utf-8 /anyway/, so it's entirely possible that it wouldn't be nearly as painful as it might otherwise be, but it's still unlikely to be a "flip this one switch and never think about it again" sort of solution.

Worth experimenting with on a dev environment, though!  Converting the test suite databases to utf8mb4 and seeing how well the test suite handles things would hopefully give a good first approximation as to how much effort would be required.

[1] In MySQL parlance, that'd be CHARSET utf8mb4, because their first try (CHARSET utf8) doesn't allow codepoints outside the BMP.  Yeah. 
[2] https://www.percona.com/blog/2013/10/16/utf8-data-on-latin1-tables-converting-to-utf8-without-downtime-or-double-encoding/
[3] https://github.com/nicjansma/mysql-convert-latin1-to-utf8

rahaeli:
The issue with UTF8-everywhere is that a) the code does awful things with text encodings in some places because the LJ codebase dates back from the days before Perl had proper UTF8 handling and so we had to roll our own, and b) we have to maintain compatability with a large data set that was stored under that roll-your-own processing -- for purposes of user data, we have to remember that we're not dealing with data from just the 2008-2016 time frame that DW has existed, but the whole 1999-2016 time frame that LJ existed. (Due to imports.) I wouldn't want to rely just on test suites/dev environments for a safety check without finding someone whose journal data dates from those early days and checking against it.

(Theoretically, anything imported from LJ should have been through LJ's old UTF8 conversion process, but we all know what happens when theory meets practice -- and LJ's old UTF8 conversion process was not always great at Just Doing The Right Thing.)

azurelunatic:
(in other words, trust that we will find the *one* LJ user who hasn't used
LJ since ~2003 who decides to go ahead and move their lost-but-now-found
journal over to DW without having done the convert yet.)

pinterface:
I may have accidentally edited out too much of my apprehension in my
previous comment.  To clarify: the thought of switching to
utf8-everywhere for realsies is *terrifying*.

Even assuming everything went smoothly, it'd be a lot of effort for very
little gain--the existing SvUTF8 whack-a-mole works pretty okayish.

(The only reason I know any of this crap is because my current employer
has/had the same issue of utf-8 in latin1 columns, including utilizing
Sphinx and so having search on non-ascii characters buggered because of
the charset differences.  We solved our sphinx-related issues by
switching to Elasticsearch, but now I'm getting way off the topic.)

rahaeli:
Huh. I don't think we're really *wedded* to Sphinx; does Elasticsearch handle that sort of stuff better? It would be really neat to fix the thing where people writing, eg, in Cyrillic characters can't search their journals...

zorkian:
Yeah IMO these days Elasticsearch is way better an option, but it's
always :effort: to switch these things. Back when I did search it
was really not a viable option and I didn't want to do Lucene stuff
on my own.
 
--
Mark Smith
mark@qq.is
 
 
On Mon, May 23, 2016, at 10:43 PM, Denise Paolucci wrote:
> Huh. I don't think we're really *wedded* to Sphinx; does Elasticsearch
> handle that sort of stuff better? It would be really neat to fix the
> thing where people writing, eg, in Cyrillic characters can't search
> their journals...
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub[1]
>
 

Links:

  1. https://github.com/dreamwidth/dw-free/pull/1792#issuecomment-221171895

pinterface:
Well, I can't speak to cyrillic specifically, but we had a lot of things
get better through switching.

Indexing is less painful (shove JSON at it via workers to add documents
and get near-realtime indexing, no big multi-table joins or warehouse
tables necessary), document ids can be arbitrary values rather than a
single 64-bit int (so no need to try to keep a mapping of ids, just make
the id "$userID:$docID" or whatever), searches can return the full data
so there's no need to immediately turn around and ask the database for
the actual data, and it has a lot of things I would have needed to build
on top of sphinx (fuzzy indices for better matching in the face of bad
spelling, etc.).  Our unicode test cases got better, too, though in our
case, it was more along the lines of "cafe" finding "café" (which sphinx
understood as "caf"), which is a far cry from Cyrillic.

But like zorkian says, it does take work to switch--it's essentially
rebuilding the search infrastructure.  Took me a couple of months at the
dayjob, but I don't know DW's search facilities well enough to say how
they would compare in complexity.

PR #1795, opened 2016-05-20
Cleanup 2016q2

kareila:
Details in individual commit messages.  The usual, plus removing an unused userprop, rewriting a faulty conditional in CleanHTML.pm, and fixing some calls to DW::FormErrors->add_string that had one argument instead of two.

### DIFF
--- a/cgi-bin/LJ/Time.pm
+++ b/cgi-bin/LJ/Time.pm
@@ -29,6 +29,8 @@ use Time::Local ();
 sub days_in_month
 {
     my ($month, $year) = @_;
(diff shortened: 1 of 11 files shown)


cgi-bin/LJ/Time.pm
```diff
@@ -29,6 +29,8 @@ use Time::Local ();
 sub days_in_month
 {
     my ($month, $year) = @_;
+    return unless $month;  # not a mind reader
```
zorkian: lol comment

PR #1799, opened 2016-05-28
Update crazy eyes to use a slower version from rahaeli

### LINKED ISSUE #1798: accessibility issue with inherited LJ subject icons
http://www.dreamwidth.org/support/see_request?id=33421

We've received a report from a user having "neurological issues" caused by the rapid flashing of the side-to-side pair of eyes (htdocs/img/talk/sm10_eyes.gif), which are located in the optional subject icons on the comment form.

The user declined to bring up the issue in dw_suggestions or dw_accessibility because they don't have an account on the site, so I'm opening an issue here for discussion.

The two solutions that occur to me are to remove the animation entirely, or to slow the rate of refresh to something that isn't triggering.  Thoughts?

pinterface:
Fixes #1798.


kareila:
Has this been updated to use the most recent animation from the issue discussion?

pinterface:
Yup!  That was done in a1d79b451.

kareila:
Thanks!  Just wanted to make sure this had the agreed-upon version since we went through a few iterations.

PR #1801, opened 2016-06-02
Make alttext and titletext on /icons consistent with other uses of icons

### LINKED ISSUE #1703: icon page includes (Default) in alt text for ALL icons
Reported here: http://www.dreamwidth.org/support/see_request?id=32773

Verified on the reporting user's icons page which uses journal style, and on my own account's icons page which uses site style.

pinterface:
lightgetsin in dw-accessibility made an excellent point that it's useful to have icons use the same format on the /icons page as everywhere else, so the redundant information can help guide people to understand how the alt and title texts are formatted elsewhere.

So... this makes it use largely the same format, and practically by accident fixes #1703.  Unlike comment/entry icons, the default marker is included alongside a userpic's keywords.  Also unlike comment/entry icons, the "Upload Order" view includes all of an icon's keywords rather than only one.

As a bonus commit, fixes double-encoding of alt and title attributes.

### DIFF
--- a/cgi-bin/LJ/S2/IconsPage.pm
+++ b/cgi-bin/LJ/S2/IconsPage.pm
@@ -116,10 +116,12 @@ sub IconsPage {
             });
         }
 
+        my $kwstr = join( ', ', @{$keywords} );
+
         push @pics_out, {
             '_type' => 'Icon',
             id => $pic->picid,
-            image => Image( $pic->url, $pic->width, $pic->height, LJ::ehtml( $pic->alttext ), title => LJ::ehtml( $pic->keywords ) ),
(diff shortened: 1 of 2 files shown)


azurelunatic:
"practically by accident" means fixed-by-proxy, which is one of my favorite
ticket resolutions!

rahaeli:
My personal favorite is RESOLVED/ITJUSTSTARTEDWORKINGOUTOFNOWHERE.

kareila:
This looks fantastic.  Thank you!

emnem:
"practically by accident"... This makes me smile (maybe because I was the bug reporter on this). Awesome to see it get fixed.

PR #1803, opened 2016-06-03
Fix #1599 consistency when joining moderated-membership comm

### LINKED ISSUE #1599: Clarity & consistency when joining moderated-membership communities
If you have been invited to join a moderated-membership community, the invitation shows up at http://www.dreamwidth.org/manage/invites with the options Accept and Decline.

If you hit "join community" on the hover information (from hovering over a userhead), after you have been invited, nothing obvious happens. 

It seems to me that either you should get an error message at this point <i>or</i> hitting "join community" on the hover info (sorry, proper terminology has gone missing) should be equivalent to hitting "accept" on the invitations management page.

cesy:
Contextual hover now shows accept instead of join if you have
a pending invitation. See #1599 for details.

### DIFF
--- a/htdocs/js/contextualhover.js
+++ b/htdocs/js/contextualhover.js
@@ -333,14 +333,18 @@ ContextualPopup.renderPopup = function (ctxPopupId) {
             if (!data.is_closed_membership || data.is_member) {
                 var membershipLink  = document.createElement("a");
 
-                var membership_action = data.is_member ? "leave" : "join";
-
                 if (data.is_member) {
                     membershipLink.href = data.url_leavecomm;
                     membershipLink.innerHTML = "Leave";
+                    var membership_action = "leave"
+                } else if (data.is_invited) {
+                    membershipLink.href = data.url_acceptinvite;
+                    membershipLink.innerHTML = "Accept invitation";
+                    var membership_action = "accept"
                 } else {
                     membershipLink.href = data.url_joincomm;
                     membershipLink.innerHTML = "Join community";
+                    var membership_action = "join";
                 }
 
                 if (!ContextualPopup.disableAJAX) {
--- a/htdocs/js/contextualhover.js
+++ b/htdocs/js/contextualhover.js
@@ -333,18 +333,20 @@ ContextualPopup.renderPopup = function (ctxPopupId) {
             if (!data.is_closed_membership || data.is_member) {
                 var membershipLink  = document.createElement("a");
 
+                var membership_action;
+
                 if (data.is_member) {
                     membershipLink.href = data.url_leavecomm;
                     membershipLink.innerHTML = "Leave";
-                    var membership_action = "leave"
+                    membership_action = "leave";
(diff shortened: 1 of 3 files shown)


htdocs/js/contextualhover.js
```diff
@@ -333,14 +333,18 @@ ContextualPopup.renderPopup = function (ctxPopupId) {
             if (!data.is_closed_membership || data.is_member) {
                 var membershipLink  = document.createElement("a");
 
-                var membership_action = data.is_member ? "leave" : "join";
-
                 if (data.is_member) {
                     membershipLink.href = data.url_leavecomm;
                     membershipLink.innerHTML = "Leave";
+                    var membership_action = "leave"
```
kareila: I do see one possible issue here, which is that you've moved the variable scope of membership_action (the "var" declaration) into the conditional, which may mean (if the rules for JS are the same as for Perl) that the assignment is lost outside of that conditional.  I would keep the scope declaration above:

 `var membership_action;`

And then assign to the variable in each conditional without the var qualifier:

`membership_action = "leave"`
`membership_action = "accept"`
`membership_action = "join"`

htdocs/js/jquery.contextualhover.js
```diff
@@ -285,6 +285,8 @@ _renderPopup: function() {
         if ( ! data.is_closed_membership || data.is_member ) {
             if ( data.is_member )
                 this._addAction( data.url_leavecomm, "Leave", "leave" );
+            else if ( data.is_invited )
+                this._addAction( data.url_acceptinvite, "Accept invitation", "accept");
```
kareila: Every time I look at jQuery code, I'm impressed by how much simpler it is to read and write.  I really need to learn more about it. :)

htdocs/js/contextualhover.js
cesy: Okay, fixed.

htdocs/js/contextualhover.js
```diff
@@ -329,18 +329,23 @@ ContextualPopup.renderPopup = function (ctxPopupId) {
         // member of community
         if (data.is_logged_in && data.is_comm) {
             var membership      = document.createElement("span");
+            var membership_action;
 
             if (!data.is_closed_membership || data.is_member) {
                 var membershipLink  = document.createElement("a");
 
-                var membership_action = data.is_member ? "leave" : "join";
```
kareila: Let's put the `var membership_action` statement back here where it was previously, just to keep the scope the same.

kareila:
I did discover one edge case when testing this: if you accept the invitation, then click again to leave the community, it goes back to "Accept Invitation" instead of "Join Community", and if you click a third time, nothing happens.  Not sure if it's worth the effort to correct, since it's not likely that someone will want to do that on purpose, but it might get double-clicked accidentally.

htdocs/js/contextualhover.js
cesy: Okay, done.

PR #1806, opened 2016-06-07
[#1804] increase icon size limit from 40k to 60k

### LINKED ISSUE #1804: raise icon file size limit to 60K
Icon limits currently: 100x100px, 40K file size. After some internal discussion, we think we can raise that to 60K for file size (disk is cheaper than it was in 1999!)

Keep the 100x100px limit, though, since like everything on the site is visually designed around that constraint.

kareila:
This is a one line change accompanied by some related cleanup.

Once this is live, http://www.dreamwidth.org/support/faqbrowse?faqid=140 will need to be updated.

Fixes #1804.

### DIFF
--- a/bin/upgrading/deadphrases.dat
+++ b/bin/upgrading/deadphrases.dat
@@ -50,6 +50,12 @@ general /editinfo.bml.schools.title
 general /editinfo.bml.topicdir.about
 general /editinfo.bml.topicdir.header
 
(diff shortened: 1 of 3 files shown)


rahaeli:
I think the reason it's not a config option is because nobody really expected it to ever change...

zorkian:
Yeah.

kaberett:
FAQ edited!

Issue #1808, opened 2016-06-13
ESN Inbox expand/collapse button is not keyboard accessible


pinterface:
A user complained about this in http://www.dreamwidth.org/support/see_request?id=29428

I don't have JAWS, but I can confirm that the button works via mouse, but does not work via keyboard.  I suspect they're running into that, and not a JAWS-specific issue.

Technically, it's also a link, but the link doesn't do anything useful (it reloads the page, but doesn't change anything).  Only users running without JS (hi!) are likely to notice that aspect, though.

rahaeli:
Oh wow, embarrassed that I missed the user had replied back to that one so long ago. Eep!

If you want to run it through actual JAWS testing, @deborahgu has it installed and has gotten pretty good at testing with it. But I agree that this is likely to be the culprit.

kareila:
Hey @pinterface you can decide if you want to leave this open for the non-JS case and if so, if you want to keep it claimed or leave it for someone else.

pinterface:
I think it makes sense to close this.  Yes, having the Inbox work without JS would be nice, but there are already two open bugs about redoing it entirely (#1094 and #1224), so it seems silly to expend very much effort on the current design.

kareila:
Fair enough :)

PR #1809, opened 2016-06-13
Move the inbox's expando activation event to the link

pinterface:
The link is what makes the expando button selectable from the keyboard, but the event was on the img tag.  That works fine for clicking, but results in keyboard-activation following the link instead of firing the desired event.

This is a partial fix for bug #1808.  It fixes keyboard activation if JS is enabled, but does not attempt to deal with the no-JS case.


kareila:
I'm sure this is far from the only problem with using the inbox without JS. Thanks for fixing the most common case!

PR #1810, opened 2016-06-13
remove vcard-related pages and code

kareila:
I don't see any evidence that these are in use.  Correct me if I'm wrong...


zorkian:
If there is one or two users, I'm still fine removing them.

rahaeli:
I don't even remember what vcards are...

azurelunatic:
Weren't they for ads? V stood for "vertical" if I recall correctly.

kareila:
Heh, no. They are a standard XML-based format for communicating a set of contact information suitable for importing into a contact management program such as Outlook or Apple's Address Book.

PR #1811, opened 2016-06-16
Add LJ::dhtml to mirror LJ::ehtml

### LINKED ISSUE #1737: Should we add an LJ::dhtml function?
When I was fixing #1643 just now, I thought it was odd that we didn't have an LJ::dhtml function for undoing LJ::ehtml, similar to LJ::durl and LJ::eurl.  Current practice is to use HTML::Entities::decode (or decode_entities - same function, different name).

I think it might be a good idea to add this, (a) for symmetry and ease of remembering, and (b) so that if we ever want to use a different method of decoding in the future, we only need to change it in one place.

kaberett:
Fixes #1737.

Adds LJ::dhtml for symmetry with LJ::ehtml, which applies
HTML::Entities::decode (or equivalently HTML::Entities::decode_entities)
to an input string. Replaces HTML::Entities::decode(_entities) with
LJ::dhtml in the codebase.

---

I have a couple of questions about preferred implementation...

Based on http://search.cpan.org/dist/HTML-Parser/lib/HTML/Entities.pmv I think using decode_entities() rather than decode() is Better Practice, so that's what I've done.

Comparing with LJ::ehtml, would it be preferable to add in
`return $a unless $a =~ /[&]/;`
... (or regex that actually definitely works as opposed to Alex being sort of perplexed and suspicious) for the common case of "contains no HTML entities"? Because any HTML entity will be of the form `&entity;`?

ETA: replying to comments by e-mail requires a paid/seed account on Dreamhacks (but not in production) -- is this intentional?

Behaviour is appropriate for comments that are e-mailed in (http://kaberett.kaberett.hack.dreamwidth.net/20617.html#comments). Ditto posts that are e-mailed in (http://kaberett.kaberett.hack.dreamwidth.net/20746.html).

### DIFF
--- a/cgi-bin/DW/EmailPost/Entry.pm
+++ b/cgi-bin/DW/EmailPost/Entry.pm
@@ -30,7 +30,6 @@ use strict;
 use LJ::Protocol;
 
 use Date::Parse;
-use HTML::Entities;
 use IO::Handle;
 use XML::Simple;
 
@@ -513,7 +512,7 @@ sub _cleanup_mobile_carriers {
         return $self->err( "Unable to find XML content in PictureMail message." )
             unless $xml_string;
 
-        HTML::Entities::decode_entities( $xml_string );
+        LJ::dhtml( $xml_string );
         my $xml = eval { XML::Simple::XMLin( $xml_string ); };
         return $self->err( "Unable to parse XML content in PictureMail message." )
             if ! $xml || $@;
@@ -522,8 +521,7 @@ sub _cleanup_mobile_carriers {
             unless $xml->{messageContents}->{type} eq 'PICTURE';
 
         my $url =
-          HTML::Entities::decode_entities(
-            $xml->{messageContents}->{mediaItems}->{mediaItem}->{url} );
+          LJ::dhtml( $xml->{messageContents}->{mediaItems}->{mediaItem}->{url} );
         $url = LJ::trim($url);
         $url =~ s#</?url>##g;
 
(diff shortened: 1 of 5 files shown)


kareila:
> Comparing with LJ::ehtml, would it be preferable to add in
> `return $a unless $a =~ /[&]/;`

I think this would be premature optimization.  In the case of ehtml, we are trying to avoid a sequence of potentially time-consuming regular expression substitutions that would have no effect, but for dhtml, you can just pass the whole thing off to decode_entities and assume it will operate efficiently.  (It appears to be coded as an XS module, so it will certainly be faster than the pure-Perl equivalent.)

> replying to comments by e-mail requires a paid/seed account on Dreamhacks (but not in production) -- is this intentional?

Not as far as I know; can you open that as a new issue for investigation?

rahaeli:
> > replying to comments by e-mail requires a paid/seed account on Dreamhacks (but not 
> in production) -- is this intentional?
> 
> Not as far as I know; can you open that as a new issue for investigation?


Probably a cap set differently in the default 'hack config than on prod.

kareila:
Possibly, although if so I would have thought it would have been caught when I did #1576.

cgi-bin/DW/EmailPost/Entry.pm
```diff
@@ -513,7 +512,7 @@ sub _cleanup_mobile_carriers {
         return $self->err( "Unable to find XML content in PictureMail message." )
             unless $xml_string;
 
-        HTML::Entities::decode_entities( $xml_string );
+        LJ::dhtml( $xml_string );
```
kareila: Can you add a comment here to indicate that $xml_string is being modified by the function call?  It's documented in the POD for HTML::Entities, but it's unusual for a function to behave this way in Perl.

cgi-bin/DW/Worker/ContentImporter/LiveJournal/Userpics.pm
```diff
@@ -157,7 +156,7 @@ sub get_lj_userpic_data {
     my $cleanup_string = sub {
         # FIXME: If LJ ever fixes their /data/userpics feed to double-escepe, this will cause issues.
```
kareila: Might be nice to fix the "escepe" typo while we're in here.

kaberett:
Done! I think. Except I dithered over how best to lay out the first requested comment, so I'm entirely willing to be told to go away and amend my horrible postfix-commenting ways ;)

kaberett:
Also hurrah re efficient operation of decode_entities :-)

PR #1812, opened 2016-06-17
free accounts can post entries and comments via email

kareila:
As pointed out in #1811, this is allowed in production, so let's turn it on in the default config.pl so that it works on dreamhacks as well.

### DIFF
--- a/etc/config.pl
+++ b/etc/config.pl
@@ -419,7 +419,7 @@
             'checkfriends_interval' => 0,
             'directory' => 1,
             'edit_comments' => 0,
-            'emailpost' => 0,
+            'emailpost' => 1,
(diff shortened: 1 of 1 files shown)


kaberett:
Free accounts can't post entries via e-mail in production! That one IS a paid feature. But they can reply to comments by e-mail. Sorry for being confusing!

kareila:
You're not being confusing, the code is.  :)

It looks like cgi-bin/DW/EmailPost/Entry.pm and cgi-bin/DW/EmailPost/Comment.pm
 both rely on `$u->can_emailpost` to determine who can use them, and that ties directly back to the 'emailpost' cap, which is turned on for every account type in production.

PR #1813, opened 2016-06-17
Correctly render special characters in PM subjects

### LINKED ISSUE #1652: Punctuation marks in PM subjects are converted to HTML entities
e.g. (from a seven-reply-deep conversation): *Re: [...] just so I don&amp;amp;amp;amp;amp;amp;#39;t go more [...]*

This should probably not happen; compare with #1643.

kaberett:
Fixes #1652.

Depends on #1811 -- this makes use of the LJ::dhtml function introduced
in #1811 to ensure that special characters in PM subject lines are
preserved on reply, instead of ending up in an endlessly-nested sequence
of &amp;amp;amp;amp;amp;..., making use of the fact that special
characters do not need to be encoded as HTML entities in the compose
form.


kaberett:
Thank you @kareila for suggesting how to approach making the fix &lt;3

afuna:
I am slightly worried about this making us vulnerable to HTML attacks because we do the unescaping after cleaning. 

Eg have someone do &quot;&lt;script&gt;, that gets unescaped, and then suddenly we have a live script tag on the site

kaberett:
My understanding is that what I'm doing here unescapes text *in the reply form subject line only*. The reason that  a multiple-reply chain of PMs ended up with a subject line containing *don&amp;amp;amp;amp;amp;#39;t* is that the subject line is then re-escaped at the point the form's submitted (so when a message is first sent, ' gets escaped to &#39;, and on subsequent replies the ampersand gets escaped in turn). I *believe* (and I think Jen does too?) that what I'm doing here isn't any more unsafe than letting people write "<3" in a subject line in the first place?

afuna:
Hm actually I retract my objection. It gets unescaped here and then gets escaped again as part of html_text so this shouldn't open us up to further risk...

We do allow <3 etc in subjects but we run the subject text through an html cleaner before we print it on the page, or else run through an escape function as we do before throwing into html elements (as we do here)

pinterface:
Just a thought, but it might be less convoluted to use `$msg->subject_raw`, which doesn't escape the subject in the first place, rather than escaping, unescaping, and reescaping the subject.  Or maybe add a `$msg->subject_unescaped`, which drops the `ehtml()` but keeps the `|| "(no subject)"` (which at present is entirely redundant in compose.bml). 

Arguably, the fundamental problem is that `->subject` is doing escaping at all, since it encourages situations like this where things end up double-encoded, but figuring out whether it can be removed would involve analyzing every single caller and fixing said caller if necessary, and that's just no fun.

htdocs/inbox/compose.bml
```diff
@@ -203,7 +203,7 @@ body<=
         $reply_u = $msg->other_u;
         $reply_to = $reply_u->display_name;
         $disabled_to = 1;
-        $msg_subject = $msg->subject || "(no subject)";
+        $msg_subject = LJ::dhtml( $msg->subject ) || "(no subject)";
```
kareila: @pinterface is correct - if you look at LJ/Message.pm, the difference between `subject` and `subject_raw` is that the former is wrapping the latter in an ehtml call and adding the `|| "(no subject)"` text.  So the better change here would be to simply change `subject` to `subject_raw`, instead of escaping a thing and then immediately unescaping it.

Thanks @kaberett for working on this and to everyone else for their reviews and comments.  ❤️

rahaeli: Would that open us up to an injection attack, or is subject_raw already cleaned at that point?

kareila: As @afuna pointed out above, it's passed to `html_text` which escapes it again.

rahaeli: Ahh, okay, don't mind me then! CARRY ON

kareila: I think it's awesome that everyone is being SUPER EXTRA CAREFUL. 🐝 🐝 🐝

rahaeli: WE HAVE BEEN WELL TRAINED

kaberett:
Thank you all for the feedback and advice; change duly made :-)

Issue #1735, opened 2016-04-04
eliminate specialcasing for default messaging settings for under-18 accounts


rahaeli:
See #1733 -- if your account is marked under-18, only mutually trusted accounts can PM you by default. Fix that to be the same as all accounts, defaulting to "registered users".

Issue #1733, opened 2016-04-03
fix UserMessaging.pm to use method instead of raw prop


kareila:
http://www.dreamwidth.org/support/see_request?id=33001

If a user is a minor and hasn't changed their UserMessaging setting from the default, the Settings page will tell them that they can receive PMs from Registered Users, when in fact they are only allowed to receive PMs from Mutually Trusted Users.

Fix: on line 35 of LJ/Setting/UserMessaging.pm, change `$u->prop("opt_usermsg")` to `$u->opt_usermsg`.  This will use the correct logic to display the effective value of the setting.

rahaeli:
Let's just remove the specialcasing for under-18 accounts -- that has to be holdover from LJ.

kareila:
We can do that too, but using the method instead of the raw prop is still the correct approach.

rahaeli:
Opened 1735 for eliminating the special casing!

PR #1819, opened 2016-06-20
Issue 1733 / 1735 - Eliminate special casing for PMs to minor accounts

hotlevel4:
Changed minor accounts to receive PMs from all registered users.
Fixed call when checking whether a PM can be sent.

Fixes issue #1733. Fixes issue #1735.

### DIFF
--- a/cgi-bin/LJ/Setting/UserMessaging.pm
+++ b/cgi-bin/LJ/Setting/UserMessaging.pm
@@ -32,7 +32,7 @@ sub option {
     my ($class, $u, $errs, $args) = @_;
     my $key = $class->pkgkey;
 
-    my $usermsg = $class->get_arg($args, "usermsg") || $u->prop("opt_usermsg");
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/Setting/UserMessaging.pm
```diff
@@ -32,7 +32,7 @@ sub option {
     my ($class, $u, $errs, $args) = @_;
     my $key = $class->pkgkey;
 
-    my $usermsg = $class->get_arg($args, "usermsg") || $u->prop("opt_usermsg");
+    my $usermsg = $class->get_arg($args, "usermsg") || $u->opt_usermsg);
```
kareila: Oops! It looks like there is a stray close paren here that also needs to be removed.

kareila:
One trivial syntax fix before this can go in.  Thank you!

hotlevel4:
Argh, sorry, fixed.

kareila:
No worries :)

Issue #1820, opened 2016-06-20
extra padding on /entry/new


rahaeli:
After the code push, there's now extra padding on the security/location buttons on /entry/new:

![extrapadding](https://cloud.githubusercontent.com/assets/1716035/16188658/9895ce54-36a4-11e6-9b02-9684df473f34.png)

Previously, those drop-downs were closer to each other (and I think they still should be).

Happens in Tropo and Celerity; haven't tried it in anything else.

kareila:
Probably due to #1793.  I'll look into it.

kaberett:
(This also affects the tag box, and the bullet-pointed "you can view this entry/view your journal entries/etc" section on the success landing page.)

kareila:
... Okay!  Thanks for the additional info.

kareila:
I've updated the PR to address the tag widgets. The list on the success page doesn't currently belong to any CSS classes, so there's nothing to target an adjustment with.

PR #1793, opened 2016-05-19
add margins for Foundation list elements

kareila:
When working on dreamwidth/dw-nonfree#113, I discovered the lack
of marginal padding around various list elements made the pages
hard to read.  These are my suggested changes for improvements.

If this isn't the best place to put these definitions, let me
know of a preferable alternative.

PR #1821, opened 2016-06-20
[#1820] remove left margin for beta entry page buttons caused by #1793

kareila:
I'll go ahead and get this live if it gets the thumbs up.

### DIFF
--- a/htdocs/scss/pages/entry/new.scss
+++ b/htdocs/scss/pages/entry/new.scss
@@ -75,6 +75,7 @@
 
     li {
         display: inline-block;
+        margin-left: 0 !important;
 
         input, .button {
             border-radius: 0;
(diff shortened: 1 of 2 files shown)


kareila:
Oh, yeah: fixes #1820.

htdocs/scss/pages/entry/new.scss
```diff
@@ -75,6 +75,7 @@
 
     li {
         display: inline-block;
+        margin-left: 0 !important;
```
zorkian: My understanding of CSS is that `!important` is a hack that shouldn't be used, or is there legitimate reasons to use it?

kareila: It's used for things that should never be overridden by more specific targets, so it's appropriate here.  There are other lines in the same SCSS file that also use it.

zorkian:
Thanks for the explanation!

PR #1824, opened 2016-06-26
Remove userblob

zorkian:
These are the only references in the code. We track this stuff but don't
expose it anywhere or use it. Murderize it.


rahaeli:
"Userblob" sounds vaguely dirty.

/is twelve

PR #1829, opened 2016-06-27
Start BlobStore

zorkian:
This starts implementation of BlobStore which will replace all the
various storage layers we use for userpics/media. The idea is that we
should have one simple storage system and offload the work to other
people instead of doing it all ourself.

This is first part in removing MogileFS. It's a really complicated
system that is better replaced by using something simple with frontend
caching and/or a file storage service other people provide.

(Although the BlobStore interface was made with the idea that it can
support MogileFS pretty easily, so it's not actually hard... if we added
'namespace' to the API then it'd be a drop-in replacement.)


zorkian:
Tested MogileFS codepaths and they work on my dev server. Localdisk also works. Need to implement S3.

This diff also makes it so that BlobStore works in only-MogileFS mode, so basically it could be reviewed and landed ~now. We could deploy it to prod in this mode and it would just basically continue to work as it does today, except via the BlobStore API.

Then, when S3 is ready, we shim that in front and allow files to start writing to S3. We'll have to write up something that can transition files from MogileFS to S3. That's not hard, but will probably be a one-off script rather than some built in functionality.

cgi-bin/ljlib.pl
```diff
@@ -54,6 +54,14 @@ BEGIN
     }
 }
 
+# Now set up logging support for everybody else to access
+use Log::Log4perl;
```
kareila: All right. Looks like we will need to make sure Log::Log4perl gets installed before the code push, either in the dw user's extlib, or on each production system.  Otherwise the server won't even start, since you've added it to ljlib.pl.

bin/incoming-mail-inject.pl
```diff
@@ -40,55 +41,46 @@ ()
...
-
-        if ($len > IN_MEMORY_THRES && ! $ieh) {
-            if (should_ignore($msg)) {
-                $ignore_message = 1;
-                next;
-            }
-
-            # allocate a mogile filehandle once we cross the line of
-            # what's too big to store in memory and in a schwartz arg
-            $ieh = LJ::IncomingEmailHandle->new;
```
kareila: It seems to me that these edits in bin/incoming-mail-inject.pl remove all uses of LJ::IncomingEmailHandle.  Was that intentional, given that you also updated the module below? Or can we rip it out entirely?

bin/upgrading/update-db.pl
```diff
@@ -701,47 +701,6 @@ sub populate_moods {
     }
 }
 
-sub populate_mogile_conf {
```
kareila: Here you remove the definition of populate_mogile_conf but not the place earlier in the file where the function is called.

bin/upgrading/update-db.pl
```diff
@@ -25,7 +25,7 @@
 use Cwd qw/ abs_path /;
 use Image::Size ();
 use LJ::S2;
-use MogileFS::Admin;
+use DW::BlobStore;
```
kareila: I don't see any reason to include DW::BlobStore?

bin/upgrading/update-db.pl
```diff
@@ -701,47 +701,6 @@ sub populate_moods {
     }
 }
 
-sub populate_mogile_conf {
-    # create/update the MogileFS database if we use it
-    return unless defined $LJ::MOGILEFS_CONFIG{hosts};
```
kareila: Shouldn't we leave this as is, as long as we support the option of using MogileFS?  Maybe just make the inclusion of MogileFS::Admin also dependent on whether MOGILEFS_CONFIG is defined?

cgi-bin/LJ/Config.pm
```diff
@@ -58,17 +58,6 @@ sub reload {
...
-        # reload MogileFS config
-        if (LJ::mogclient()) {
-            LJ::mogclient()->reload
-                ( domain => $LJ::MOGILEFS_CONFIG{domain},
-                  root   => $LJ::MOGILEFS_CONFIG{root},
-                  hosts  => $LJ::MOGILEFS_CONFIG{hosts},
-                  readonly => $LJ::DISABLE_MEDIA_UPLOADS, );
-              LJ::mogclient()->set_pref_ip(\%LJ::MOGILEFS_PREF_IP)
-                  if %LJ::MOGILEFS_PREF_IP;
-          }
```
kareila: Similar to update-db.pl above - shouldn't we continue to include this, until MogileFS support is completely removed?

cgi-bin/ljlib.pl
```diff
@@ -242,30 +250,6 @@ sub gearman_client {
     return $client;
 }
 
-sub mogclient {
-    return $LJ::MogileFS if $LJ::MogileFS;
```
kareila: We can't remove LJ::mogclient - it's used in many, MANY other places that aren't touched here.

etc/workers.conf
```diff
@@ -120,4 +120,3 @@ reference-list:
     taglib-gm: 1            # loads tags using a worker (performance thing)
     talklib-gm: 1           # loads comments using a worker (performance thing)
     t-memlimit: 1           # for running certain tests
-    userpic-resize-gm: 1    # resizes userpics
```
kareila: Out of curiosity, what's the behavior if workers.conf includes a nonexistent worker?  Will it complain about trying to start a nonexistent worker, or silently fail?

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+    # to point to our run directory which will work for test environs.
+    if ( exists $LJ::BLOBSTORE{s3} ) {
+        $log->debug( 'Initializing S3 blobstore.' );
+        push @{$blobstores ||= []}, DW::BlobStore::S3->init( %{$LJ::BLOBSTORE{s3}} );
+    }
+    if ( exists $LJ::BLOBSTORE{localdisk} ) {
+        $log->logcroak( 'Must only define a single $LJ::BLOBSTORE entry.' )
+            if defined $blobstores;
+        $log->debug( 'Initializing localdisk blobstore.' );
+        push @{$blobstores ||= []}, DW::BlobStore::LocalDisk->init( %{$LJ::BLOBSTORE{localdisk}} );
```
kareila: It would be a good idea to add an example %LJ::BLOBSTORE to the config.  I don't see the format of the hash documented anywhere.  It looks like it's supposed to be a hash with a single key, the value of which is a hashref of a format that can't be determined without parsing through the submodule's code.

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+package DW::BlobStore;
+
+use strict;
+use v5.10;
+use Log::Log4perl;
+my $log = Log::Log4perl->get_logger( __PACKAGE__ );
+
+use DW::BlobStore::LocalDisk;
+use DW::BlobStore::MogileFS;
+use DW::BlobStore::S3;
```
kareila: You can use `LJ::ModuleLoader->require_subclasses('DW::BlobStore')` instead of listing each submodule separately. Besides being easier to read, it lets any future submodules be loaded automatically, without having to add them here.

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+
+my $blobstores;
+
+sub _get_blobstores {
+    # If we've already created one, simply return it.
+    return $blobstores if defined $blobstores;
+
+    # Site owners must configure one of these. Disk is configured by default
+    # to point to our run directory which will work for test environs.
```
kareila: I don't see where that default you mention here is set.  Looks like the LocalDisk init method will croak if a path isn't explicitly specified.

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+    # As a way to support migration of MogileFS data to the new storage
+    # system, we support a way of specifying a fallback MogileFS cluster which
+    # is activated if we ask for it. This is temporary and will be removed
+    # when Dreamwidth is fully off of MogileFS.
+    if ( %LJ::MOGILEFS_CONFIG && $LJ::MOGILEFS_CONFIG{hosts} ) {
+        $log->debug( 'Initializing MogileFS blobstore.' );
+        push @{$blobstores ||= []}, DW::BlobStore::MogileFS->init;
+    }
+
+    $log->logcroak( 'Must configure $LJ::BLOBSTORE or %LJ::MOGILEFS_CONFIG.' )
```
kareila: Typo, should be %LJ::BLOBSTORE.

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+    return $blobstores if defined $blobstores;
+
+    # Site owners must configure one of these. Disk is configured by default
+    # to point to our run directory which will work for test environs.
+    if ( exists $LJ::BLOBSTORE{s3} ) {
+        $log->debug( 'Initializing S3 blobstore.' );
+        push @{$blobstores ||= []}, DW::BlobStore::S3->init( %{$LJ::BLOBSTORE{s3}} );
+    }
+    if ( exists $LJ::BLOBSTORE{localdisk} ) {
+        $log->logcroak( 'Must only define a single $LJ::BLOBSTORE entry.' )
```
kareila: Typo, should be %LJ::BLOBSTORE.

cgi-bin/DW/BlobStore.pm
```diff
@@ -0,0 +1,157 @@
...
+    # Try blobstores in priority order.
+    foreach my $bs ( @{$class->_get_blobstores} ) {
+        my $rv = $bs->exists( $namespace, $key );
+        return $rv if $rv;
+    }
+    $log->info( "Meta-blobstore: file doesn't exist in any store ($namespace, $key)" );
+    return undef;
+}
+
+1;
```
kareila: I'm assuming that you've tested all the functionality of this and the various submodules (except S3 which looks like a placeholder), but I'd also like to see some automated tests for these methods.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+        eval {
+            $up->set_description ($POST->{"desc_$picid"})
+                unless $POST->{"desc_$picid"} eq $POST->{"desc_orig_$picid"};
+        } or push @$errors, $@;
+
+    }
+
+    foreach my $kw (keys %used_keywords) {
+        next unless $used_keywords{$kw} > 1;
+        push @$errors, BML::ml('.error.keywords', {ekw => $kw});
```
kareila: I see that you've moved the functions from editicons.bml into a controller module, which is a good thing to start on, but I'm fairly certain BML::ml won't work the same in the new context.  (The strings that are %ML lookups are presumably fine.)  The way we've handled this elsewhere is to define a helper function that prepends the full page context in front of any string keys.  This would also let you stop passing around %ML, unless you have a reason to think these functions will ever be called from other page contexts.  Also, use LJ::Lang::ml instead of BML::ml.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+
+            if ( $size == 0 ) {
+                $current_upload{error} = $ML->{'.error.nofile'};
+            } else {
+                my ( $imagew, $imageh, $filetype ) =
+                    Image::Size::imgsize( $current_upload{image} );
+
+                # couldn't parse the file
+                if ( !$imagew || !$imageh ) {
+                    $current_upload{error} = BML::ml( '.error.unsupportedtype', { filetype => $filetype } );
```
kareila: See above re: BML::ml.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+
+                # let's try to use the factory
+                } elsif ( int($imagew) > 100 || int($imageh) > 100 || $size > $MAX_UPLOAD ) {
+                    # file wrong type for factory
+                    if ( $filetype ne 'JPG' && $filetype ne 'PNG' ) {
+                        # factory only works on jpegs and pngs because Image::Magick has issues
+                        if ( int($imagew) > 100 || int($imageh) > 100 ) {
+                            $current_upload{error} = $ML->{'.error.giffiledimensions'};
+                        } else {
+                            $current_upload{error} = BML::ml( '.error.filetoolarge',
```
kareila: See above re: BML::ml.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+        $upload{md5sum} = $upload{md5ctx}->digest;
+        $POST->{$curr_name} = \$upload{spool_data};
+        return 1;
+    };
+
+
+    # parse multipart-mime submission, one chunk at a time,
+    # calling our hooks as we go to put uploads in temporary
+    # MogileFS filehandles
+    my $retval = eval { BML::parse_multipart_interactive($errorref, {
```
kareila: Do we have a non-BML equivalent for this?  I'm trying not to commit BML functions into controllers.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+    # still have refcounts, they won't really be deleted because
+    # the async job will realize and leave them alone
+    unless ( $retval ) {
+        # if we hit a parse error, delete the uploaded files
+        foreach my $mogkey ( @uploaded_files ) {
+            DW::BlobStore->delete( temp => $mogkey );
+        }
+
+        if (index(lc($$errorref), 'unknown format') == 0) {
+            $$errorref = BML::ml(".error.unknowntype");
```
kareila: See above re: BML::ml.

cgi-bin/LJ/ConfCheck/General.pm
```diff
@@ -759,7 +758,6 @@ my %bools = (
              "OPENID_COMPAT" => "Support pre-1.0 OpenID specs as well as final spec.",
              "OPENID_STATELESS" => "Speak stateless OpenID.  Slower, but no local state needs to be kept.",
              "ONLY_USER_VHOSTS" => "Don't allow www.* journals at /users/ and /~ and /community/.  Only allow them on their own user virtual host domains.",
-             "USERPIC_MOGILEFS" => "Store userpics on MogileFS.",
```
kareila: You remove the USERPIC_MOGILEFS setting here, and the mogfs_userpic_key user method below, but those are both still required by bin/upgrading/migrate-userpics.pl. Should we remove that script?  (It applies to any rows in userpic2 located somewhere other than mogile - given that you've added blobstore as a new possible value in that column, I'm guessing it should go away.)

htdocs/editicons.bml
```diff
@@ -117,32 +119,7 @@ use strict;
 
             my $MAX_UPLOAD = LJ::Userpic->max_allowed_bytes($u);
             my $error;
-
-            # act like the factory is disabled if no mogile
-            my $factory_disabled = !LJ::is_enabled('userpicfactory') || !LJ::mogclient();
```
kareila: It looks like we don't allow the userpicfactory to be disabled any more? This removes a check for LJ::is_enabled('userpicfactory') from /editicons - not that it was ever documented in the config files anywhere that I can see.  Do you know why it was allowed to be disabled in the past?

htdocs/editicons.bml
```diff
@@ -202,33 +180,28 @@ use strict;
...
-                $current_upload{make_default} = $POST{make_default};
-                $current_upload{index} = 0;
-
+                my %current_upload = (
+                    key   => 'userpic_0',
+                    image => \${$picinfo->[0]},
+                    index => 0,
+                );
+                $current_upload{$_} = $POST{$_}
+                    foreach qw/ keywords default comments description make_default /;
```
kareila: Typo!!! Key is descriptions, not description.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -0,0 +1,391 @@
...
+    my $requires_factory;
+
+    # go through each key and create a %current_upload for it, then
+    # put it in @uploads.
+    foreach my $userpic_key ( keys %$POST ) {
+        my @tokens = split(/_/, $userpic_key);
+        my $counter = $tokens[1];
+
+        my %current_upload = (
+            image        => $POST->{$userpic_key},
```
kareila: This assignment would have been `\$POST->{$userpic_key}` in the old version.  Was removing the backslash a typo?

htdocs/editicons.bml
rahaeli: IIRC (and this is just vague memory) the userpic factory requires some other modules to be installed and potentially can take up a lot of disk space, and it was also a little bit buggy at first when it was added, so it was set up as an is_enabled so that LJ clones that didn't want to use it could turn it off.

cgi-bin/ljlib.pl
zorkian: Right, running `bin/checkconfig.pl` pre-push is usually part of the process. (Or should be, and any new module is added there.)

cgi-bin/ljlib.pl
zorkian: We'll remove it and I'll update the ~28 other places. I just didn't yet, this diff isn't final.

cgi-bin/ljlib.pl
kareila: It wasn't on the checklist that we put together however many months ago. I'll add it. ✅

bin/upgrading/update-db.pl
zorkian: Haha, I think it was a braino... but yeah, I have no idea. Will remove.

cgi-bin/ljlib.pl
kareila: Does it make more sense to do that now, or to keep it around as a way of quickly seeing which other places still need to be updated? Or are you planning to update them all before this is merged?

bin/incoming-mail-inject.pl
zorkian: I think I was just grepping for mogclient and converting things and did IncomingEmailHandle before I got to the worker and realized I could just get rid of the handle. I'll remove the module.

bin/upgrading/update-db.pl
zorkian: Oh :'( Nice catch.

bin/upgrading/update-db.pl
zorkian: MogileFS is deprecated. People need to migrate to plain-store or S3 (to be included soon). I'm not supporting MogileFS going forward. So, this code doesn't have any purpose -- it's only useful for setting up a new MogileFS cluster.

cgi-bin/LJ/Config.pm
zorkian: Same as above.

etc/workers.conf
zorkian: I am not sure this file is actually attached to anything anymore... we certainly don't use it, I don't even know what parses this. So, I think nothing happens either way.

cgi-bin/DW/BlobStore.pm
zorkian: Oh wow! TIL. Thank you!

cgi-bin/DW/BlobStore.pm
zorkian: Comment removed. I decided against that later. (Although we can ship a default config file with a setting like this, which might be worth doing.)

cgi-bin/DW/BlobStore.pm
zorkian: Fixed!

cgi-bin/DW/BlobStore.pm
zorkian: Fixed.

cgi-bin/DW/BlobStore.pm
kareila: Yeah, the config hash you shared online earlier today with a default local disk setup would be a good thing to include in the stock config.

cgi-bin/DW/BlobStore.pm
zorkian: Good call, updated.

cgi-bin/DW/BlobStore.pm
zorkian: I'd like to see those too. :'(

etc/workers.conf
kareila: I think dev sites may still use it? Pretty sure mine is set up that way, although I rarely run the entire worker suite in the course of testing things.

cgi-bin/DW/Controller/EditIcons.pm
zorkian: So LJ::Lang::ml just calls BML::ml which uses $BML::ML_SCOPE which is set whenever a BML file is viewed. So, I think the prefixing with the fully-qualified path isn't technically necessary? I can't find one of these helper functions, can you link one?

kareila: Oh, OK! That means it works for now because the calling context is still BML, but it will suddenly stop working when the rest of the page is converted, so that's still a long-term sadness.

Here's an example of a the sort of thing I was talking about - this example uses a variable instead of a function: https://github.com/dreamwidth/dw-free/blob/develop/cgi-bin/DW/Controller/Settings.pm#L51

kareila: (Also using LJ::Lang::ml is a forward-looking thing; when all the pages are converted and we can replace BML::ml with something else, it will be less painful to make the change.)

cgi-bin/DW/Controller/EditIcons.pm
zorkian: Fixed.

cgi-bin/DW/Controller/EditIcons.pm
zorkian: Fixed.

cgi-bin/DW/Controller/EditIcons.pm
zorkian: Fixed.

cgi-bin/LJ/ConfCheck/General.pm
zorkian: Yeah. Deleted as-is, but I'll probably use parts of it for BlobStore migration code, maybe.

htdocs/editicons.bml
zorkian: Yeah it was optional b/c of resources/setup costs (it required MogileFS) but BlobStore makes all file storage work (even in simple mode) so we can use those features even on Hacks. I figured with that true, it'd be worth simplifying/deleting the option.

htdocs/editicons.bml
zorkian: Wow, nice catch!

cgi-bin/DW/Controller/EditIcons.pm
zorkian: Agreed, moving everything to LJ::Lang::ml and removed the %ML usage. I'm leaving things un-anchored for the moment and agree re: long-term sadness. This whole mess of code is sad and needs to be cleaned up to migrate to TT so I figure in that refactor we can fix the language stuff.

cgi-bin/DW/Controller/EditIcons.pm
zorkian: I... don't know. I've re-added it and will test. Okay. That breaks everything. I think removing it is necessary but I don't know why. Uh. It's probably an interaction with the difference in what BlobStore expects as the input? :|

cgi-bin/DW/Controller/EditIcons.pm
zorkian: Investigation reveals yes, `DW::Request::Base` has `sub uploads` which is a multipart parser. I'll have to do some refactoring but I think I can ditch the BML call.

zorkian: This turns out to be a little harder than expected. Will do tomorrow. zz.

zorkian:
@kareila Updated with feedback, should include everything. Even basic tests of the functionality. I also rebased and merged in your updates and ported them to BlobStore. I'm going to proceed with migrating the rest of the mogclient functions now.

zorkian:
Ported all the rest of the mogclients.

```
All tests successful.
Files=127, Tests=5333, 213 wallclock secs ( 0.92 usr  0.18 sys + 139.63 cusr 15.70 csys = 156.43 CPU)
Result: PASS
```

cgi-bin/DW/BlobStore.pm
```diff
@@ -21,24 +21,33 @@ use v5.10;
 use Log::Log4perl;
 my $log = Log::Log4perl->get_logger( __PACKAGE__ );
 
-use DW::BlobStore::LocalDisk;
-use DW::BlobStore::MogileFS;
-use DW::BlobStore::S3;
+use File::Temp;
+
+LJ::ModuleLoader->require_subclasses('DW::BlobStore');
```
kareila: I think you also need to `use LJ::ModuleLoader; ` here - it's not included by default.

cgi-bin/DW/Controller/EditIcons.pm
```diff
@@ -9,7 +9,7 @@
 # Authors:
 #      Mark Smith <mark@dreamwidth.org>
 #
-# Copyright (c) 2016 by Dreamwidth Studios, LLC.
+# Copyright (c) 2016-2017 by Dreamwidth Studios, LLC.
```
kareila: ⌛️ :small_airplane:

cgi-bin/LJ/User/Account.pm
```diff
@@ -12,8 +12,11 @@
 # part of this distribution.
 
 package LJ::User;
+
 use strict;
-no warnings 'uninitialized';
```
kareila: We may get a bunch of new warnings in dev server environment by taking this out, on code we inherited from LJ.

cgi-bin/ljlib.pl
```diff
@@ -54,12 +54,24 @@ BEGIN
...
+        my $conf = q{
+log4perl.rootLogger=FATAL, DevNull
+
+log4perl.appender.DevNull=Log::Log4perl::Appender::File
+log4perl.appender.DevNull.filename=/dev/null
+log4perl.appender.DevNull.layout=Log::Log4perl::Layout::SimpleLayout
+        };
+        Log::Log4perl::init( \$conf );
+    } else {
+        Log::Log4perl::init_and_watch($LJ::HOME . '/etc/log4perl.conf', 10);
```
kareila: Did you address my comment about changing the default log level in log4perl.conf from DEBUG to INFO?  Or did you decide it wasn't necessary?

etc/config-private.pl
```diff
@@ -165,6 +165,16 @@
...
+
+    # Configuration of BlobStore. This is the new storage abstraction used to
+    # store any blobs (images, userpics, media, etc) that need storage. For small
+    # sites/single servers, the localdisk mode is useful. For production
+    # systems S3 should be used.
+    # %BLOBSTORE = (
+    #     localdisk => {
+    #         path => "$LJ::HOME/var/blobstore",
+    #     },
+    # );
```
kareila: Isn't this a better fit for config-local.pl?

cgi-bin/DW/BlobStore/LocalDisk.pm
```diff
@@ -120,7 +120,7 @@ sub delete {
 	my $fqfn = $self->get_location_for_key( $namespace, $key );
 
 	# If the file doesn't exist, pretend success
```
kareila: I think this comment needs to be updated since it no longer matches the given behavior...

bin/delete-bogus-media.pl
```diff
@@ -61,20 +63,17 @@ sub usage {
...
-    say "MogileFS not available, exiting.";
-    exit 0;
-}
-
 foreach my $id ( @mv ) {
     # create a fake object to get the mogkey
     my $fakeobj = bless { userid => $u->id, versionid => $id }, 'DW::Media::Photo';
     my $mogkey = $fakeobj->mogkey;
+
     # this will return true whether the file existed or not
```
kareila: This comment should also be updated 😊

etc/config-private.pl
zorkian: I don't think so?

* config.pl - stuff that is useful to every site
* config-local.pl - stuff that customizes the site to DW, but isn't production related; i.e., a Hack is made to 'behave just like DW' by sharing the same config-local.pl
* config-private.pl - stuff that is particular to a running version of the site like DB credentials, IPs, paths, etc.

cgi-bin/ljlib.pl
zorkian: The config file for production isn't the one here (that one will certainly be different), and I think for Hacks/non-DW prod then DEBUG is a reasonable default? Do you disagree?

cgi-bin/LJ/User/Account.pm
zorkian: Yeah that's probably code we should fix, yeah? I fixed one of them in the RPC popup stuff. Do you think it's too much/we shouldn't do that?

kareila: As long as we are aware of the consequences 😁

cgi-bin/ljlib.pl
kareila: I agree with using DEBUG for hacks, yes.

etc/config-private.pl
kareila: I thought everything in config-private ended up in package DW::PRIVATE and referenced in config-local, at least that has been my experience - but I'm fine either way, just wanted to check.

zorkian: I don't honestly remember why I created DW::PRIVATE, but it's only a small part of the config-private.pl configuration file. Some things in the LJ package reach into DW::PRIVATE. If there's a place to document what _should_ be in these files I'm all ears, btw, I can update.

bin/delete-bogus-media.pl
zorkian: Done!

cgi-bin/DW/BlobStore/LocalDisk.pm
zorkian: Whoops, nice.

cgi-bin/DW/BlobStore.pm
zorkian: Weirdly this was all working last night but I think due to some dark magic... today it wasn't working and this was the fix. Wat. Thanks!

zorkian:
I think I addressed everything, would you like me to do anything else before merging?

bin/incoming-mail-inject.pl
```diff
@@ -40,55 +41,46 @@ ()
...
+$tempfail->( $@ ) if $@;
 
-# just shut postfix up
-if ($ignore_message || should_ignore($msg)) {
-    exit(0);
-}
+if ( $len > IN_MEMORY_THRES ) {
+    my $md5 = md5_hex( $msg );
+    DW::BlobStore->store( temp => "ie:$md5", \$msg );
+    $log->info( "Storing email in blobstore at key: ie:$md5")
```
kareila: This fails to compile - missing terminating semicolon.

bin/incoming-mail-inject.pl
```diff
@@ -40,55 +41,46 @@ ()
...
-my $len = 0;
-my $ieh;
-my $ignore_message = 0;  # bool: to ignore rest of message.
+my $len = 0;   # length of message
+
 eval {
-    while ($rv = sysread(STDIN, $buf, 1024*64)) {
+    my ( $buf, $rv );
+    while ( $rv = sysread STDIN, $buf, 1024*64 ) {
         next if $ignore_message;
```
kareila: This fails to compile - `Global symbol "$ignore_message" requires explicit package name`

I think you've streamlined the process such that this line should just be removed?

kareila:
If you will fix the errors in incoming-mail-inject.pl and make sure it's working correctly, I think this can proceed. Thanks!

cgi-bin/LJ/Userpic.pm
```diff
@@ -524,27 +548,11 @@ sub keywords {
...
-    }
-
-    # check userpicblob2 table
-    my $dbb = LJ::get_cluster_reader($u)
-        or return undef;
-
-    my $data = $dbb->selectrow_array( "SELECT imagedata FROM userpicblob2 WHERE ".
-                                      "userid=? AND picid=?", undef, $self->userid,
-                                      $self->picid );
-    return $data ? $data : undef;
```
kareila: So I just discovered another problem - this is dropping backward compatibility for retrieving icons stored in userpicblob2, which will include any site which wasn't previously running MogileFS (which is most of them because it was too hard to set up).  I didn't notice this immediately because of my browser cache.  We need to either keep this fallback, or include a process for migrating these images out of userpicblob2 when upgrading.

zorkian:
@kareila Updated to fix incoming-mail-inject.pl, also tested. Also fixed other syntax errors that cropped up.

kareila:
LGTM, I'll give it one more spin in the morning.

kareila:
I think I'm ready to squash and merge this, unless you'd rather do the honors yourself.  Any further suggestions for improvements I can make as new PRs.

kareila:
Further comment can be directed to http://dw-dev.dreamwidth.org/195174.html

PR #1832, opened 2016-07-02
Per-user Selective Screening

### LINKED ISSUE #1234: Demi-ban: screen all future comments from specific user
From http://dw-suggestions.dreamwidth.org/1352415.html

> Force-screen comments from one particular user, when it's just that user whose comments warrant screening.

woggy:
New methodology to allow forced screening of all comments from
a particular user. In terms of backend, functions much like a ban, but
with neutral wording to reflect the fact that this behavior is desired
for many users, and not a negative consequence. There is not (yet) an
equivalent page to /manage/banusers for this functionality, so for the
moment it is only accessible through the admin console.

Fixes #1234

### DIFF
--- a/cgi-bin/DW/Console/Command/ScreenSet.pm
+++ b/cgi-bin/DW/Console/Command/ScreenSet.pm
@@ -2,10 +2,8 @@
 #
 # DW::Console::Command::ScreenSet
 #
-# Console commands for setting and clearing suspend notes. If a
-# suspend note is set for an account, trying to suspend that
-# account will cause an error and make you confirm you really
-# want to do that.
+# Console command for listing adding a user to selective screening for a given account.
+# Based on LJ::Console::Command::BanSet
 #
 # Authors:
 #      Paul Niewoonder <woggy@dreamwidth.org>
--- a/cgi-bin/DW/Console/Command/ScreenSet.pm
+++ b/cgi-bin/DW/Console/Command/ScreenSet.pm
@@ -1,17 +1,23 @@
-# This code was forked from the LiveJournal project owned and operated
-# by Live Journal, Inc. The code has been modified and expanded by
-# Dreamwidth Studios, LLC. These files were originally licensed under
-# the terms of the license supplied by Live Journal, Inc, which can
-# currently be found at:
+#!/usr/bin/perl
 #
-# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+# DW::Console::Command::ScreenSet
 #
-# In accordance with the original license, this code and all its
-# modifications are provided under the GNU General Public License.
-# A copy of that license can be found in the LICENSE file included as
-# part of this distribution.
+# Console commands for setting and clearing suspend notes. If a
+# suspend note is set for an account, trying to suspend that
+# account will cause an error and make you confirm you really
+# want to do that.
+#
+# Authors:
+#      Paul Niewoonder <woggy@dreamwidth.org>
+#
+# Copyright (c) 2016 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
 
-package LJ::Console::Command::ScreenSet;
+package DW::Console::Command::ScreenSet;
 
 use strict;
 use base qw(LJ::Console::Command);
@@ -19,7 +25,7 @@ use Carp qw(croak);
 
 sub cmd { "screen_set" }
 
-sub desc { "Set another user's comments to be automatically screened in your journal or community." }
+sub desc { "Set another user's comments to be automatically screened in your journal or community. Requires priv: none." }
 
 sub args_desc { [
                  'user' => "The user you want to screen comments from.",
@@ -57,12 +63,12 @@ sub execute {
 
     my $screenlist = LJ::load_rel_user($journal, 'S') || [];
     return $self->error("You have reached the maximum number of users to automatically screen. Remove a user and try again.")
-        if scalar(@$screenlist) >= 500;
+        if scalar(@$screenlist) >= $LJ::SEL_SCREEN_LIMIT;
 
     LJ::set_rel($journal, $screenuser, 'S');
     $journal->log_event('screen_set', { actiontarget => $screenuser->id, remote => $remote });
 
-    return $self->print("User " . $screenuser->user . "'s comments in " . $journal->user . " will now be automatically screened.");
+    return $self->print("Comments from user " . $screenuser->user . " in " . $journal->user . " will now be automatically screened.");
 }
 
 1;
--- a/cgi-bin/LJ/Console/Command/ScreenList.pm
+++ b/cgi-bin/LJ/Console/Command/ScreenList.pm
@@ -1,0 +1,67 @@
+# This code was forked from the LiveJournal project owned and operated
+# by Live Journal, Inc. The code has been modified and expanded by
+# Dreamwidth Studios, LLC. These files were originally licensed under
+# the terms of the license supplied by Live Journal, Inc, which can
+# currently be found at:
+#
+# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+#
+# In accordance with the original license, this code and all its
+# modifications are provided under the GNU General Public License.
+# A copy of that license can be found in the LICENSE file included as
+# part of this distribution.
+
+package LJ::Console::Command::ScreenList;
+
+use strict;
+use base qw(LJ::Console::Command);
+use Carp qw(croak);
+
+sub cmd { "screen_list" }
+
+sub desc { "Lists users who are being automatically screened by an account." }
+
+sub args_desc { [
+                 'user' => "Optional; lists automatic screens in a community you maintain, or any user if you have the 'finduser' priv."
+                 ] }
+
+sub usage { '[ "from" <user> ]' }
+
+sub can_execute { 1 }
+
+sub execute {
+    my ($self, @args) = @_;
+    my $remote = LJ::get_remote();
+    my $journal = $remote;         # may be overridden later
+
+    return $self->error("Incorrect number of arguments. Consult the reference.")
+        unless scalar(@args) == 0 || scalar(@args) == 2;
+
+    if (scalar(@args) == 2) {
+        my ($from, $user) = @args;
+
+        return $self->error("First argument must be 'from'")
+            if $from ne "from";
+
+        $journal = LJ::load_user($user);
+        return $self->error("Unknown account: $user")
+            unless $journal;
+
+        return $self->error("You are not a maintainer of this account")
+            unless $remote && ( $remote->can_manage( $journal )
+                                || $remote->has_priv( "finduser" ) );
+    }
+
+    my $screenids = LJ::load_rel_user($journal, 'S') || [];
+    my $us = LJ::load_userids(@$screenids);
+    my @userlist = map { $us->{$_}{user} } keys %$us;
+
+    return $self->info($journal->user . " is not automatically screening any other users.")
+        unless @userlist;
+
+    $self->info($_) foreach @userlist;
+
+    return 1;
+}
+
+1;
--- a/cgi-bin/LJ/Console/Command/ScreenSet.pm
+++ b/cgi-bin/LJ/Console/Command/ScreenSet.pm
@@ -1,0 +1,68 @@
+# This code was forked from the LiveJournal project owned and operated
+# by Live Journal, Inc. The code has been modified and expanded by
+# Dreamwidth Studios, LLC. These files were originally licensed under
+# the terms of the license supplied by Live Journal, Inc, which can
+# currently be found at:
+#
+# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+#
+# In accordance with the original license, this code and all its
+# modifications are provided under the GNU General Public License.
+# A copy of that license can be found in the LICENSE file included as
+# part of this distribution.
+
+package LJ::Console::Command::ScreenSet;
+
+use strict;
+use base qw(LJ::Console::Command);
+use Carp qw(croak);
+
+sub cmd { "screen_set" }
+
+sub desc { "Set another user's comments to be automatically screened in your journal or community." }
+
+sub args_desc { [
+                 'user' => "The user you want to screen comments from.",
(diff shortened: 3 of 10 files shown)


cgi-bin/LJ/Console/Command/ScreenList.pm
```diff
@@ -0,0 +1,67 @@
...
+# Dreamwidth Studios, LLC. These files were originally licensed under
+# the terms of the license supplied by Live Journal, Inc, which can
+# currently be found at:
+#
+# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+#
+# In accordance with the original license, this code and all its
+# modifications are provided under the GNU General Public License.
+# A copy of that license can be found in the LICENSE file included as
+# part of this distribution.
```
kareila: I had this discussion with @kaberett recently, and I'm not sure this is adequately documented: new Console commands need to go in DW/Console/Command instead of LJ/Console/Command, with a Dreamwidth copyright notice instead of a LiveJournal one.  You can look at the existing commands in DW/Console/Command to see what that copyright should look like, and do let me know if you have any questions about the details.

cgi-bin/LJ/Console/Command/ScreenList.pm
```diff
@@ -0,0 +1,67 @@
...
+
+package LJ::Console::Command::ScreenList;
+
+use strict;
+use base qw(LJ::Console::Command);
+use Carp qw(croak);
+
+sub cmd { "screen_list" }
+
+sub desc { "Lists users who are being automatically screened by an account." }
```
kareila: Make sure to also include the "Requires priv: none." language here, for consistency with the rest of the command reference.

cgi-bin/LJ/Console/Command/ScreenSet.pm
```diff
@@ -0,0 +1,68 @@
...
+            unless $remote && $remote->can_manage( $journal );
+    }
+
+    my $screenuser = LJ::load_user($user);
+    return $self->error("Unknown account: $user")
+        unless $screenuser;
+
+    my $screenlist = LJ::load_rel_user($journal, 'S') || [];
+    return $self->error("You have reached the maximum number of users to automatically screen. Remove a user and try again.")
+        if scalar(@$screenlist) >= 500;
```
kareila: 500 seems like a good limit, but it also seems like something that should be able to be overridden.  I suggest making this a parameter in cgi-bin/LJ/Global/Defaults.pm, something like `$LJ::SEL_SCREEN_LIMIT`.  (Banning has a similar parameter, `$LJ::MAX_BANS`, which should also be defined in cgi-bin/LJ/Global/Defaults.pm, but isn't, presumably because it predates that file.)

cgi-bin/LJ/Console/Command/ScreenSet.pm
```diff
@@ -0,0 +1,68 @@
...
+
+    LJ::set_rel($journal, $screenuser, 'S');
+    $journal->log_event('screen_set', { actiontarget => $screenuser->id, remote => $remote });
+
+    return $self->print("User " . $screenuser->user . "'s comments in " . $journal->user . " will now be automatically screened.");
```
kareila: I'm not looking very closely at the overall structure of the commands here, since I can tell you've based them off the existing commands for bans, but I do want to recommend that this message be rephrased to avoid the possessive "'s" after the username.  I suggest instead "Comments from user [user] in [journal] will now be..."

cgi-bin/LJ/Console/Command/ScreenUnset.pm
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    my $screenuser = LJ::load_user($user);
+    return $self->error("Unknown account: $user")
+        unless $screenuser;
+
+    LJ::clear_rel($journal, $screenuser, 'S');
+    $journal->log_event('screen_unset', { actiontarget => $screenuser->id, remote => $remote });
+
+    return $self->print("User " . $screenuser->user . "'s comments in " . $journal->user . " will no longer be automatically screened.");
```
kareila: Same as above, re possessives.

cgi-bin/LJ/Talk.pm
```diff
@@ -1599,7 +1599,8 @@ sub talkform {
 
             $ret .= " " . $BML::ML{'.opt.willscreen'} if $screening eq 'A'
                     || ( $screening eq 'R' && !$remote->is_validated )
-                    || ( $screening eq 'F' && !$journalu->trusts($remote) ) ;
+                    || ( $screening eq 'F' && !$journalu->trusts($remote) )
+                    || ( $journalu->has_autoscreen( $remote ) );
```
kareila: Style nitpick - you don't need parentheses around a method call: `|| $journalu->has_autoscreen( $remote );`

cgi-bin/LJ/Talk.pm
```diff
@@ -1643,9 +1644,9 @@ sub talkform {
                 $ret .= "<strong>$logged_in</strong>";
 
                 # show willscreen if a) all comments are screened b) anonymous is screened and OpenID user not validated, c) non-access is screened and OpenID user
-                # is not on access list
+                # is not on access list d) this user is being automatically screened
                 $ret .= $BML::ML{'.opt.willscreen'} if $screening eq 'A' || ( $screening eq 'R' && !$remote->is_validated )
-                    || ( $screening eq 'F' && !$journalu->trusts($remote) ) ;
+                    || ( $screening eq 'F' && !$journalu->trusts($remote) ) || ( $journalu->has_autoscreen( $remote ) ) ;
```
kareila: Same as above.  Thank you for updating the comment!

cgi-bin/LJ/Talk.pm
```diff
@@ -1705,7 +1706,7 @@ sub talkform {
 
                     $ret .= "<strong>$logged_in</strong>";
 
-                    $ret .= $BML::ML{'.opt.willscreen'} if $screening;
+                    $ret .= $BML::ML{'.opt.willscreen'} if $screening || ( $journalu->has_autoscreen( $remote ) );
```
kareila: Same as above.

I see a similar line at L1628 that may also need updating?

cgi-bin/LJ/Talk.pm
```diff
@@ -1758,7 +1759,7 @@ sub talkform {
                                                   { user => $journalu->user } );
                 }
 
-                $ret .= $BML::ML{'.opt.willscreen'} if $screening eq 'A';
+                $ret .= $BML::ML{'.opt.willscreen'} if $screening eq 'A' || ( $journalu->has_autoscreen( $remote ) );
```
kareila: Same as above.

cgi-bin/LJ/Talk.pm
```diff
@@ -1830,7 +1831,8 @@ sub talkform {
 
             $ret .= "<input type='hidden' name='cookieuser' value='$remote->{'user'}' id='cookieuser' />\n";
             if ($screening eq 'A' ||
-                ($screening eq 'F' && !$journalu->trusts_or_has_member( $remote ))) {
+                ($screening eq 'F' && !$journalu->trusts_or_has_member( $remote )) ||
+                ( $journalu->has_autoscreen( $remote ) )) {
```
kareila: Same as above.  I also see a similar construction at L1662 and L1775; do those also need updating?  They are used for OpenID users, but presumably those can be subject to selective screening as well.

cgi-bin/LJ/Talk.pm
```diff
@@ -3331,7 +3337,8 @@ sub init {
     my $screening = LJ::Talk::screening_level($journalu, $ditemid >> 8) || "";
     if ($screening eq 'A' ||
         ($screening eq 'R' && ! $up) ||
-        ($screening eq 'F' && !($up && $journalu->trusts_or_has_member( $up )))) {
+        ($screening eq 'F' && !($up && $journalu->trusts_or_has_member( $up ))) ||
+        ($journalu->has_autoscreen( $up ) )) {
```
kareila: Same as above.

cgi-bin/LJ/User/Administration.pm
```diff
@@ -329,6 +329,22 @@ sub unban_user_multi {
...
+# return if $target's comments will automatically be screened in $u's journal
+sub has_autoscreen {
+    my ( $u, $target ) = @_;
+
+    my $uid = LJ::want_userid( $u );
+    my $jid = LJ::want_userid( $target );
+    return 1 unless $uid && $jid;
+    return 0 if $uid == $jid;  # can't autoscreen yourself
+
+    return LJ::check_rel( $uid, $jid, 'S' );
```
kareila: We need to make sure the 'S' relationship is also documented in bin/upgrading/update-db-general.pl at line 1386.

kareila:
Overall this looks great and I have no major objections, just a few housekeeping issues and suggestions for improvements.  Is LJ::Talk::init really the only place in the entire codebase where we create comments?  I figured there would be at least two different paths to worry about, but if there's another one, it's not jumping out at me either.

Thanks again for working on this!  I'm looking forward to including it in the next code push. :)

azurelunatic:
What path do emailed comments go through? I'm not quite sure how to follow
this rabbit hole.

kaberett:
I'm pretty sure you want to be checking cgi-bin/DW/EmailPost/Comment.pm!

kareila:
Emailed comments go through LJ::Protocol, which calls LJ::Comment->create, which uses the same subroutine in LJ::Talk. I do think everything eventually ends up routed through that one function.

cgi-bin/LJ/Console/Command/ScreenSet.pm
woggy: Nifty. (I totally agree that it shouldn't be hardcoded, but wasn't sure where to put the parameters.)

cgi-bin/LJ/Talk.pm
woggy: ...Right. It looks mismatched to me without them, but i'm also a somewhat overcautious parenthesizer. I'll take them out.

cgi-bin/LJ/Talk.pm
woggy: 1628 is dealing with Anonymous comments in particular, so don't need it there.

cgi-bin/LJ/Talk.pm
woggy: 1662 and 1775 are for OpenID, yes, but they (both of them) are the path that is "you're not currently logged in, this is what'd happen if you logged in with OpenID". As such, we don't actually have data on "who they'd be logged in as".

woggy:
I had not thought to check emailed comments, but I did test every on-site path i could think of and they all seemed to route through here.

kareila:
Yeah I think it's all good.  I'm just surprised is all. 😆

cgi-bin/LJ/Talk.pm
kareila: Thanks for double checking.  I feel like that logic could be cleaned up to be less confusing, but it can wait for another day.

cgi-bin/LJ/Console/Command/ScreenList.pm
woggy: Given how very little I am changing in these commands, it is unclear to me whether or not the original LJ wording should apply. I can use the updated DW wording, but it seems...disingenuous...to imply myself as Sole Author.

kareila: You can just say "based on LJ::Console::Command::BanFoo" in the module header if you like.

cgi-bin/DW/Console/Command/ScreenSet.pm
```diff
@@ -1,25 +1,31 @@
...
-# by Live Journal, Inc. The code has been modified and expanded by
-# Dreamwidth Studios, LLC. These files were originally licensed under
-# the terms of the license supplied by Live Journal, Inc, which can
-# currently be found at:
+#!/usr/bin/perl
 #
-# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+# DW::Console::Command::ScreenSet
+#
+# Console commands for setting and clearing suspend notes. If a
```
kareila: This block of header text needs to be updated for each command.  As I said previously, you can mention it's based on the corresponding commands from the ban suite.  I think that's all this still needs before it can go in.

woggy: *facepalm* copied the format from another file and didn't finish changing the content. right.

cgi-bin/LJ/User/Administration.pm
```diff
@@ -329,6 +329,22 @@ sub unban_user_multi {
...
 ########################################################################
+### Selective Screening functions
+
+# return if $target's comments will automatically be screened in $u's journal
+sub has_autoscreen {
+    my ( $u, $target ) = @_;
+
+    my $uid = LJ::want_userid( $u );
+    my $jid = LJ::want_userid( $target );
+    return 1 unless $uid && $jid;
```
kareila: Sorry for waiting so long for another round of review.  This one line just jumped out at me as problematic: if $jid is undefined (as it would be for logged-out comments), shouldn't this method return 0 and not 1?  Otherwise we will end up always screening logged-out comments, which is... not what we want.

woggy: You are entirely correct. I'd copied the code from has_banned, which neatly avoids the problem by only ever being called in places where the logic is such that anon posters never hit the function call. has_autoscreen is in *almost* the same situation - six of the seven function calls are wrapped in similar logic. The one that isn't is, of course, the one that actually determines whether the comment should be screened at all (the others are all logic about whether to display the "this comment will be screened" message at the top of the replybox).

It's still good to have that early return for anons, I think, to avoid having to make a check_rel call just to find out that no, we're not autoscreening anons. So I've flipped the return value to 0.

kareila:
Okay! After staring and staring at this, I think I've finally figured out a better way to handle the multiple logic checks in Talk.pm.  Right after the first line that says `my $screening = LJ::Talk::screening_level(...)` what if we just did this:
```
$screening = 'A' if $journalu->has_autoscreen( $up );
```
I think that covers all the different conditionals for form display by basically saying "pretend the screening level is All Comments if the remote user is subject to selective screening."

What do you think?

woggy:
While I find the content of the suggestion to be valid, and leading to (marginally) faster code, albeit at the price of a small bit of obfuscation, I am *deeply* annoyed to have such a minor nitpick arrive so far after my initial commit.

It makes me wonder what the next thing that "needs to be fixed" will be, and if this pull request will ever actually be accepted.

kareila:
I do apologize! The review process can take several iterations, and new features tend to be more closely scrutinized than small fixes. I am determined to get this merged before the next code push, which is tentatively scheduled for October 1st. Hopefully this will be my last suggested improvement.

PR #1836, opened 2016-07-03
Alway show Beta info, even when the user is not presently involved in a beta

### LINKED ISSUE #1835: include "beta testing" support request info even when no active beta options are selected
I thought this had already been done at some point, but I don't see any evidence of it:

In the "Dreamwidth" section of the support request header, in addition to style, cluster, sitescheme, etc. it would be nice to see whether the user has opted into the Beta New Entries page.  Or in the general case, any beta features currently active on the site...

pinterface:
Because the absense of information is very different from an explicitly indicated negative.

Fixes #1835, assuming I'm understanding what $u->in_class() does, anyway.

### DIFF
--- a/htdocs/support/see_request.bml
+++ b/htdocs/support/see_request.bml
@@ -295,10 +295,11 @@ body<=
                 if $view_userlog;
         }
 
-        if ( $u->in_class( LJ::BetaFeatures->cap_name )
+        if ( %LJ::BETA_FEATURES
             && LJ::Support::has_any_support_priv( $remote ) ) {
             $ret .= "<br />$ML{'.betatesting'}: ";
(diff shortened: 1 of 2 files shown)


kareila:
How about we replace that `$u->in_class` conditional with a test to see if `%LJ::BETA_FEATURES` is defined?  That way if the site hasn't configured any beta features, the line won't print at all (as before), but if it has, the line will print whether or not the user has opted into the beta.

pinterface:
Seems reasonable.  PR updated.

It just uses a simple boolean test.  I was going to use defined, but `perldoc -f defined` says defined on aggregates is deprecated and to just use a straight boolean, so... I did.

Issue #1713, opened 2016-03-18
Allow embeds from JSFiddle


kaberett:
Embed code:

`<iframe width="100%" height="300" src="//jsfiddle.net/5c0ruh8s/10/embedded/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>`

Support request: https://www.dreamwidth.org/support/see_request?id=32942

srukle:
claim

PR #1837, opened 2016-07-06
Allow embeds from JSFiddle

srukle:
adds JSFiddle to whitelist for issue #1713

### DIFF
--- a/cgi-bin/DW/Hooks/EmbedWhitelist.pm
+++ b/cgi-bin/DW/Hooks/EmbedWhitelist.pm
@@ -66,7 +66,7 @@ my %host_path_match = (
     
     "books.google.com"      => [ qr!^/ngrams/!, 1 ],
     
-    "jsfiddle.net"          => [ qr!/embedded/$! ],
+    "jsfiddle.net"          => [ qr!/embedded/$!, 1 ],
 
     "www.kickstarter.com"   => [ qr!/widget/[a-zA-Z]+\.html$!, 1 ],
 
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Hooks/EmbedWhitelist.pm
```diff
@@ -65,6 +65,8 @@ my %host_path_match = (
     "docs.google.com"       => [ qr!^/(document|spreadsheets?|presentation)/!, 1 ],
     
     "books.google.com"      => [ qr!^/ngrams/!, 1 ],
+    
+    "jsfiddle.net"          => [ qr!^/.*/embedded/$!, 1 ],
```
kareila: Using `.*` in regexps is usually a bad idea, causing the computer to do a lot of extra work to match characters we don't care about.  Just match the end of the path: `qr!/embedded/$!`

kareila:
It's also a good idea to add a test case for the new site to t/embed-whitelist.t to make sure things are working as expected.

Thanks :)

srukle:
Updated whitelist entry and tested embedded example in dreamhack.

Example in initial request works with changes to whitelist entry as suggested in September within dreamhack environment.
https://www.dreamwidth.org/support/see_request?id=32942

spinda:
Any hope of this getting merged?

kareila:
I should have time to take a look at it soon!

kareila:
So, there is one thing still preventing this from getting merged: the second element of the array, the '1' has been removed and should be readded.  This signals whether or not the given site supports https and this one does.

kaberett:
@kareila ooh, that's not mentioned explicitly at http://wiki.dreamwidth.net/wiki/index.php/Adding_Embed_Whitelist_Items -- might be good to add it? Will try to remember to get to it...

srukle:
@kareila, change was made today.

kareila:
@kaberett wiki is updated - I didn't know we had a page for these :)

@srukle thanks for updating this so quickly!

kaberett:
@kareila thank you!

@srukle thank you also :-)

PR #1844, opened 2016-07-12
[#1843] let has_same_email_as ignore +mailbox components of emails

### LINKED ISSUE #1843: new has_same_email_as comparison doesn't discard + in addresses
Discovered when trying to use the bonus_icons xfer console command that the comparison function added for that ($u->has_same_email_as) doesn't ignore +whatever in email addresses, which a lot of people use to filter their inbox. (For instance, account username1 has example+username1@example.com, username2 has example+username2@example.com, etc.)

For purposes of comparison, we should only be looking at things to the left of the +.

(This comparison function is also used in renaming, which should improve things there as well.)

kareila:
This does a simple string equality check as before, but if
the match fails, strip out anything between + and @ and try
again before giving up.

Fixes #1843.

### DIFF
--- a/cgi-bin/LJ/User/Message.pm
+++ b/cgi-bin/LJ/User/Message.pm
@@ -75,7 +75,15 @@ sub email_raw {
 sub has_same_email_as {
     my ( $u, $other ) = @_;
     croak "invalid user object passed" unless LJ::isu( $u ) && LJ::isu( $other );
-    return lc( $u->email_raw ) eq lc( $other->email_raw );
+
+    my $email_1 = lc( $u->email_raw );
+    my $email_2 = lc( $other->email_raw );
+    return 1 if $email_1 eq $email_2;
+
+    # if unequal, try stripping any +mailbox addressing
+    $email_1 =~ s/\+[^@]+@/@/;
+    $email_2 =~ s/\+[^@]+@/@/;
+    return $email_1 eq $email_2;
 }
 
 sub email_status {


kaberett:
So having forgotten about this edge case, I think it's probably *also* worth flagging up that specifically gmail addresses will let you put full stops anywhere in an e-mail address and treat the with-full-stops version of the address as identical to the without-full-stops version (https://support.google.com/mail/answer/10313?hl=en).

I think it's *less* likely that we'll run into that than +es (sorry for overlooking that!) but it's possibly also worth accounting for this as well?

rahaeli:
I think no -- that's Gmail-specific behavior (RFCs say that example.username@example.org and exampleusername@example.org are separate addresses), while the +addressing is universal. I also don't see people using that often "in the wild", while the +addressing is pretty common.

(And don't worry about it, I forgot about it as well!)

pinterface:
Strictly speaking, according to RFCs, Bob@example.com and bob@example.com are separate addresses, though case-insensitive local-parts are pretty much universal outside of pedants trying to break things. :)

+addressing is also host-specific.  It's not as widespread as case-insensitivity, but you're probably reasonably safe assuming that if you see a + in a local-part it's being used to shuffle mail around inside the same mailbox.

Obscure E-mail Knowledge Sprite, away!

cgi-bin/LJ/User/Message.pm
```diff
@@ -75,7 +75,15 @@ sub email_raw {
 sub has_same_email_as {
     my ( $u, $other ) = @_;
     croak "invalid user object passed" unless LJ::isu( $u ) && LJ::isu( $other );
-    return lc( $u->email_raw ) eq lc( $other->email_raw );
+
+    my $email_1 = lc( $u->email_raw );
```
zorkian: For next time, the parenthesis are unnecessary in this usage and stylistically to be avoided.

Issue #1180, opened 2015-02-01
Quick Update not reflecting security settings


ninetyd:
On http://www.dreamwidth.org/, the security menu of Quick Update is set to Everyone (Public) no matter which setting you've selected as your minimum security level. Posts get the correct security level but this is confusing.
Also, wording doesn't change when you post to a community: it's still access list/private instead of members/admins.

chrisboyle:
claim

PR #1861, opened 2016-09-04
Make Quick Update respect minsecurity

chrisboyle:
Fixes: #1180

With apologies for leaving this claimed and inactive for so long!


chrisboyle:
Oh, hang on, I remember what was difficult about this: the option disablement needs to update based on the neighbouring dropdown that sets where you're posting. :-(

PR #1865, opened 2016-09-18
convert /tools/search to TT/Foundation

kareila:


zorkian:
Another one bites the dust...!

Issue #1849, opened 2016-08-07
Amazon SES worker needs to verify SMTP connection was established


kareila:
`send-email-ses` recently needed to be restarted because it lost communication with Amazon.  @alierak says that the worker currently doesn't do anything to verify that the requested SMTP connection was established before beginning to use it.

kareila:
I am now being told that perhaps the Net::SMTP timeout argument isn't being honored by the SSL handshake, which looks for SSL_* options as implemented by IO::Socket::SSL.

alierak:
send-email used TheSchwartz::Worker::SendEmail, which contained a replacement for the Net::SMTP constructor that would impose separate timeouts for SMTP commands versus the SMTP connection itself. send-email-ses implements its own TheSchwartz::Worker::SendEmail package without actually using the real one as best I can tell, so it's probably also going to need similar improvements to its Net::SMTPS or Net::SMTP constructor.

kareila:
I've done the easy bits, but it still needs better connection status checking.

PR #1871, opened 2016-09-26
better handling of smtp variables

### LINKED ISSUE #1850: Amazon SES worker has unused $_SMTP variable
Again from @alierak pointing out that there's a variable in `send-email-ses` that looks like it was intended to cache an open SMTP connection and reuse it across requests, but the variable is never assigned a value other than undef.

kareila:
The first commit partially addresses #1849 and is already live in production.

The second commit should fix #1850, but hasn't been tested (and can't be without an SES setup) - although it seems straightforward, unforeseen side effects are a possibility.

### DIFF
--- a/bin/worker/send-email-ses
+++ b/bin/worker/send-email-ses
@@ -55,6 +55,8 @@ sub work {
         Port    => 587,
         Timeout => 60,
     );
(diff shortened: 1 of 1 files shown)


kareila:
OK, so ordered.

zorkian:
Much better -- @alierak do you not have merging powers? I'd be happy with you being able to review and merge things if you want.

PR #1875, opened 2016-09-29
define support box colors and classes for Foundation/SCSS

kareila:
For use with #1666.  See also dreamwidth/dw-nonfree#116.

I copied the Gradation colors from the older gradation.css file.  For the other site schemes, I used the given color from #1666 and calculated a corresponding border color using the SCSS "darken" function.


zorkian:
Doing the CSS finger cross

PR #1883, opened 2016-10-01
Use properly generated font CSS

rshatch:


kareila:
Looks like the version without the semicolons is the correct one.

PR #1897, opened 2016-12-01
Move concatenating resources to DW

zorkian:
This functionality was previously handled by Perlbal. As part of the
question to get rid of Perlbal in production, we can now do this all in
the Apache layer.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -39,6 +39,7 @@ use DW::Auth;
 use DW::Request::XMLRPCTransport;
 use Cwd qw/abs_path/;
 use Carp qw/ croak confess /;
+use Fcntl ':mode';
 
 BEGIN {
     $LJ::OPTMOD_ZLIB = eval "use Compress::Zlib (); 1;";
@@ -226,6 +227,10 @@ sub send_concat_res_response {
         return 404
             if $mime ne $res->[3];
     }
+
+    # No files? Something went wrong
+    return 404
+        unless $body;
 
     # Got a bunch of files, let's concatenate them into the output
     $apache_r->status( 200 );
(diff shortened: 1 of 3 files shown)


rahaeli:
Here's to simplifying systems!

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -1129,6 +1182,31 @@ sub vgift_content {
...
+    return undef unless -f $fn;
+
+    my $contents;
+    open FILE, "<$fn"
+        or return undef;
+    { local $/ = undef; $contents = <FILE>; }
+    close FILE;
+
+    my $mime;
+    if ( $fn =~ /\.([a-z]+)$/ ) {
```
alierak: Either this should be case-insensitive, or the lc is unnecessary below.

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -1129,6 +1182,31 @@ sub vgift_content {
     return OK;
 }
 
+# load_file_for_concat will retrieve the given file from the disk and return:
+# (contents, size, mtime, mime). If the file doesn't exist, undef is returned.
+sub load_file_for_concat {
+    my $fn = $_[0];
+    my @stat = stat( $fn );
+    return undef unless scalar @stat > 0;
+    return undef unless -f $fn;
```
alierak: I think the info used by -f is already in $stat[2] if you want to avoid statting it twice, but it will probably not look elegant.

alierak:
General comment: does this allow concatenating non-css/js files outside the specified directory, as though they were text/plain, with ../ in filenames?

alierak:
I get a 500 internal server error rather than 404 when I ask it for something nonexistent.

cgi-bin/Apache/LiveJournal.pm
```diff
@@ -198,6 +198,47 @@ sub totally_down_content
     return OK;
 }
 
+sub send_concat_res_response {
+    my $apache_r = $_[0];
+    my $args = $apache_r->args;
+    my $uri = $apache_r->uri;
+
+    my $dir = ( $LJ::STATDOCS // $LJ::HTDOCS ) . $uri;
+    return 404
```
kareila: If you look around line 370 of the current version of the module, you can see what we do to print our 404 page, but that assumes the request status is already 404.

PR #1899, opened 2016-12-02
merge get and post args for Net::OpenID::Server

alierak:
get_args and post_args are deprecated. If both are specified, as we do in htdocs/openid/server.bml, get_args is ignored, which can produce the "not a human-readable resource" error. Let's just merge them into one args hash.


kareila:
I'm not sure anyone completely understands why this is producing the observed behavior, but it looks like a reasonable change and I've confirmed it fixes the known problem, so I'm deploying it.

PR #1901, opened 2016-12-02
[#1900] remove any additional cachebuster arguments

### LINKED ISSUE #1900: failure to load some concatenated resources
http://www.dreamwidth.org/support/see_request?id=35334

As a consequence of #1897 we are now seeing an issue where some concatenated resources fail to load because they have an additional cachebuster argument that doesn't match the regular expression.

Example: http://www.dreamwidth.org/js/??pages/entry/options.js?v=1422503230&_=1480691898357

kareila:
Fixes #1900.

(If we don't think the "v=" will always be first maybe we should just remove everything? But I decided to try the more conservative approach first.)

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -208,8 +208,9 @@ sub send_concat_res_response {
     return 404
         unless -d $dir;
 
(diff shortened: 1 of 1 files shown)


zorkian:
I was slightly worried about eating too many things, but eating is probably OK.

PR #1903, opened 2016-12-03
ignore client IP when checking for trusted proxies

alierak:
ignore client IP when checking for trusted proxies, since it may have been changed by the previous request on this connection.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -101,25 +101,21 @@ sub handler
             @req_hosts = ( $apache_r->connection->client_ip );
             if ( my $forward = $apache_r->headers_in->{'X-Forwarded-For'} ) {
                 my @hosts = split( /\s*,\s*/, $forward );
-                my %seen;
-
-                foreach (@hosts) {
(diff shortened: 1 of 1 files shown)


cgi-bin/Apache/LiveJournal.pm
```diff
@@ -101,25 +101,21 @@ sub handler
...
-                                $real = pop @hosts;
-                            } while ( @hosts && $LJ::IS_TRUSTED_PROXY->( $real ) );
-                        } else {
-                            # Remote IP is not a trusted proxy, just keep it
-                            $real = $req_hosts[0];
-                        }
+                        # Find last IP in X-Forwarded-For that isn't a trusted proxy.
+                        do {
+                            $real = pop @hosts;
+                        } while ( @hosts && $LJ::IS_TRUSTED_PROXY->( $real ) );
```
kareila: If this block doesn't refer to `@req_hosts` any more, where else is the variable used? And do those other references care that you just randomized the order of the list by assigning to it using `keys` instead of `foreach`?

alierak: It's used for sysbans, and order doesn't matter.

PR #1904, opened 2016-12-06
avoid proxy hanging, never return from handleProxyFileRequests

alierak:
This change already in production.


zorkian:
No idea why this was ever a return. :-(

PR #1914, opened 2016-12-27
QuickReply box appears in appropriate location in Drifting

### LINKED ISSUE #1509: Show name of poster on "Edit tags" page

rshatch:
Adds the print_entry_footer code form core2 into Drifting to
give proper targets for the QuickReply box to anchor to.

Fixes #1509

### DIFF
--- a/styles/drifting/layout.s2
+++ b/styles/drifting/layout.s2
@@ -1203,11 +1203,14 @@ function Page::print_entry (Entry e)
         $e->print_text();
         if ($*entry_metadata_position == "bottom") { $e->print_metadata(); }
         $e->print_tags();
-        $e->print_management_links();
-        $e->print_interaction_links("topcomment");
-        $this->print_reply_container({ "target" => "topcomment" });
-
+        $this->print_entry_footer($e);
     $e->print_wrapper_end();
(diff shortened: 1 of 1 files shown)


zorkian:
FOOTER: PRINTED.

PR #1916, opened 2016-12-28
handle not using active-active clusters in prod

### LINKED ISSUE #1909: error in stats/site
Error: undef error - Illegal division by zero at /home/dw/production/views/stats/site.tt line 152.

I'm guessing a cron job isn't running (ping @alierak @zorkian) to produce something, hence the divide-by-zero error, but we should also fix it so it fails gracefully with a "hey dumbass start this cron job" error.

alierak:
Probably fixes #1909 since we switched to writer/reader clusters in prod. Code should check whether active-active clusters are defined, not whether this is a dev server.

### DIFF
--- a/cgi-bin/LJ/DB.pm
+++ b/cgi-bin/LJ/DB.pm
@@ -263,7 +263,7 @@ sub foreach_cluster {
     die $@ if $@;
 
     foreach my $cluster_id (@LJ::CLUSTERS) {
-        my $dbr = ($LJ::IS_DEV_SERVER) ?
+        my $dbr = (defined %LJ::CLUSTER_PAIR_ACTIVE) ?
             LJ::get_cluster_reader($cluster_id) : LJ::DBUtil->get_inactive_db($cluster_id, $opts->{verbose});
         $coderef->($cluster_id, $dbr);
     }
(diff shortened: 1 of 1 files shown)


cgi-bin/LJ/DB.pm
```diff
@@ -263,7 +263,7 @@ sub foreach_cluster {
     die $@ if $@;
 
     foreach my $cluster_id (@LJ::CLUSTERS) {
-        my $dbr = ($LJ::IS_DEV_SERVER) ?
+        my $dbr = %LJ::CLUSTER_PAIR_ACTIVE ?
```
zorkian: Does this do what we want? This will, on prod, make it call `LJ::DBUtil->get_inactive_db` which immediately calls out to do the a/b role selection and inversion, which won't work in our environment.

I think that this can just return `LJ::get_cluster_reader` in all cases, at least for Dreamwidth. For people who are still using an actual master-master setup (which we haven't in years) then it won't do the right thing, but honestly, I think we can drop support for master-master.

alierak: No, that's the problem - currently it is calling get_inactive_db in prod. We don't have CLUSTER_PAIR_ACTIVE defined, so that dies.

alierak: Oh, but my logic is backwards, yeah.

alierak:
Okay, here's a version I wrote earlier and discarded since I thought somebody might want to keep active-active cluster support just in case.

zorkian:
Thanks!

PR #1918, opened 2016-12-28
[#1909] don't divide by zero when undefined

### LINKED ISSUE #1909: error in stats/site

kareila:
In Template Toolkit, undefined does not equal zero, so test for both.

Fixes #1909.


zorkian:
Thanks, TT.

PR #1919, opened 2016-12-28
use dummy encoders for sphinx, strings are already utf8

### LINKED ISSUE #1334: enable Cyrillic in site search
http://www.dreamwidth.org/support/see_request?id=29609

We had a bug open for this in Bugzilla, but it went foom: there's something that needs to be done to enable Cyrillic language support in Sphinx, though I'm not sure what it is. (Assigning to Mark as a sysadmin thing and because I remember you knowing what needs to be done, unassign if I'm wrong)

alierak:
Fixes #1334 in my testing.

### DIFF
--- a/bin/worker/sphinx-search-gm
+++ b/bin/worker/sphinx-search-gm
@@ -34,6 +34,8 @@ sub _run_search {
     my $index = '*';
     $sx->SetServer( @LJ::SPHINX_SEARCHD );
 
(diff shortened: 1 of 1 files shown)


zorkian:
(Fast merge, I had already read your findings on the bug so I was READY!)

This is great. Thank you for tracking this down!

PR #1923, opened 2017-01-01
Cleanup 2016q4

kareila:
Two minor undefined warning fixes:

* Use of uninitialized value $name in string ne ... cgi-bin/LJ/Capabilities.pm line 123.

* Use of uninitialized value in subroutine entry ... cgi-bin/LJ/S2.pm line 300.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -296,6 +296,7 @@ sub s2_run
     };
     my $out_clean = sub {
         my $text = shift;
+        $text = '' unless defined $text;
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/S2.pm
```diff
@@ -296,6 +296,7 @@ sub s2_run
     };
     my $out_clean = sub {
         my $text = shift;
+        $text = '' unless defined $text;
```
zorkian: Alternately:

```
$text //= '';
```

PR #1924, opened 2017-01-02
Safer payments

zorkian:
Fix an issue where preload_props could get no database handle and
wouldn't fail immediately (but would instead fail two lines later).
Also, make the payment system catch failures in the post-payment tasks
and not fail hard.

### DIFF
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -822,6 +822,8 @@ sub preload_props {
                 LJ::get_db_reader();
             $used_slave = 1;
         }
+        die "No db\n" unless $db;
+
         $sql = "SELECT upropid, value FROM $table WHERE userid=$uid";
         if (ref $loadfrom{$table}) {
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/User/Account.pm
```diff
@@ -822,6 +822,8 @@ sub preload_props {
                 LJ::get_db_reader();
             $used_slave = 1;
         }
+        die "No db\n" unless $db;
+
```
kareila: This should probably either give a bit more context, or remove the "\n" so that the line number gets printed.

zorkian:
Updated to use `confess` which prints the backtrace, infinitely more useful

zorkian:
Going to assume this is OK and merge

kareila:
Yeah, I haven't tested it but it looked fine to me. Thanks for tracking it down.

rahaeli:
shit, I wonder if this might be the cause of the occasional mysterious failures that we've been trying to figure out for like FOUR GODDAMN YEARS? A GIRL CAN HOPE

PR #1925, opened 2017-01-02
add stats collection for resolve-extacct

kareila:
If my understanding of DW::Stats is correct, this should report `resolve-extacct` successes and failures to Datadog or similar.

I haven't actually tested this and don't know if any further setup in Datadog is needed.


kareila:
@alierak confirms this is now recording data, but we'll want to observe it for a while before deciding what thresholds to set for any warnings.

Issue #1747, opened 2016-04-12
our https OpenID URLs can't be used to log in to other services


kareila:
DIscovered by @alierak when trying to connect his DW OpenID to his wiki account.  Our site prints a page saying, "The site you just came from wants to verify an identity that you can't provide as [name of logged in user]."  It works when prefaced with http but not https.

PR #1926, opened 2017-01-02
Add HTTPS Everywhere beta

zorkian:
This adds a beta option for testing use of HTTPS Everywhere. This will
allow us to ensure the site is fully working before turning it on for
everybody. When we do turn it on everywhere we'll have to do a little
cleanup to revert most of this.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -483,11 +483,21 @@ sub trans {
         }
     }
 
+    # force SSL if not currently and user is in httpseverywhere beta
+    my $remote = LJ::get_remote();
+    if ( $apache_r->is_initial_req && $LJ::USE_SSL && ! $is_ssl && $remote &&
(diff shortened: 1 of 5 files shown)


zorkian:
Also requires updating %LJ::BETA_FEATURES with:

```
        "httpseverywhere" => {
            start_time => 0,
            end_time => "Inf",
        },
```

afuna:
Oh nice, didn't think of using beta features. Looks good.

rahaeli:
FYI, the only thing I know of that's still outstanding for going all-HTTPS is making our OpenID endpoint support HTTPS (it currently doesn't, so once we flip the switch for HTTPSE, people will be unable to use their DW account as an OpenID elsewhere).

kareila:
Nope, we fixed that!  It was #1747.

rahaeli:
o hai, i thought i tried that last month and it didn't work, but I just tried it again and it did. NEVERMIND ME MY BRAIN IS STILL IN VACATION MODE

PR #1929, opened 2017-01-06
remove references to unused comm_promo_list

kareila:
Found this when investigating the bin/maint directory.  Probably an ljcom thing.


zorkian:
THIS MERGE BROUGHT TO YOU BY CABERNET SAUVIGNON

rahaeli:
Don't remember if it was livejournal or ljcom, but either way, it was being introduced just when we forked -- we caught some of the infrastructure for it but none of the frontend. I think it was part of the whole "discover these top communities" thing.

PR #1930, opened 2017-01-06
[#1896] only populate notifyarchive if esn_archive enabled

### LINKED ISSUE #1896: notifyarchive table shouldn't be used if esn_archive is disabled
From conversation elsewhere...

@zorkian: "DID YOU KNOW... About 5-10% of our database is the `notifyarchive` table which contains
notifications that have been archived. It doesn't contain text -- just the link saying 'user X has notification Y in this state'. This is bonkers..."

@rahaeli: "LOL, that's hysterical. I think the maximum # of notifications you can have in the inbox is 2000 for premium paid accounts? Whatever that # is (it's in the config) we can cap it at that."

Looks like the cap is "inbox_max" which we have set at 2000, 4000, or 6000 depending on account type.  Not sure if we want to do lazy cleanup when a message is archived, or create a worker for the job; since many of the affected accounts are inactive, using a worker may make more sense.

kareila:
Metadata for every notification is saved in this table,
but if the feature is disabled, it's never used anywhere.

Perhaps this "feature" could be entirely removed in the
future, but I think that decision should be delayed until
we have a spec for the requested inbox redesign.  It might
be useful when trying to implement a trash folder, for example.

Fixes #1896.

### DIFF
--- a/cgi-bin/LJ/NotificationInbox.pm
+++ b/cgi-bin/LJ/NotificationInbox.pm
@@ -417,7 +417,7 @@ sub enqueue {
     my ($self, %opts) = @_;
 
     my $evt = delete $opts{event};
-    my $archive = delete $opts{archive} || 1;
+    my $archive = delete $opts{archive} || LJ::is_enabled( 'esn_archive' );
     croak "No event" unless $evt;
     croak "Extra args passed to enqueue" if %opts;
 


kareila:
Even though this is a simple one line change, I would like a close pair of eyes to make sure I'm not missing anything in my attempts to parse the creeping horror of the notification system.

zorkian:
I went through the codepaths and reached the same conclusion. This should be safe for now.

PR #1932, opened 2017-01-09
fixes for failed uploads breaking /file/edit page

### LINKED ISSUE #1928: Error on /file/edit
I keep getting an error when I go to https://www.dreamwidth.org/file/edit

[Error: Column 'height' cannot be null at /home/dw/production/cgi-bin/DW/Media.pm line 62. @ va-web04]

I've gotten this message since uploading some pics apparently failed. I got this on /file/new then:

Unable to instantiate file in MogileFS. at /home/dw/production/cgi-bin/DW/Controller/API/Media.pm line 71 on /file/new

The uploader has been working fine since then but /file/edit won't.

kareila:
This fix is in three stages as detailed in each individual commit:

- Provide a more informative error if get_file_data fails in the _resize method.

- Print the error to the log instead, and hide any problematic images from the user.

- Reorder operations in upload_media to prevent the problem from happening again.

Fixes #1928.

### DIFF
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -126,6 +126,19 @@ sub upload_media {
     # now we can cook -- allocate an id and upload
     my $id = LJ::alloc_user_counter( $opts{user}, 'A' )
         or croak 'Unable to allocate user counter for uploaded file.';
+
+    # to avoid having database rows for an image that failed to upload,
+    # do the upload first - we can create a fake object to get the mogkey
+
+    # FIXME: have different MogileFS classes for different media types
+
+    my $fakeobj = bless { userid => $opts{user}->id, versionid => $id }, 'DW::Media::Photo';
+    my $fh = $mog->new_file( $fakeobj->mogkey, 'media' )
+        or croak 'Unable to instantiate file in MogileFS.';
+    $fh->print( $opts{data} );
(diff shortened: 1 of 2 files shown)


cgi-bin/DW/Media.pm
```diff
@@ -202,8 +202,17 @@ sub get_active_for_user {
     return () unless $rows && ref $rows eq 'ARRAY';
 
     # construct media objects for each of the items and return that
-    return sort { $b->logtime <=> $a->logtime }
-           map { DW::Media->new( user => $u, mediaid => $_, %opts ) } @$rows;
+    my @media;
+    foreach ( @$rows ) {
+        # use eval to catch croaks
+        my $obj = eval { DW::Media->new( user => $u, mediaid => $_, %opts ) };
```
zorkian: I sort of wonder what the best way to actually handle errors in Perl is. If the library hits a problem, croak? And let upstreams catch it? Or return undefs and errstrs? I don't even know anymore and we use every single paradigm in the DW code...

zorkian:
I'm slightly dubious about inverting the order here -- i.e., this just abandons things in the other way, but it's at least in a way that doesn't show up to the user and we can handle internally (running scrubbers to clean up unlinked files).

So, it's a better user experience and so I'll merge. It would be nice to actually catch errors in either process and do a total cleanup, but we don't need to for now.

Issue #1931, opened 2017-01-09
Backend: deleting an image on /file/edit deletes the original but not the thumbnail


kareila:
I don't know if this is a bug or not, but it surprised me:

The delete method as defined in DW::Media::Base only deletes the instantiated object, which always points to the original image.  But whenever an image is uploaded, it also automatically creates a reduced thumbnail if the image is larger than 200x200.  That thumbnail lives on in storage after the original image is deleted.

I noticed this in testing when after uploading and deleting a test image a couple of times, I expected my MogileFS database to be empty and it wasn't.

kareila:
The following SQL statement should find the information for any media files that would be orphaned by this condition:
```sql
SELECT mv.userid, mv.versionid FROM media_versions AS mv, media AS m WHERE \
m.mediaid=mv.mediaid AND m.userid=mv.userid AND	mv.mediaid != mv.versionid AND m.state='D';
```

kareila:
I have drafted a fix for this but don't want to add any new code that requires MogileFS.  Once Blobstore is landed, I'll revise and submit if we agree that this is a bug.

PR #1933, opened 2017-01-09
Script for deleting bogus media data

kareila:
This will allow a sysadmin to easily change the deletion
state for any problematic rows logged by the fix for #1928,
or to delete any orphaned thumbnails as reported in #1931.

### DIFF
--- a/bin/delete-bogus-media.pl
+++ b/bin/delete-bogus-media.pl
@@ -1,0 +1,80 @@
+#!/usr/bin/perl
+
+use v5.10;
+use strict;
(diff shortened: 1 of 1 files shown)


zorkian:
To make sure I understand -- this script deletes a single media item + content, and is to be used when an admin gets a list of things to delete from some external source. I.e., from running the 'find abandoned thumbnails' query given separately?

kareila:
Right, although @alierak said something about that query not being phrased well for production... The changes for get_active_media will also log errors for the existing bogus rows (when visiting /file/edit) that could be used with this.

PR #1943, opened 2017-01-13
[#1338] display different payment error for duplicate transactions

### LINKED ISSUE #1338: payment processor "duplicate transaction" response code not separate from "payment type not supported" response code
I _think_ this is -free and not -nonfree, but I'm not 100% sure -- please let me know if I need to move it.

There are several possible failure states for payments. The two most common are:
1. payment type not supported (customer trying to use a CC we don't support or trying to use a debit card that is not also a credit card)
2. customer is trying to make a second payment for an identical amount within a certain range of time (I think I've narrowed it down to ~1 hour) -- most commonly, "buying X months of time for Account X" and then "buying X months of time for Account Y" on two separate transactions instead of as one transaction -- and the payment processor is rejecting the transaction as a possible duplicate.

The error message that both of those failure conditions display to the user is /shop/cart.bml.payment.failed, which only makes reference to situation 1, payment type not supported. This results in a lot of very confused people who ask me "but I know the payment type is supported because I JUST USED IT, why is it failing", and I have to explain situation 2 to them and tell them to chill out for an hour and then try again (oh and next time either do it on the same transaction or buy points and apportion them out).

Ideally, I would love it if we could detect the two different failure modes based on the response code we get from the processor and display a different error message, something like:

> Duplicate transaction
> Our payment processor prevents <a href="http://www.dreamwidth.org/support/faqbrowse?faqid=232">two transactions for the same amount</a> within a short timeframe. This helps to prevent accidental double charges. Please try your payment again in an hour or two. To keep this from happening in the future, you can either upgrade all your accounts in a single transaction, by buying the upgrades for your other accounts as a gift, or you can buy Dreamwidth points for the total amount of the payments you want to make and then transfer them to your other accounts.

If that can't be done, and I'm genuinely not sure if it can be, let me know and I will edit the error message in /shop/cart.bml.payment.failed to be inclusive of that language instead.

kareila:
I haven't tested this because figuring out how to fake a credit card transaction on my dreamhack is a deeper rabbit hole than I care to explore today, but this seems like a straightforward change, as long as I didn't make any typos.  😬 

Fixes #1338.

### DIFF
--- a/htdocs/shop/cart.bml
+++ b/htdocs/shop/cart.bml
@@ -39,7 +39,14 @@ body<=
     $ret .= qq{
         <div class='warning-box message-box'>
             <div class='title'>$ML{'.payment.failed.title'}</div>
-            <p>$ML{'.payment.failed'}</p>
+            <p>$ML{'.payment.failed.duplicate'}</p>
+        </div>\n}
+        if $GET{duplicate};
+
+    $ret .= qq{
+        <div class='warning-box message-box'>
+            <div class='title'>$ML{'.payment.failed.title'}</div>
+            <p>$ML{'.payment.failed.other'}</p>
         </div>\n}
         if $GET{failed};
 
--- a/htdocs/shop/cart.bml.text
+++ b/htdocs/shop/cart.bml.text
@@ -3,7 +3,9 @@
 
 .gift=Give a Gift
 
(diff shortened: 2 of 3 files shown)


zorkian:
I don't see any typos!

PR #1947, opened 2017-01-20
remove unused module SixDegrees.pm

kareila:
I found another one - doesn't appear to relate to anything else except the 'friends' table which was dropped long ago.


zorkian:
The machete giveth the machete.. wait, it doesn't give.. IT JUST TAKES AWAY!

PR #1949, opened 2017-01-24
better automatic word-break behavior

### LINKED ISSUE #1948: problems with HTML cleaner's insertion of "wbr" tags
http://www.dreamwidth.org/support/see_request?id=35705

Looks like it may need to be told to ignore URL text, and definitely needs to update the list of void elements.

kareila:
Now the HTML cleaner will prefer punctuation marks for word breaks, if available, and has been told that &lt;/wbr> is not a thing.  Includes many tests for the new behavior.

Fixes #1948.

### DIFF
--- a/t/clean-event.t
+++ b/t/clean-event.t
@@ -17,7 +17,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 42;
+use Test::More tests => 45;
 
 BEGIN { $LJ::_T_CONFIG = 1; require "$ENV{LJHOME}/cgi-bin/ljlib.pl"; }
 use LJ::CleanHTML;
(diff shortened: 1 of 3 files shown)


rahaeli:
ooh yay that's been a problem for like 15 years, it just happens rarely enough that nobody ever bothered fixing it...

kareila:
My new motto: no problem too rare, no fix too over-engineered.

Actually, that's a terrible motto, never mind 😜

rahaeli:
yeah but let's face it it's TOTALLY TRUE

<3

t/clean-event.t
```diff
@@ -293,4 +293,37 @@ note( "mismatched and misnested tags" );
...
+$clean->({ wordlength => 8 });
+is( $orig_post, $clean_post, "Word break handling of HTML entities OK" );
+
+$orig_post  = qq{multi-character-string test};
+$clean_post = qq{multi-character-<wbr />string test};
+$clean->({ wordlength => 20 });
+is( $orig_post, $clean_post, "Choose last punctuation in string" );
+
+$orig_post  = qq{"This_is_a_test_of_the_emergency_word_break_system."};
+$clean_post = qq{"This_is_a_test_of_the_emergency_word_br<wbr />eak_system."};
```
zorkian: This is supposed to break mid-word as opposed to post-`_`?

kareila: I used the underscore character in this test because it's a special character included in \w, and I was making sure the word break wasn't inserted after the initial quotation mark.

kareila: (Although if you don't want underscore treated differently, that's understandable, but I couldn't think of a simple way to overcome that exception.)

zorkian: Nope this is fine! I was just curious about the thinking.

PR #1953, opened 2017-02-02
1436/customize2 tt

rshatch:
This is probably *almost* beta-ready - I think there may be a few major bugs left but I can't remember where they are and my testing methods are not particularly organized or robust in coverage, and I'd really just like some other eyes on it at this point.


zorkian:
There's definitely debugging code in here with warn and stuff.

Also a lot of the formatting seems odd, the Javascript files are all indented several spaces?

This is a pretty large diff of course, which is why I keep putting off reviewing it. It's hard to find a chunk of time to sit down and fully understand. I'm also not sure how to break it down given what it does.

How bad would it be if we just committed it so people could start testing it on their hacks/or we could fix forward in live when issues are found? I feel like the risk is just 'stuff doesn't work' rather than 'exposes data that it shouldn't'?

rshatch:
Yeah, I think failure mode of most of the stuff is just 'doesn't work'. I'll take out the debugs and see if I can split this up into smaller chunks - the scope kind of spiraled out of control as I was working on it because of all the embedded widgets across multiple sections, etc.

zorkian:
Given the scope of breakage, I wouldn't work too hard to break it up. If
you can that's great but I am comfortable committing it together and
having people test it.
(But it's totally up to you of course!)

rshatch:
Let me take a look - I think the /advanced/ pages are an easy split-out, and I miiiight be able to split /customize and /customize/options, but unfortunately all the stuff under /customize/options (which is probably half the code, at a minimum) is where things are liable to break, and also can't be split up because it's all interdependent.

rshatch:
Closing, as I'm breaking this up into multiple smaller requests

PR #1956, opened 2017-02-04
OpenID fixes

### LINKED ISSUE #1744: Replying to PMs from OpenID users should Just Work but Really Doesn't
http://www.dreamwidth.org/support/see_request?id=33116

> Hello, I recently got a message from someone via their wordpress openID. When I try replying from the inbox page, when I hit the send button it complains about their wordpress url not being a valid user name.
> 
> I did manage to send a reply by using the link in the email I received about the message (which uses their 'ext_9999' identifier rather than their wordpress url as the "name"). But I thought I should still send this report so y'all are aware of the inconsistency/bugginess.

Hitting Reply from the Inbox page should, in fact, automagically map to the ext_XXXX username, without the user having to apply workarounds.
#1691, #1615, and #303 might be useful to look at for guidance!

### LINKED ISSUE #1907: strip off http/https when checking OpenID logins
http://www.dreamwidth.org/support/see_request?id=34570

We've ended up with a situation where someone has two OpenID accounts from the same provider, because one started with https and the other with http.  Let's prevent this from happening again.

### LINKED ISSUE #1939: don't let someone try to claim a deleted OpenID account
Had somebody having OpenID-claiming problems, and when I looked into it, the OpenID account they were trying to claim was statusviz D. I'm not sure if that's the reason it was failing (I told them to undelete the OID account and try again, then let me know if it still failed), but the system shouldn't have let them try at all -- it should return an error message. (Something like "You must first <a href="https://www.dreamwidth.org/openid">log in</a> to the OpenID account and <a href="https://www.dreamwidth.org/accountstatus">undelete it</a> before you can claim the account.")

kareila:
* Fixes #1744 - failure when replying to PMs from OpenID users.
* Fixes #1939 - don't allow a user to claim a deleted OpenID account.
* Fixes #1907 - using https when http was used previously when logging in with OpenID.

### DIFF
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -994,8 +994,26 @@ sub load_existing_identity_user {
     my ($type, $ident) = @_;
 
     my $dbh = LJ::get_db_reader();
-    my $uid = $dbh->selectrow_array("SELECT userid FROM identitymap WHERE idtype=? AND identity=?",
-                                    undef, $type, $ident);
+    my $uid;
+
+    # if given an https URL, also look for existing http account
+    # (we should have stripped the protocol before storing these, sigh)
+    if ( $ident =~ s/^https:// ) {
+        my $secure_ident= "https:$ident";
+        $ident = "http:$ident";
+
+        # do the secure lookup first; if it fails, try the fallback below
+        $uid = $dbh->selectrow_array( "SELECT userid FROM identitymap WHERE " .
+                                      "idtype=? AND identity=?",
+                                      undef, $type, $secure_ident );
+    }
+
+    unless ( $uid ) {
+        $uid = $dbh->selectrow_array( "SELECT userid FROM identitymap WHERE " .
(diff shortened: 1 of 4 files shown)


cgi-bin/LJ/User/Account.pm
```diff
@@ -994,8 +994,26 @@ sub load_existing_identity_user {
     my ($type, $ident) = @_;
 
     my $dbh = LJ::get_db_reader();
-    my $uid = $dbh->selectrow_array("SELECT userid FROM identitymap WHERE idtype=? AND identity=?",
-                                    undef, $type, $ident);
+    my $uid;
+
+    # if given an https URL, also look for existing http account
+    # (we should have stripped the protocol before storing these, sigh)
```
zorkian: Agreed. Is this something worth doing now as part of a data migration or something? Do we think we've got collisions if we do that?

kareila: There would be at least one collusion, from the accounts that the support request was referring to.

PR #1959, opened 2017-02-07
add compile checking of command line scripts

kareila:
Per @zorkian's request, t/00-compile.t will now also check CLI scripts in bin (as long as they end in .pl).


zorkian:
Yay, thank you so much!

PR #1960, opened 2017-02-08
enforce limits for media (photo) usage

### LINKED ISSUE #1940: we should determine and enforce limits for media hosting
Now that the file upload area is being more actively used and developed, we should make sure we have quotas in place for the different account types, and make sure they are documented.

kareila:
Don't allow offsite loading of images, and enforce storage quotas.

Fixes #1940.

### DIFF
--- a/cgi-bin/DW/Controller/API/Media.pm
+++ b/cgi-bin/DW/Controller/API/Media.pm
@@ -59,6 +59,9 @@ sub file_new_handler {
     LJ::isu( $rv->{u} )
         or return api_error( $r->HTTP_UNAUTHORIZED, 'Not logged in' );
 
+    return api_error( $r->HTTP_BAD_REQUEST, 'Quota exceeded' )
+        unless $u->can_upload_media;
+
     my $uploads = $r->uploads;
     return api_error( $r->HTTP_BAD_REQUEST, 'No uploads found' )
         unless ref $uploads eq 'ARRAY' && scalar @$uploads;
--- a/cgi-bin/DW/EmailPost/Entry.pm
+++ b/cgi-bin/DW/EmailPost/Entry.pm
@@ -297,6 +297,8 @@ sub _upload_images {
     my @imgs = $self->get_entity( $self->{_entity}, 'image' );
     return 1 unless scalar @imgs;
 
+    return 1401 unless $u->can_upload_media;  # error code from insert_images
+
     my @images;
     foreach my $img_entity ( @imgs ) {
         my $obj = DW::Media->upload_media(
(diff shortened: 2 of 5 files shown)


rahaeli:
I haven't looked into the exact behavior of check_referer, but: what does it do if the image is viewed inline on DW but by someone who has referer checking disabled in their browser? (I ran into that problem when I tried to block remote loading of MP3s from my web space ages ago....)

cgi-bin/DW/Media.pm
```diff
@@ -217,5 +217,38 @@ sub get_active_for_user {
...
+
+    # convert megabytes -> bytes
+    return $cap * 1024 * 1024;
+}
+
+sub get_usage_for_user {
+    my ( $class, $u ) = @_;
+    confess 'Invalid user' unless LJ::isu( $u );
+
+    my ( $usage ) = $u->selectrow_array(
```
zorkian: We might need to memcache this at some point ... but maybe not? Since it's just an upload check?

alierak:
It's cool, check_referer accepts requests with no Referer header.

rahaeli:
Cool, thank you for easing my mind!

cgi-bin/DW/Controller/API/Media.pm
```diff
@@ -59,6 +59,9 @@ sub file_new_handler {
     LJ::isu( $rv->{u} )
         or return api_error( $r->HTTP_UNAUTHORIZED, 'Not logged in' );
 
+    return api_error( $r->HTTP_BAD_REQUEST, 'Quota exceeded' )
+        unless $u->can_upload_media;
```
zorkian: $u was undefined here

cgi-bin/DW/EmailPost/Entry.pm
```diff
@@ -297,6 +297,8 @@ sub _upload_images {
     my @imgs = $self->get_entity( $self->{_entity}, 'image' );
     return 1 unless scalar @imgs;
 
+    return 1401 unless $u->can_upload_media;  # error code from insert_images
```
zorkian: also $u is not defined here

cgi-bin/DW/Controller/API/Media.pm
kareila: whooooops, sorry!

kaberett:
IT OCCURS TO ME that banning off-site loading of images is... going to break the way users expect cross-posting images to work. I've had a quick chat with Jen about this, but could we maybe consider whitelisting LJ/IJ/DJ/etc for the sake of the support queue?

PR #1961, opened 2017-02-08
Updates to the image uploader and management system

### LINKED ISSUE #1936: Most-requested file-manager tweaks for roll-out
http://dw-dev.dreamwidth.org/194532.html

Most changes want to happen to https://www.dreamwidth.org/file/edit. Requested UI changes are:
* _paginate_ (so that the page remains useful for people who've uploaded a lot of files); Azz's preference is to have the first page be the first images uploaded, the last page be the most recent images uploaded, and the default landing place to be the last page
* option to display in grid instead of list?
* if dates are recorded anywhere, display date-of-upload
* show file size and height*width of full image
* provide image title and description (and ability to edit/save edits)
* provide the HTML for embedding an image
* bare bone link to your full-size image or simple HMTL with IMG instead of FIGURE/A etc (req'd by 90d)
* provide options for embedding a range of preview sizes of image (radio buttons?)
* if you've never uploaded anything, visiting /file/edit says "Here are all the files you've uploaded from mobile." (i) this should say "You haven't uploaded anything" with a link to /file/new; (ii) if users HAVE uploaded files, the message should read "Here are all the files you've uploaded."

Navigation:
* Organize --> Manage Files
* have internal links between the subpages (/file, /file/new, /file/edit

Features (stretch goals!):
* allow user-set preview sizes
* allow users to rename images
* allow users to sort images according to... criteria?
* user-friendly error messages
* allow tags/folders/etc
* a thumbnail gallery

Needs docs:
* supported image types
* max size (if any)
* ... an entire FAQ

rshatch:
-adds an index page with links to all subpages
-adds editable textfields for title, description and alt-text
-adds two views for showing all files, with pagination
-list view shows all info about the original file with embed code
for fullsized and selectable thumbnail sized versions of the image
-grid view shows title, file size and dimensions, and thumbnail embed code
-pload date and time is now shown in list view
-file list now displays correct text when no files have been uploaded
-all subpaes are now crosslinked with other subpages
-uploader page now has alt-text field, and displays message on successful upload
-language across pages has been standardized to say 'images' instead of a mix
of 'files' and 'images'.
-edit page has been converted to Foundation
-backend changes make original file's url, size, and dimensions available to
resized versions of that image

Fixes #1936

### DIFF
--- a/cgi-bin/DW/Controller/Media.pm
+++ b/cgi-bin/DW/Controller/Media.pm
@@ -85,7 +85,7 @@ sub media_bulkedit_handler {
 
             my %props;
             while ( my ($key, $val) = each %post ) {
-                next if $key =~ m/delete/;
+                next if $key eq "delete";
                 next unless $key =~ m/^(\w+)-(\d+)/;
                 my $mediaid = $2 >> 8;
                 if ( exists $props{$mediaid} ) {
--- a/cgi-bin/DW/Controller/Media.pm
+++ b/cgi-bin/DW/Controller/Media.pm
@@ -25,8 +25,9 @@ use DW::BlobStore;
 use DW::Routing;
 use DW::Request;
 use DW::Controller;
+use POSXIX;
 
-my %VALID_SIZES = ( map { $_ => 1 } ( 100, 320, 200, 640, 480, 1024, 768, 1280,
+my %VALID_SIZES = ( map { $_ => $_ } ( 100, 320, 200, 640, 480, 1024, 768, 1280,
             800, 600, 720, 1600, 1200 ) );
 
 DW::Routing->register_regex( qr!^/file/(\d+)$!, \&media_handler,
@@ -45,15 +46,13 @@ sub media_manage_handler {
     # load all of a user's media.  this is inefficient and won't be like this forever,
     # but it's simple for now...
     my @media = DW::Media->get_active_for_user( $rv->{remote}, width => 200, height => 200 );
-    my @valid_sizes;
-    map { push @valid_sizes, ( $_, $_ ) } keys %VALID_SIZES;
 
     $rv->{media} = \@media;
     $rv->{make_embed_url} = \&make_embed_url;
     $rv->{page} = $rv->{r}->get_args->{page} || '1';
     $rv->{view_type} = $rv->{r}->get_args->{view} || '';
     $rv->{maxpage} = POSIX::ceil(scalar @media / 20);
-    $rv->{valid_sizes} =\@valid_sizes;
+    $rv->{valid_sizes} =[ %VALID_SIZES ];
     $rv->{convert_time} = \&LJ::mysql_time;
 
     return DW::Template->render_template( 'media/index.tt', $rv );
@@ -86,7 +85,7 @@ sub media_bulkedit_handler {
 
             my %props;
             while ( my ($key, $val) = each %post ) {
-                next unless $key =~ m/^(\w*)-(\d+)/;
+                next unless $key =~ m/^(\w+)-(\d+)/;
                 my $mediaid = $2 >> 8;
                 if ( exists $props{$mediaid} ) {
                     $props{$mediaid}{$1} = $val;
--- a/cgi-bin/DW/Controller/Media.pm
+++ b/cgi-bin/DW/Controller/Media.pm
@@ -44,7 +44,17 @@ sub media_manage_handler {
 
     # load all of a user's media.  this is inefficient and won't be like this forever,
     # but it's simple for now...
-    $rv->{media} = [ DW::Media->get_active_for_user( $rv->{remote} ) ];
+    my @media = DW::Media->get_active_for_user( $rv->{remote}, width => 200, height => 200 );
+    my @valid_sizes;
+    map { push @valid_sizes, ( $_, $_ ) } keys %VALID_SIZES;
+
+    $rv->{media} = \@media;
+    $rv->{make_embed_url} = \&make_embed_url;
+    $rv->{page} = $rv->{r}->get_args->{page} || '1';
+    $rv->{view_type} = $rv->{r}->get_args->{view} || '';
+    $rv->{maxpage} = POSIX::ceil(scalar @media / 20);
+    $rv->{valid_sizes} =\@valid_sizes;
+    $rv->{convert_time} = \&LJ::mysql_time;
 
     return DW::Template->render_template( 'media/index.tt', $rv );
 }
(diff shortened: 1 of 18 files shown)


rahaeli:
this PR makes me so exciiiiiiited

cgi-bin/DW/Controller/Media.pm
```diff
@@ -39,7 +39,17 @@ sub media_manage_handler {
 
     # load all of a user's media.  this is inefficient and won't be like this forever,
     # but it's simple for now...
-    $rv->{media} = [ DW::Media->get_active_for_user( $rv->{remote} ) ];
+    my @media = DW::Media->get_active_for_user( $rv->{remote}, width => 200, height => 200 );
+    my @valid_sizes;
+    map { push @valid_sizes, ( $_, $_ ) } keys %VALID_SIZES;
```
kareila: Instead of doing this extra work here, why not just change the initial VALID_SIZES assignment from `$_ => 1` to `$_ => $_`?  None of the existing checks care what the value of the hash is, just whether the key exists.

cgi-bin/DW/Controller/Media.pm
```diff
@@ -39,7 +39,17 @@ sub media_manage_handler {
...
+
+    $rv->{media} = \@media;
+    $rv->{make_embed_url} = \&make_embed_url;
+    $rv->{page} = $rv->{r}->get_args->{page} || '1';
+    $rv->{view_type} = $rv->{r}->get_args->{view} || '';
+    $rv->{maxpage} = POSIX::ceil(scalar @media / 20);
```
kareila: When using a function from a module, it's best practice to make sure the module is included in your file, unless you know that one of the modules you're already explicitly including also uses it.  So make sure you have `use POSIX;` above, or possibly `use POSIX qw( ceil );` if you want to be specific about what functions are imported.

cgi-bin/DW/Controller/Media.pm
```diff
@@ -39,7 +39,17 @@ sub media_manage_handler {
...
+    $rv->{valid_sizes} =\@valid_sizes;
```
kareila: If you take my advice above, you should be able to change this to `[ %VALID_SIZES ]`.

cgi-bin/DW/Controller/Media.pm
```diff
@@ -63,14 +73,38 @@ sub media_bulkedit_handler {
...
-                my $media = DW::Media->new( user => $rv->{u}, mediaid => $mediaid );
+
+            # transform our HTML field names to property names
+            # and group by what media object they belong to
+            # we don't care if the id or prop might not exist
+            # right now, later steps will verify them
+
+            my %props;
+            while ( my ($key, $val) = each %post ) {
+                next unless $key =~ m/^(\w*)-(\d+)/;
```
kareila: I think this should be \w+, not \w* - otherwise you risk $1 being undefined below.

htdocs/stc/media.css
```diff
@@ -1,34 +1,48 @@
+#media-grid {
+    pasdding:0;
```
kareila: Fix the typo, or remove the declaration if it didn't make a difference :)

htdocs/stc/media.css
```diff
@@ -40,59 +54,27 @@
     padding: 1em 0;
 }
 
-#media-list .media-item div, #media-list .media-item select, #media-list .media-item img {
+/*#media-list .media-item div, #media-list .media-item select, #media-list .media-item img {
     display: inline-block;
     margin: 0 1em;
-}
+}*/
```
kareila: Should we remove these instead of leaving them commented out?

views/media/new.tt
```diff
@@ -29,6 +29,7 @@ the same terms as Perl itself.  For a copy of the license, please reference
     "stc/css/pages/media/new.css"
 ) -%]
 
+<p>[% dw.ml(".intro" {aopts1 => '/file/edit', aopts2 => '/file/new'}) %]</p>
```
kareila: I think these aopts are wrong - should be list and edit, not edit and new.

views/media/field-row.tt
```diff
@@ -0,0 +1,32 @@
+[% DEFAULT
+    labelSize = 2
+    fieldSize = 12 - labelSize
+    type      = "text"
+%]
+<div class="row">
+    <div class="large-[% labelSize %] columns">
+        <label for="[% name %]">[% label %]</label>
+    </div>
+    <div class="large-[% fieldSize %] columns">
```
kareila: Having the label and the form element in two different divs looks rather odd in my browser, and probably doesn't make sense from a semantic standpoint either (although that's not my area of expertise and I am open to counterarguments...)  But I do like that you made this a reusable element!

views/media/index.tt
```diff
@@ -9,24 +9,134 @@ This program is free software; you may redistribute it and/or modify it under
...
+function updateEmbed ( evt ) {
+    var thumbnail_value = evt.target.value;
+    var name = evt.target.id.replace(/thumbnail_sizes/, 'embed-thumb');
+    var url = document.getElementById(name).value;
+    document.getElementById(name).value = url.replace(/\d{3,4}x\d{3,4}/, thumbnail_value + 'x' + thumbnail_value);
+}
+
+window.onload = function() {
+  var dropdowns = document.getElementsByClassName('embed-select');
+  for (elem of dropdowns) {
```
kareila: I'm not sure exactly what this is trying to do, but it's giving me a JS error in Safari:

`TypeError: elem of dropdowns is not a function. (In 'elem of dropdowns', 'elem of dropdowns' is undefined)`

cgi-bin/DW/Controller/Media.pm
```diff
@@ -21,8 +21,9 @@ use warnings;
 use DW::Routing;
 use DW::Request;
 use DW::Controller;
+use POSXIX;
```
kareila: Typo - should be POSIX.

views/media/field-row.tt
kareila: To clarify, when I say "rather odd" I mean there's a lot of separation.  Or... maybe my browser was hanging onto a cached version of the CSS file, because when I refreshed a lot of the odd spacing disappeared.  But the Delete checkbox and its associated label are still separated by the entire width of the page, making both elements look orphaned.

cgi-bin/DW/Controller/Media.pm
```diff
@@ -68,14 +77,39 @@ sub media_bulkedit_handler {
...
-                my $media = DW::Media->new( user => $rv->{u}, mediaid => $mediaid );
+
+            # transform our HTML field names to property names
+            # and group by what media object they belong to
+            # we don't care if the id or prop might not exist
+            # right now, later steps will verify them
+
+            my %props;
+            while ( my ($key, $val) = each %post ) {
+                next if $key =~ m/delete/;
```
kareila: If this is exactly "delete" use `eq` instead?

cgi-bin/DW/Controller/Media.pm
```diff
@@ -68,14 +77,39 @@ sub media_bulkedit_handler {
...
+                next unless $key =~ m/^(\w+)-(\d+)/;
```
kareila: If it's "delete-id#", try `next if $1 eq delete` here?

views/media/edit.tt
```diff
@@ -9,43 +9,107 @@ This program is free software; you may redistribute it and/or modify it under
...
+                id = "edit-description-$obj.displayid" 
+                value = obj.prop('description') 
+                name = "description-$obj.displayid"
+            %]
+            [% INCLUDE 'media/field-row.tt' 
+                type = "checkbox"
+                label = dw.ml(".edit.delete") 
+                id = "delete_$obj.displayid" 
+                name = "delete-$obj.displayid"
+            %]
```
kareila: Ah ha, the old checkbox also had `value="[% obj.displayid %]` and I think that's what's causing both the undefined argument error and the failure to actually delete.

cgi-bin/DW/Controller/Media.pm
kareila: Yeah, I checked and you can change this to `next if $key eq 'delete';`

Never use regular expressions where simple equality checks will do, especially if they're not anchored :)

htdocs/stc/media.css
```diff
@@ -69,7 +69,7 @@
 
 .embed {padding: 0 1em;}
 
-@media screen and (min-width: 64em) {
+@media screen and (min-width: 64.01em) {
```
kareila: I think you also need to apply a similar adjustment to the labels that appear after uploading a new file - those are still floating above and to the right at widths less than 960 but more than 600.

kareila:
This is so close to being finished. The only other issue I'm seeing (besides the two I mentioned above) is a warning in the logs every time I upload an image, but all it says is "Use of uninitialized value."  No code description or file location or anything, which makes it just about impossible to track down.  Not sure if it's worth spending any time on.

zorkian:
OH YESSSSSSSSSS

PR #1962, opened 2017-02-09
Migrate userpicblob2 icons

kareila:
New version of bin/upgrading/migrate-userpics.pl that will pull images out of the userpicblob2 table and save them to BlobStore.  Doesn't migrate anything stored in MogileFS, that can be added later.


zorkian:
Nice use of log4perl :)

PR #1964, opened 2017-02-09
Blobstore-related cleanup

kareila:
* Indicate in checkconfig.pl that MogileFS::Client is now optional, and remove it from modperl_subs.pl.

* A couple of other minor checkconfig.pl issues.

* Fix incorrect file name in header of blobstore.t.

* Update various other textual MogileFS references, mostly code comments.

### DIFF
--- a/t/blobstore.t
+++ b/t/blobstore.t
@@ -1,4 +1,4 @@
-# t/media-security.t
+# t/blobstore.t
 #
 # Test some BlobStore functionality.
 #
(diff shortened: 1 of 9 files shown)


t/blobstore.t
```diff
@@ -1,4 +1,4 @@
-# t/media-security.t
+# t/blobstore.t
```
zorkian: My dirty cut and paste secrets revealed for all to see!

rahaeli:
if the clover+C or clover+V keys ever broke off my keyboard i'd never be able to code again ;)

PR #1967, opened 2017-02-09
allow etc/log4perl.conf to be superceded by ext/local

kareila:
Use LJ::resolve_file to prefer a version in ext/local
over the stock version in LJHOME, like we do for other
config files.


zorkian:
I totally forgot about that function. Whoops. Thanks!

PR #1969, opened 2017-02-09
delete associated thumbnails when a media image is deleted

### LINKED ISSUE #1931: Backend: deleting an image on /file/edit deletes the original but not the thumbnail

kareila:
This calls the parent method to do the deletion as before, then looks for other versions of the same image to also delete from storage once the original is gone.

While testing this, I realized that the relevant rows in media_versions are preserved, so we need to crosscheck against the 'state' column of the media table and make sure we aren't counting the sizes of deleted files against the user's quota.

Fixes #1931.

### DIFF
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -227,7 +227,10 @@ sub get_usage_for_user {
     confess 'Invalid user' unless LJ::isu( $u );
 
     my ( $usage ) = $u->selectrow_array(
-        q{SELECT SUM(filesize) FROM media_versions WHERE userid = ?},
+        q{SELECT SUM(mv.filesize) FROM media_versions AS mv, media AS m
+          WHERE mv.userid=? AND m.userid=mv.userid AND m.mediaid=mv.mediaid
+          AND m.state = 'A';
+         },
         undef, $u->id
     );
     croak 'Failed to get file sizes: ' . $u->errstr . '.' if $u->err;
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -229,7 +229,7 @@ sub get_usage_for_user {
     my ( $usage ) = $u->selectrow_array(
         q{SELECT SUM(mv.filesize) FROM media_versions AS mv, media AS m
           WHERE mv.userid=? AND m.userid=mv.userid AND m.mediaid=mv.mediaid
-          AND m.state = 'A';
+          AND m.state = 'A'
          },
         undef, $u->id
     );
--- a/cgi-bin/DW/Media/Photo.pm
+++ b/cgi-bin/DW/Media/Photo.pm
@@ -174,4 +174,45 @@ sub _select_version {
     # the resized image.
 }
 
+# this adds on to the base method by also deleting any associated thumbnails
+sub delete {
+    my $self = $_[0];
+    my $deleted = $self->SUPER::delete;
+    return 0 unless $deleted;  # was already deleted
+
+    # at this point the image has just been deleted - look for thumbnails
+    my $u = $self->u
+        or croak 'Sorry, unable to load the user.';
+    my @mv = $u->selectrow_array( "SELECT mediaid FROM media_versions" .
+                                  " WHERE userid=? AND versionid=?", undef,
+                                  $u->id, $self->versionid );
+    # this shouldn't happen unless something was saved in an incomplete state
+    # or the database has taken an unexpected vacation
+    return $deleted unless @mv;
+    my ( $mediaid ) = @mv;
+
+    if ( $mediaid == $self->versionid ) {
+        # find any resized versions and queue them for deletion as well
+        @mv = $u->selectrow_array(
+            "SELECT versionid FROM media_versions WHERE userid=? AND mediaid=?" .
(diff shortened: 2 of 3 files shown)


cgi-bin/DW/Media.pm
```diff
@@ -227,7 +227,10 @@ sub get_usage_for_user {
     confess 'Invalid user' unless LJ::isu( $u );
 
     my ( $usage ) = $u->selectrow_array(
-        q{SELECT SUM(filesize) FROM media_versions WHERE userid = ?},
+        q{SELECT SUM(mv.filesize) FROM media_versions AS mv, media AS m
+          WHERE mv.userid=? AND m.userid=mv.userid AND m.mediaid=mv.mediaid
+          AND m.state = 'A';
```
zorkian: No semicolons in the programmatic queries please.

cgi-bin/DW/Media/Photo.pm
```diff
@@ -174,4 +174,45 @@ sub _select_version {
...
+    return 0 unless $deleted;  # was already deleted
+
+    # at this point the image has just been deleted - look for thumbnails
+    my $u = $self->u
+        or croak 'Sorry, unable to load the user.';
+    my @mv = $u->selectrow_array( "SELECT mediaid FROM media_versions" .
+                                  " WHERE userid=? AND versionid=?", undef,
+                                  $u->id, $self->versionid );
+    # this shouldn't happen unless something was saved in an incomplete state
+    # or the database has taken an unexpected vacation
```
zorkian: It's pining for the fjords.

cgi-bin/DW/Media/Photo.pm
```diff
@@ -174,4 +174,45 @@ sub _select_version {
...
+    return $deleted unless @mv;
+    my ( $mediaid ) = @mv;
+
+    if ( $mediaid == $self->versionid ) {
```
zorkian: Prefer not to indent large sections of code. In this case, this would be nice:

```
    # not an original, so no further work
    return $deleted unless $mediaid == $self->versionid;

    # was an original, do work...
```

cgi-bin/DW/Media/Photo.pm
```diff
@@ -174,4 +174,45 @@ sub _select_version {
...
+        # find any resized versions and queue them for deletion as well
+        @mv = $u->selectrow_array(
```
zorkian: And actually this whole function could probably be simplified. Instead of the first select, you can do something like:

```
SELECT versionid FROM media_versions WHERE userid=? AND mediaid=? AND versionid != ?
```

In the case of a thumbnail, it'll return 0 rows, in the case of an original with no thumbnails, it'll return 0 rows, and for an original with thumbnails, it'll return the appropriate rows.

cgi-bin/DW/Media/Photo.pm
```diff
@@ -174,4 +174,45 @@ sub _select_version {
...
+            "SELECT versionid FROM media_versions WHERE userid=? AND mediaid=?" .
+            " AND mediaid != versionid", undef, $u->id, $mediaid );
+        return $deleted unless @mv;  # no other versions found
+
+        foreach my $id ( @mv ) {
+            # create a fake object to get the mogkey
+            my $fakeobj = bless { userid => $u->id, versionid => $id },
+                                'DW::Media::Photo';
+            # we aren't concerned whether the file existed or not
+            DW::BlobStore->delete( media => $fakeobj->mogkey );
```
zorkian: This doesn't end up marking the media row as 'D' but it deletes the underlying data, that seems inconsistent. Probably need to just call `delete` on the fake object? (Or might have to load a real object?)

kareila: Thumbnails don't have media rows, only originals do.

kareila:
Updated - thanks for the SQL help :nerd_face:

cgi-bin/DW/Media/Photo.pm
zorkian: But they have `media_versions` rows. Now those will dangle forever -- we'll have to clean them up sometime.

kareila: I'm confused. You said in the comments on the code that you wanted to keep those rows for record-keeping purposes.

zorkian: The `media` rows, yes, but there's no real metadata in a `media_versions` row.

The thinking with the former was that we could then always tell the user 'you used to have a private PDF uploaded at 5PM, but you deleted it'. I'm not entirely sure if that would ever be useful, but that was the idea.

The `media_versions` rows don't have a purpose like that -- they can't really tell you much other than what size the version was.

I'll approve it as-is because I don't imagine the traffic in media-we-resized-that-then-got-deleted will be high enough to ever matter, tho.

kareila:
At least the current behavior is consistent. I would suggest opening a new issue if we want to delete rows from that table in the future. Thanks for the feedback!

PR #1977, opened 2017-02-13
change LF to CRLF in graph labels to display properly

alierak:
GD interprets LF as a vertical tab, which it displays as a little V over T glyph.
That's weird. It wants CRLF instead.


zorkian:
`o_o;`

PR #1978, opened 2017-02-13
re-enable stats graph printing

kareila:
This early return prevented the "pretty graphs" from being printed, with a comment saying they were broken.  @alierak says they are not broken - looks like newbyday.png is generated by bin/maint/statspics.pl.  The early return dates from the original LJ import.

While we're here, only try to display the image if the image file exists, and English-strip the text.

### DIFF
--- a/htdocs/stats.bml
+++ b/htdocs/stats.bml
@@ -376,10 +376,6 @@ body<=
  }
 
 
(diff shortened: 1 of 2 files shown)


zorkian:
I suspect they were broken on LJ because of some changes ... lol, no, I don't remember anything from that far back. :(

PR #1980, opened 2017-02-13
fixes for media file access

kareila:
Use a controller to do basic setup checking, including bouncing for a user domain cookie if needed - LJ::get_remote was failing when the user domain cookie wasn't available.

With that working, we can add functional support for viewall, and remove the FIXME comment that requested it.

Also, as @kaberett pointed out, crossposted entries that reference hosted images will fail to load on the remote site, so let's whitelist all known crosspost sites.

### DIFF
--- a/cgi-bin/DW/Controller/Media.pm
+++ b/cgi-bin/DW/Controller/Media.pm
@@ -115,7 +115,7 @@ sub media_bulkedit_handler {
                     } else {
                         $media->prop( $key, $val );
                     }
-                }  
+                }
             }
 
         } elsif ( $post_args->{"action:delete"} ) {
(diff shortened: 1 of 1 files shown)


zorkian:
Mmm. Interesting point about the crossposting. I hadn't considered that. I suppose if we wanted to get really fancy we could ensure it's on the exact page, but this is probably good enough? (I mean, it means you can use DW to host any image for posting on LJ? There's abuse potential but probably not enough to worry about.)

alierak:
Well, exact page is kind of vague, it could be in that entry, on anybody's friends page.

zorkian:
My dreams of nice, clean systems :(

kareila:
If you wanted a nice, clean system, you shouldn't have based it on LJ 😆

zorkian:
As they say .. #rekt :)

PR #1982, opened 2017-02-14
[#1981] anchor regexp in break_word subroutine

### LINKED ISSUE #1981: HTML cleaner no longer autolinkifying URLs
I'm guessing this is related to the fix for #1948. See, for instance, https://dw-dev.dreamwidth.org/195668.html:

> url1;http://www.dreamwidth.org/file/edit&&urlend; works! url2;http://www.dreamwidth.org/file/edit&/&urlend; should also work. It didn't. Now it will. 

It's adding url1 at the beginning and &urlend; at the end, as well as a stray & in the middle, which is causing them to not auto-link and not even be copy/pastable.

kareila:
I committed the cardinal sin of not anchoring my regexp.  In my defense, the nearly-identical regexp in the original code wasn't exactly anchored either.

Fixes #1981.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1696,7 +1696,7 @@ sub break_word {
     # the last punctuation point; otherwise it will be placed at the maximum
     # length of the unbroken word as defined by $at.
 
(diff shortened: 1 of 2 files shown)


zorkian:
Ah, Cardinal sin.

PR #1983, opened 2017-02-14
fix uploading of non-factory, non-URL icons

kareila:
For the record, I questioned this line the first time I saw it.

I've tested editicons on my hack with this fix in place and all three upload flows (factory file, non-factory file, URL) seem to work fine now for me.

### DIFF
--- a/cgi-bin/DW/Controller/EditIcons.pm
+++ b/cgi-bin/DW/Controller/EditIcons.pm
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
 #
 # DW::Controller::EditIcons
-# 
+#
 # This controller is for creating and managing icons. NOTE: The actual file
 # is still a BML file, this is just a forward looking controller to help us
 # migrate.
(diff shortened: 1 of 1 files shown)


zorkian:
I don't know what to tell you. I tested it a bunch of ways and it didn't work. Maybe something insane in my code at the time, but I can't really give you a good review here. I believe I tried this and bad things happened. If you've tried this and did not get the same result, sure, go for it.

kareila:
In that case, since the page on the site is already broken, I'll try deploying this. If it makes things worse, I'll back it out immediately.

zorkian:
🐳

kareila:
WFM

kareila:
So I thought there were 3 paths but there were actually 4. I did not manage to test the parse_large_upload function, which doesn't work with this in place.  The eventual fix will be slightly more complex.

kareila:
OK, I think we've finally worked it out.  If this fix is wrong I don't want to be right.

zorkian:
Alas I can't provide coherent review. I spent a good 10 hours in this code but have no idea still how it's passing data around and what expects what. It's really a mess.

kareila:
I feel your pain!  I really want to rewrite the whole page now because it's SO BAD.

This is live on the site now.

alierak:
I am with this commit. I agree this is works but not can brain very much.

PR #1993, opened 2017-02-16
use correct user object when deleting existing slugs

### LINKED ISSUE #1992: entry loses slug on edit if poster and journal are on different clusters
http://www.dreamwidth.org/support/see_request?id=35808

In the part of editevent that deletes the current slug before adding the new one, it deletes from the poster's cluster instead of the journal's cluster, then if the slug didn't change, complains that the current one already exists and can't be reused.

kareila:
I can't test this on a single cluster setup, but the fix seems pretty obvious to me.

Fixes #1992.


zorkian:
Wherps.

PR #1994, opened 2017-02-16
[#1988] remove LJ/EventLogRecord.pm and submodules

### LINKED ISSUE #1988: remove null-op LJ/EventLogRecord.pm and submodules
I've determined that this isn't actually doing anything.  If you look at the most recently available LJ version of the module, you'll see a key difference: after the fork, LJ added EventLogSink methods which defined the actions to take.

Our routine, abbreviated:
```perl
sub work {
    my $evt_class = delete $params{_event_class} or die "No event_class specified";
    my $evt = LJ::EventLogRecord::new($evt_class, %params);
    $job->completed;
}
```
Versus theirs, also abbreviated:
```perl
sub work {
    my $evt_class = delete $params{_event_class} or die "No event_class specified";
    my $evt = LJ::EventLogRecord::new($evt_class, %params);

    foreach my $sink (LJ::EventLogSink->sinks) {
	$sink->log($evt) if $sink->should_log($evt);
    }

    $job->completed;
}
```

It seems unlikely we would ever want to finish implementing this, so let's rip it out.

kareila:
I thought this was going to be a simple (albeit really satisfying) machete job, but things got weird when I noticed one of the automated tests was suddenly failing.  Details in commit log.

Fixes #1988.


cgi-bin/LJ/Protocol.pm
```diff
@@ -1757,6 +1754,11 @@ sub postevent
         # TODO: error on failure?  depends on the job I suppose?  property of the job?
     }
 
+    # To minimize impact on legacy code, let's make sure the entry object in
+    # memory has been populated with data. Easiest way to do that is to call
+    # one of the methods that loads the relevant row from the database.
+    $entry->valid;
```
zorkian: I'll merge it but this feels vaguely magical, which it is, and like we might get bitten at some other point. I like the fix of switching from reaching into the object's members to using the accessor methods. Maybe one step further is having the accessor methods know how to 'instantiate' the right internal data if they're called for the first time and it hasn't been done yet? Does that make sense?

kareila: Absolutely, the methods already do that, and if we used the methods everywhere, I wouldn't worry about it! But legacy stuff often uses hashrefs instead of methods - we saw the same thing with user objects early on.

PR #2003, opened 2017-02-22
additional headers for media file requests

kareila:
Discussion with @alierak led to the conclusion that media files should use the Cache-Control and Last-Modified headers.

Inspection of Apache/LiveJournal.pm and Apache/BML.pm indicates that Cache-Control should be applied to all user content responses, so we make that the new default behavior for user content in DW/Routing.pm.

The Last-Modified header is set in the media file controller as the upload date of the file, which will never change.

Without the Last-Modified header, the image may be redownloaded unnecessarily.  Without the Cache-Control header, Cloudflare caching may interfere with the ability to load the image correctly.

### DIFF
--- a/cgi-bin/DW/Controller/Media.pm
+++ b/cgi-bin/DW/Controller/Media.pm
@@ -233,6 +233,7 @@ sub media_handler {
         unless defined $dataref && ref $dataref eq 'SCALAR';
 
     # now we're done!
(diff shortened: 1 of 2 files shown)


zorkian:
Verified and agree with your assessment of Apache::LiveJournal behavior in this case.

PR #2004, opened 2017-02-22
[#1990] don't transform username tags in URLs

### LINKED ISSUE #1990: markdown parser shouldn't transform user tags inside hrefs
http://www.dreamwidth.org/support/see_request?id=35804

Someone needs to visit the clean_as_markdown subroutine and tell it to ignore "@username" text that happens inside hrefs, because it is mangling URLs from sites like medium.com.  This may be beyond the capability of a simple regular expression.

Example of a valid URL that the Markdown parser breaks is in the linked support request.

kareila:
I thought about bringing out the heavy artillery (aka HTML::Parser) but realized that even if we just transformed on plain text, things that look like usernames in plain URLs would still be mangled.  So this just tweaks the existing regular expression to ignore `@username` if immediately preceded by a slash.

Fixes #1990.

### DIFF
--- a/t/cleaner-markdown.t
+++ b/t/cleaner-markdown.t
@@ -1,0 +1,53 @@
+# t/cleaner-markdown.t
+#
+# Test LJ::CleanHTML with Markdown text.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
+#
+# Copyright (c) 2017 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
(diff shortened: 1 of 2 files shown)


t/cleaner-markdown.t
```diff
@@ -0,0 +1,53 @@
+# t/cleaner-markdown.t
+#
+# Test LJ::CleanHTML with Markdown text.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
```
zorkian: I still love that you use an @livejournal.com address here.

zorkian:
Thanks for not busting out the heavy artillery! The system as-is is bad enough... :)

PR #2007, opened 2017-02-24
[#2003] also use error header for Cache-Control

kareila:
@alierak explains that there are two header tables, one for successes (which header_out uses) and one for errors and redirects (which err_header_out uses).  So we need to make sure that user cookie bounces use err_header_out to avoid caching.

He says we could probably set this header on every request, but let's start with the one place where we know we need it.

(Most of this is whitespace cleanup; the last line is the bit that does the thing.)

### DIFF
--- a/cgi-bin/DW/Controller.pm
+++ b/cgi-bin/DW/Controller.pm
@@ -73,8 +73,8 @@ sub render_success {
 # Supported arguments: (1 stands for any true value, 0 for any false value)
 # - anonymous => 1 -- lets anonymous (not logged in) visitors view the page
 # - anonymous => 0 -- doesn't (default)
-# - authas => 1  or { args } -- allows ?authas= in URL, generates authas form (not permitted
-#                  if anonymous => 1 specified)
+# - authas => 1  or { args } -- allows ?authas= in URL, generates authas form
+#                              (not permitted if anonymous => 1 specified)
 # - authas => 0 -- doesn't (default)
 # - specify_user => 1 -- allows ?user= in URL (Note: requesting both authas and
 #                        specify_user is allowed, but probably not a good idea)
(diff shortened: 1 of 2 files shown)


alierak:
Two places we know we need it. I don't think this addresses the potential for caching a 403 permission denied response when trying to retrieve an access-locked image anonymously.

kareila:
Heh, yeah, I forgot about that aspect of it in my annoyance at the infinite redirects.  Change made, but I can't test the effectiveness of these outside prod.

alierak:
Well, to illustrate the perversity of Apache's dual header tables, I think in that particular code path header_out is the one that would be used because the sub returns $r->OK (although the status code is set to something other than 200 with $r->status). In production, now, the 403 error has a Cache-Control header while the 404 does not (it returns earlier with $r->NOT_FOUND).

kareila:
I didn't want to use the no-cache header for the 404 errors, they all looked like permanent failures, not transient ones... but I did wonder about whether we should be returning $r->OK there?

alierak:
Well, I guess you don't know exactly what method to use for whatever status code the caller supplied so OK should be fine, but it apparently pulls from the other header table and you'll get Cache-Control: private, proxy-revalidate.

kareila:
Well, that means the previous change is working then? Is that sufficient for our purposes, or do we want to make sure it's "no-cache"?

alierak:
I, uh, well, it's sufficient but not correct. I guess just worry about the redirects for now.

kareila:
Okay Professor, I changed it from err_header_out to header_out.  Do I get an A this time?

alierak:
That ought to do it.

rahaeli:
you two are adorable sometimes

kareila:
Only sometimes?

OK, since I have thumbs up from @alierak I'm going to push this live and see what happens.

zorkian:
I was about to say, ‘only sometimes’?

Issue #1997, opened 2017-02-20
add Medium to profile other-sites block


rahaeli:
Given a user "username", the URL for the name link would be https://medium.com/@username/latest and the link for the profile link would be https://medium.com/@username. 

For the profile, we can linkify the username provided as https://medium.com/@username/latest

For the userhead image: hmm, I guess let's take https://medium.com/favicon.ico and turn it into a 16x16 png or something.

srukle:
Claim

srukle:
@rahaeli Should we link to the latest posts or the profile? The spec seems to want latest, which is a column in the user's profile. I think we should link to just the profile. I also have a thread in #2008.

srukle:
Shouldn't this be closed now?

alierak:
Github won't auto-close the issue unless a pull request that is properly linked to the issue gets merged and closed. I think this requires something along the lines of a commit message in the pull request containing "Fixes #issue-number", but then I've screwed it up myself... Oh, there we go, https://help.github.com/articles/closing-issues-using-keywords/

In this case, neither of your pull requests (2008 nor 2110) contained such a keyword, and neither has been merged. @kareila's pull request (2145) has been merged, but implemented medium support in the user tag only, so I believe she chose to reference this issue but not close it at that point.

This issue should stay open until dreamwidth/dw-free supports medium in both the profile and the user tag.

srukle:
Ok, I'll rebase and send another PR tonight.

kareila:
Unassigning this since #2110 was closed, but feel free to pick it back up later!

Also I will update the title of this issue to just request the profile listing, since the user tag was done in #2145.

kareila:
I just noticed that the user tag did get done already, so this is all done now!

PR #2008, opened 2017-02-25
add medium

srukle:
Adds the option to list a Medium account under your other services header of the Dreamwidth profile.

Addresses issue #1997


cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -851,6 +851,17 @@ sub external_services {
         };
     }
 
+    if ( my $medium = $u->prop( 'medium' ) ) {
+	my $medium = LJ::eurl( $medium );
```
zorkian: This should be `my $emedium = ...` so as not to overwrite `$medium`. Else, you end up with the content being double-escaped: eurl'd and then ehtml'd in the `text` value later.

So, change this one to `my $emedium ...` and then later use `$emedium` in the `url` value.

htdocs/manage/profile/index.bml
```diff
@@ -431,6 +431,7 @@ body<=
                            ["icq", $ML{'.chat.icquin'}, 12],
                            ["jabber", $ML{'.chat.jabber'}, 60],
                            ["last_fm_user", $ML{'.services.last_fm'}, 255],
+			   ["medium", $ML{'services.medium'}, 40],
```
zorkian: Please use spaces like the other lines -- it looks like you've got tabs here.

cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -851,6 +851,17 @@ sub external_services {
...
+	push @ret, {
+	    type => 'medium',
+	    text => LJ::ehtml( $medium ),
+	    url => "https://medium.com/\@$medium",
```
kareila: According to the spec in the issue text, I believe this URL should end with /latest?

srukle: Thanks for the feedback. The URL `. . . /latest` opens the _Latest_ column of the profile. Though this was in the original request, I believe having a profile link point to a profile is the intent. Latest entrees will appear here along with any stickied or featured posts, depending on the user's preference.

If we are sticking with the specs, should I change "Medium user account" to "Medium user posts"? I ask -- so that we don't end up with conflicting information in the future.

kareila: I'm fine with either choice. @rahaeli do you have an opinion?

htdocs/manage/profile/index.bml
srukle: Thanks for the feedback. I will make spaces here instead of tabs.

cgi-bin/DW/Logic/ProfilePage.pm
srukle: Thanks for the feedback. I will make the changes.

cgi-bin/DW/Logic/ProfilePage.pm
rahaeli: I'm okay with either, too -- I don't  use Medium enough to say and just copied what the original person who asked for it to be added said.

srukle:
@zorkian I've made the changes to PR.

srukle:
The merge conflict seemed to be due to the removal of mailencoding. I took the current development code and put it into my commit. I hope this will resolve the merge conflict.

PR #2015, opened 2017-02-28
Api framework

rshatch:
Provides the basics of an OpenAPI-compatible framework along with handlers for the following paths:
/api/v1/users/{username}/icons
/api/v1/users/{username}/icons/{picid}
/api/v1/journals/{journal}/entries
/api/v1/journals/{journal}/entries/{entry_id}
/api/v1/moods
/api/v1/spec


rshatch:
Yeah, I can split out the core stuff and some of the most basic API calls pretty easily. I agree that we probably don't want to allow anon access - I think I was maybe working off the routing controllers that were in Deborah and Allen's work and not thinking it through? I think it would be better to properly implement api keys and move that auth checking to somewhere more global, rather than each endpoint implementing it, though that may take me a little more time to sort out.

zorkian:
SGTM. Let's split it out and for today we'll leave the auth as-is so I
can review it and get it in! :)

PR #2017, opened 2017-03-01
[#2016] look for both possible saved orderings

### LINKED ISSUE #2016: log_event expects 'keys' to return a consistent order
This was fun to track down:
```
[Error: This event (uid=4, extra=new=V&old=D) was not found in logs at /dreamhack/home/8398-kareila/dw/cgi-bin/LJ/Event/SecurityAttributeChanged.pm line 47. @ newhack.dreamwidth.net]
```
This can happen whenever you delete or undelete an account.  The error also gets printed to the browser in lieu of a normal page, which is very sad.

I checked my database, and the actual value for that event was `(uid=4, extra=old=D&new=V)`. Hmm...

The `set_statusvis` user method calls `log_event` which uses a join on the return value of `keys`, which as all Perl lovers know USED TO return a consistently ordered array but no longer does so.

After logging the event, the page immediately tries to fire LJ::Event::SecurityAttributeChanged, which does this:
```perl
        my $extra = (1 == $action) ? 'new=D&old=V' : 'new=V&old=D';

        my $dbcr = LJ::get_cluster_reader($u);
        my $sth = $dbcr->prepare(
            "SELECT logtime, ip".
            " FROM userlog".
            " WHERE userid=? AND extra=?".
            " ORDER BY logtime DESC LIMIT 2");
        $sth->execute($u->{userid},$extra);
```
If the row isn't found, everyone dies, oh the embarrassment.

Other events that look up the value saved in this column use `LJ::decode_url_string` to decode the value in an order-agnostic way, but the easiest fix in this case is probably to search for either 'old=D&new=V' or 'new=V&old=D' (or vice versa).

kareila:
I can confirm that this lookup works in testing with unsorted data. I'm also adding a sort to log_event so prevent similar problems in the future, but that won't correct previously logged events.

Fixes #2016.

### DIFF
--- a/cgi-bin/LJ/Event/SecurityAttributeChanged.pm
+++ b/cgi-bin/LJ/Event/SecurityAttributeChanged.pm
@@ -31,16 +31,19 @@ sub new {
 
         die "Missing credentials" unless $ip && $action;
 
+        # We can't be sure which order the keys were saved in,
+        # so search for both possibilities (old/new or new/old).
+
         # $action == 1 -- deleted
-        my $extra = (1 == $action) ? 'new=D&old=V' : 'new=V&old=D';
+        my $extra = (1 == $action) ? "'old=V&new=D', 'new=D&old=V'"
+                                   : "'old=D&new=V', 'new=V&old=D'";
 
         my $dbcr = LJ::get_cluster_reader($u);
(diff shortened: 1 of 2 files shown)


zorkian:
I find this vaguely distressing, but agree with the result.

PR #2019, opened 2017-03-03
convert manage/index.bml to TT

kareila:
This doesn't use Foundation styling, just gets rid of the ugly BML coding blocks and layout tables, and updates some display text.  Also adds a link to /file/edit for personal accounts.


views/manage/index.tt
```diff
@@ -0,0 +1,305 @@
...
+          padding: 0.5em;
+        }
+    </style>
+[% END %]
+
+<form method="get">[% authas_html %]</form>
+
+[% iscomm = u.is_community ? '.comm' : '' %]
+
+<h1>[% '.youraccount.header' _ iscomm | ml %]</h1>
```
zorkian: Somehow I never knew that `_` is how you do string concatenation in TT?

kareila: Because `.` is reserved for object methods (in lieu of `->`)

I read somewhere (probably TT docs) that Perl 6 also decided to make the same change.

PR #2031, opened 2017-03-15
Remove unused code for browselang

kareila:
This gets rid of the `browselang` userprop and associated unused module `LJ/Setting/Language.pm`, as well as the site cookie for langpref.  We've never supported users choosing a persistent preference for a language different from the site's default language.

This is a larger set of changes than the other similar ones I've submitted recently, and the individual commit messages are informative.  I've lightly tested this by running the automated test suite, making sure "uselang=debug" still worked as expected, and testing sample event output using the /admin/eventoutput page.


rahaeli:
WOOHOO CODE MACHETE

Just checking out of a sense of paranoia (since I can't see why it would), but does this break the ability for logged out people to set a site skin preference?

kareila:
It shouldn't! There was a language cookie and a scheme cookie and I only removed the language one.

rahaeli:
I figured! But better safe than sorry, etc.

zorkian:
Thank you for the helpful commit messages.

PR #2040, opened 2017-03-20
[#2039] check for extra space in bogus_email address text

### LINKED ISSUE #2039: DW::CleanEmail::nonquoted_text not removing all quoted text
https://www.dreamwidth.org/support/see_request?id=35649
https://www.dreamwidth.org/support/see_request?id=35913

The way the function works is this: first it removes all the quoted text, then it looks back through the last three lines of non-quoted text to see if anything looks like the standard "On [date], [person] wrote:" and removes that and anything under it.

The problem is that three lines doesn't seem to be enough in some circumstances.  Example A: http://azurelunatic.dreamwidth.org/7718044.html?thread=15395740#cmt15395740 which seems to be the minimal failure case of 4 lines.  Example B: https://bethany-lauren.dreamwidth.org/489786.html?thread=1209914#cmt1209914 which looks like the entire original body was retransmitted without any quoting, and the quote header is dozens of lines back.

Are there any downsides to ignoring the `$max_backtrack` limit and searching the entire block of non-quoted text for a line that matches the header pattern, or at least for the most common one?  Or should we just bump up `$max_backtrack` slightly and tell the more pathological offenders to make sure that the original message is edited out before sending?

kareila:
Fixes #2039.

### DIFF
--- a/cgi-bin/DW/CleanEmail.pm
+++ b/cgi-bin/DW/CleanEmail.pm
@@ -42,7 +42,7 @@ sub nonquoted_text {
         last if m/^\s*-{3,}[^-]+-{3,}\s*$/;
 
         # the bogus email we sent the comments as wrapped in <>
-        last if m/<$LJ::BOGUS_EMAIL>/;
+        last if m/<\s?$LJ::BOGUS_EMAIL>/;
 
         $num_lines++;
     }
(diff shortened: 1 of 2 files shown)


zorkian:
Fascinating... the dark art of email parsing. It's almost NLP. :)

PR #2042, opened 2017-03-21
[#2034] use %LJ::KNOWN_HTTPS_SITES for simple whitelist

### LINKED ISSUE #2034: HTTPS_UPGRADE_REGEX needs modification
1. photobucket.com should be removed because their HTTPS support is broken.

2. imgur.com is not being matched because the current regex assumes the domain name is preceded by a subdomain name (e.g. www).  This is probably affecting xkcd.com as well.

3. The second most popular proxy domain (after LJ) is abload.de, which does support HTTPS and should be added.  Similarly tumblr.com (no.4), flickr.com (no.8), blogspot.com (no.10), feedburner.com (no.12), plurk.com (no.13), imageshack.us (no.14), levkonoe.com (no.15), staticflickr.com (no.17), and wikimedia.org (no.18).

Perhaps this should be a hash of approved domains instead of an increasingly unwieldy regular expression?

kareila:
Replaces regular expression in config with a list of domains that support HTTPS connections, and tests the domain of the given URL against that whitelist.

Also updates https-url.t with a working dummy proxy config.

Fixes #2034.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1670,9 +1670,11 @@ sub https_url {
     # https:// and the relative // protocol don't need proxying
     return $url if $url =~ m!^(?:https://|//)!;
 
-    # if this link is on our site, let's just switch it to https
-    if ( $url =~ $LJ::DOMAIN || defined $LJ::HTTPS_UPGRADE_REGEX
-                             && $url =~ $LJ::HTTPS_UPGRADE_REGEX ) {
+    # if this link is on a site that supports HTTPS, upgrade the protocol
+    my $https_ok = %LJ::KNOWN_HTTPS_SITES ? \%LJ::KNOWN_HTTPS_SITES : {};
+    my ( $domain ) = ( $url =~ m!://[^/]*?([^.]+\.\w{2,3})/! );
+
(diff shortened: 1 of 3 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1670,9 +1670,11 @@ sub https_url {
     # https:// and the relative // protocol don't need proxying
     return $url if $url =~ m!^(?:https://|//)!;
 
-    # if this link is on our site, let's just switch it to https
-    if ( $url =~ $LJ::DOMAIN || defined $LJ::HTTPS_UPGRADE_REGEX
-                             && $url =~ $LJ::HTTPS_UPGRADE_REGEX ) {
+    # if this link is on a site that supports HTTPS, upgrade the protocol
+    my $https_ok = %LJ::KNOWN_HTTPS_SITES ? \%LJ::KNOWN_HTTPS_SITES : {};
+    my ( $domain ) = ( $url =~ m!://[^/]*?([^.]+\.\w{2,3})/! );
```
alierak: Not a huge fan of the unanchored regexp with non-greedy "*?" match. I think at that point it could be anchored at the beginning with "^http://" since we're only trying to upgrade http. Maybe a two-step extraction of the domain (see DW::External::UserInfo::parse_domain for an implementation that looks better).

kareila: ... which I also wrote XD

PR #2049, opened 2017-03-30
Add new RP account flag

### LINKED ISSUE #2048: add "this is a RP account" settings flag
https://dw-suggestions.dreamwidth.org/1494165.html

Add a way for people to flag their accounts as RP accounts, that will display on the profile

rahaeli:
Fixes #2048.

Adds a new "rp account" setting; toggling it prints a blurb on the profile
saying that the account is an RP account, and in the future may control
access to features that only make sense for RP accounts.


htdocs/profile.bml
```diff
@@ -396,7 +396,11 @@ body<=
                           style => 'vertical-align: middle; border: 0;' } );
         $ret .= "</a>\n";
     }
+
+    if ( $u->is_rp_account ) {
+        $ret .= ": " . BML::ml( '.details.rp' );
     $ret .= q{ </div> };
```
zorkian: I think this line has to move to after the next line -- we want to keep the `</div>` whether or not the user is flagged as an RP account.

rahaeli: Oops! Fixed

zorkian:
AMAZING GOLD STAR 🗡

zorkian:
TIL that colon-D autocompletes to 🗡 ...

kareila:
🌟

azurelunatic:
Is this a feature that would make sense reflecting in the stats?

rahaeli:
Possibly in the future, yes!

PR #2057, opened 2017-03-30
Widen userpic2 column 'description' to allow longer descriptions

### LINKED ISSUE #2055: increase size for icon description field
https://dw-suggestions.dreamwidth.org/1469860.html

Suggested by: alucinari
Description: Hello! I have friends that use screen readers, and a problem I have found is that when I am trying to describe icons, especially if it's a text icon with a longer quote on it, there's not enough room for me to properly describe what's going on; I either have to leave a lackluster description, or resort to chat speak, and sometimes even that doesn't really do the trick. It would be really helpful if DreamWidth allowed longer icon descriptions! It should be easy to allow more characters, and there shouldn't be any drawbacks, I don't think? Thank you for your consideration.

In the comments, lightgetsin opined that the description field should be "Longer than a tweet, shorter than a paragraph", and suggests 250 characters; let's be a bit more generous and go for 300.

rahaeli:
Fixes #2055.

Converts the description column of the userpic2 table from varchar(255) (which lets descriptions be 120 characters long to account for bytes, yadda) to varchar(600), which allows for 300 characters.

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -599,7 +599,7 @@ CREATE TABLE userpic2 (
     picdate datetime default NULL,
     md5base64 char(22) NOT NULL default '',
     comment varchar(255) BINARY NOT NULL default '',
-    description varchar(255) BINARY NOT NULL default '',
+    description varchar(600) BINARY NOT NULL default '',
     flags tinyint(1) unsigned NOT NULL default 0,
     location enum('blob','disk','mogile','blobstore') default NULL,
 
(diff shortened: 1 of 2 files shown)


bin/upgrading/update-db-general.pl
```diff
@@ -4093,6 +4093,12 @@
         );
     }
 
+    # widen the description field for userpics
+    if ( column_type( 'userpic2', 'description' ) eq "varchar(255)" ) {
+        do_alter( 'userpic2',
+            "ALTER TABLE userpic2 MODIFY COLUMN description VARCHAR(600) DEFAULT NULL");
```
zorkian: This also changes the default nullability, so this should be:

```
ALTER TABLE userpic2 MODIFY COLUMN description VARCHAR(600) BINARY NOT NULL default ''
```

rahaeli:
Fixed and updated! (Ah, development via copy/paste.)

rahaeli:
Forgot to remove the requested-changes label.

zorkian:
Ah-hah! Yeah, that's important :D

PR #2063, opened 2017-03-31
better/more helpful text on "not authorized for this filter" 403 error

### LINKED ISSUE #2061: better error message for link to private list while logged out
https://dw-suggestions.dreamwidth.org/601682.html

Suggested by: msilverstar

Summary:
I sometimes get logged out by accident, that's fine, I understand. But the error message is user-hostile and very likely to fluster less experienced users.

Description:
I have a named reading filter named "v" that I generally use to avoid big feeds and such. When I get logged out and try to reload the page, I get the following error message:

Forbidden

You don't have permission to access /read/v on this server.
Apache/2.2.11 (Ubuntu) mod_apreq2-20051231/2.6.0 mod_perl/2.0.4 Perl/v5.10.0 Server at msilverstar.dreamwidth.org Port 80

It would be eversomuch nicer to see something like this:

Error: you can't see filters because you aren't the logged user for this account

(with an option to log in)

Notes: Other things around this particular error in cgi-bin/Apache/LiveJournal.pm have meaningful/site-skinned errors instead of a bare Apache error page, so just following those examples is fine! A site-skinned error message will still give people the chance to log in.

To avoid an enumeration attack, this should give the same error message whether the filter exists or not, and whether the remote user is logged in or not.

rahaeli:
Fixes #2061.

Alters cgi-bin/Apache/LiveJournal.pm to change the Apache-default bare-unstyled 403 in the event of a non-public or non-existant reading page filter to a still-bare-unstyled page that at least has a link to the login page so people don't get stranded.

This has specifically been checked for enumeration attacks in the following situations:

* Specified filter does exist but logged-in user isn't authorized to see it;
* Specified filter does not exist, user is logged in;
* Specified filter does exist but is not public/remote not logged in;
* Specified filter does not exist user is not logged in

The error message is the same in all four scenarios, so people won't be able to use this to fish for filters that do exist but that they don't have access to.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -1405,7 +1405,8 @@ sub journal_content
 
         # otherwise be vague with a 403
         } else {
-            return 403;
+            $status = "403 Forbidden";
+            $html = "<h1>Invalid Filter</h1><p>Either this reading filter doesn't exist or you are not authorized to view it. Try <a href='$LJ::SITEROOT/login'>checking that you are logged in</a> if you're sure you have the name right.</p>";
         }
 
         $generate_iejunk = 1;


zorkian:
LGTM, nice find :)

PR #2064, opened 2017-04-01
Convert /editicons and related tools to DW::Controller & TT

### LINKED ISSUE #1966: finish converting htdocs/editicons.bml
@zorkian created DW::Controller::EditIcons and began moving functions into it, but the page itself still needs to be converted. We could also include htdocs/tools/userpicfactory.bml in the same controller module since it provides related functionality.

kareila:
I can't be 100% sure that I haven't introduced any new bugs, and I didn't test every possible failure mode because there are dozens, but I am reasonably confident that this is ready for use.  I discovered and addressed a few existing problems with the page, which I tried to document in separate commits.

Fixes #1966.


zorkian:
I am ... reasonably ... confident ... this is code.

I did read through it but like you know.

PR #2068, opened 2017-04-04
[#2046] print warning and skip compass step if using old version

### LINKED ISSUE #2046: build_static.sh should detect & diagnose old versions of Compass
Was updating my dev VM yesterday for the first time in yonks, and got these errors while trying to compile the SCSS:

* Building SCSS...
\    error htdocs/scss/components/tables-as-list.scss (Line 20: Invalid CSS after "@media ": expected "{", was "#{$small} {")
Sass::SyntaxError on line 20 of /home/dw/dw/htdocs/scss/components/tables-as-list.scss: Invalid CSS after "@media ": expected "{", was "#{$small} {"
Run with --trace to see the full backtrace
identical htdocs/stc/css/skins/tropo/homepage.css 
    error htdocs/scss/skins/tropo/tropo-red.scss (Line 71: Invalid CSS after "@media ": expected "{", was "#{$small-only} {")
Sass::SyntaxError on line 71 of /home/dw/dw/ext/dw-nonfree/htdocs/scss/skins/tropo/_tropo-base.scss: Invalid CSS after "@media ": expected "{", was "#{$small-only} {"

Mark suggested the problem was likely an outdated version of Compass (it was) and a gem update fixed the compile errors, but it would be nice if the script could identify that was what was going on and prompt people to upgrade their Compass version.

kareila:
My development laptop has a version string starting with 0.12, so I arbitrarily chose that as the minimum viable version.

I ran into a problem, however, when checking to see what version of Compass was installed on the admin box; I got an error message saying that VERSION.yml was missing, so this change will likely break the script in production until that situation is remedied.

Fixes #2046.

### DIFF
--- a/bin/build-static.sh
+++ b/bin/build-static.sh
@@ -20,13 +20,22 @@ fi
 # if compass is installed, build that first
 compass=$(which compass)
 if [ "$compass" != "" ]; then
-    echo "* Building SCSS..."
-    cd $LJHOME
-    $compass compile -e production --force
-    if [ -d "$LJHOME/ext/dw-nonfree" ]; then
-        cd $LJHOME/ext/dw-nonfree
+    # see if we have Compass version 0.12 or later
+    compass_version_ok=$(compass version | perl -ne '/^Compass (\d\.\d+)/ && print $1 >= 0.12')
+    if [ $compass_version_ok ]; then
+        echo "* Building SCSS..."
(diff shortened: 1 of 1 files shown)


kareila:
I fixed the problem in production by adding a bogus VERSION.yml file, so this shouldn't break anything now.

PR #2069, opened 2017-04-05
fix code conversion bugs

### LINKED ISSUE #2041: can't import icons from LJ: code error
http://www.dreamwidth.org/support/see_request?id=35940
http://www.dreamwidth.org/support/see_request?id=35943

Both requests quote the same error message:

`Failure running job: Store requires data be a scalar reference. at /home/dw/production/cgi-bin/DW/Worker/ContentImporter/LiveJournal/Userpics.pm line 103. `

kareila:
Fixes #2041: DW::BlobStore->store was being given `$data` when it wanted `\$data`.

Also fixes an issue with printing icon keywords on the newly converted editicons page which was revealed during testing.  (All of my previous test icons had single keywords, and the bug only happens when an icon has multiple keywords.)

### DIFF
--- a/views/edit/icons.tt
+++ b/views/edit/icons.tt
@@ -15,6 +15,8 @@
 
 [%- dw.need_res( "stc/editicons.css", "js/editicons.js" ) -%]
 
+[%- USE Scalar; # needed to force scalar return context for keywords -%]
+
 <form method="get" id='userpic_authas' action='editicons'>[% authas_html %]</form>
 
 [%- INCLUDE components/errors.tt errors = errors -%]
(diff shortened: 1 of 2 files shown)


views/edit/icons.tt
```diff
@@ -15,6 +15,8 @@
 
 [%- dw.need_res( "stc/editicons.css", "js/editicons.js" ) -%]
 
+[%- USE Scalar; # needed to force scalar return context for keywords -%]
```
zorkian: TIL.

zorkian:
Thanks for finding/fixing the BlobStore issue.

PR #2075, opened 2017-04-07
Code machete marathon

kareila:
Ripping out unused assets of various types.


zorkian:
RIP SPUD we knew ye only barely...

rahaeli:
I don't even remember what spud was supposed to do, heh

PR #2084, opened 2017-04-09
[#1921] on import, suppress tag creation errors and bogus tag creation

### LINKED ISSUE #1921: Suppress tag creation errors on community import
http://www.dreamwidth.org/support/see_request?id=35485

There exists a non-DW community for which creation of new tags is limited to maintainers only. Community members who are not aware of this try to tag their entries anyway. The non-canonical tags are not displayed in the community, but are recorded in the post's metadata, possibly via serving the user an error about how these tags do not exist.

Some of these communities will then be imported to Dreamwidth. If the Dreamwidth community is also configured such that only maintainers/admins can create new tags, the person who kicked off the import will get notified about every single post that has non-canonical tags _because they are considered the poster by Dreamwidth_. This is Not Great.

Discussion with @zorkian concluded that probably the best thing to do is just suppress these errors, so the importer (who never saw the tags to begin with!) doesn't get spammed with baffling error notifications.

kareila:
This skips warnings for tags in `@unauthorized_add` when importing, similar to what was already being done for deletion errors in the next line.

It also addresses Azz's concern about importing bogus tags, by not creating tags on entry import that weren't already added by the import of all current tags, which is forced to happen first.

This is lightly tested under normal use - I haven't tried to recreate the exact import setup needed to reproduce the reported behavior.

Fixes #1921.

### DIFF
--- a/cgi-bin/LJ/Tags.pm
+++ b/cgi-bin/LJ/Tags.pm
@@ -772,20 +772,26 @@ sub update_logtags {
         $opts->{"${verb}_ids"} ||= [];
         foreach my $kw (@{$opts->{$verb} || []}) {
             my $kwid = $u->get_keyword_id( $kw, $can_control );
-            if ($can_control) {
-                # error if we failed to create
-                return undef unless $kwid;
-            } else {
-                # if we're not creating, who cares, just skip; also skip if the keyword
-                # is not really a tag (don't promote it)
-                unless ( $kwid && $utags->{$kwid} ) {
+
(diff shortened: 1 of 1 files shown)


azurelunatic:
*\o/*

zorkian:
I think I'm convinced by the comment/logic.

rahaeli:
I'm sure as soon as it goes live, we'll find the eight million edge cases.

kareila:
:sweat_smile:

PR #2087, opened 2017-04-10
[#2086] update list of valid TLDs

### LINKED ISSUE #2086: Support for e-mail TLDs needs expanding
Per https://www.dreamwidth.org/support/see_request?id=36226: a valid e-mail address at a .tech domain was rejected during sign-up.

Having followed check_email via valid_email_domains via load_include, I suspect the thing that needs editing is https://github.com/dreamwidth/dw-free/blob/3467339cd823c6d71b8a235bf76409d5dab93b85/htdocs/inc/tlds.

kareila:
Fixes #2086.


zorkian:
The TLDs now mean that this being a list is useless, we should probably accept any string? heh.

PR #2088, opened 2017-04-10
[#1984] proxy insecure resources loaded from journal styles

### LINKED ISSUE #1984: investigate image proxy use for background images in journal layouts
https://dw-maintenance.dreamwidth.org/74385.html?thread=2614161#cmt2614161 says they're getting a mixed content warning over images from Photobucket. Journal stylesheet looks like they're replacing some of the icons with versions from imgur loaded via http (and might also have mood theme icons loaded from elsewhere but none of their entries are public so I can't confirm). 

May be nothing we can do about this (and if so, the answer would be "change your CSS and mood theme to use a SSL version") but is that something we can run through the proxy?

kareila:
This defines a css_cleaner_transform hook that looks for style text of the form "url(...)" and tries to upgrade the specified URL to HTTPS if the page is using SSL.

Fixes #1984.

### DIFF
--- a/cgi-bin/DW/Hooks/ProxyCSSLinks.pm
+++ b/cgi-bin/DW/Hooks/ProxyCSSLinks.pm
@@ -1,0 +1,35 @@
+#
+# Use this hook to upgrade insecure CSS resources, indicated
+# by url() functions in CSS, using the https_url function.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
+#
+# Copyright (c) 2017 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Hooks::ProxyCSSLinks;
+
+use strict;
+use LJ::Hooks;
+use LJ::CleanHTML;
+
+LJ::Hooks::register_hook( 'css_cleaner_transform', sub {
+    my $textref = $_[0];
+    return unless ref $textref && $$textref;  # nothing to do
+
+    my $ssl_url = sub {
+        my ( $url, $q ) = @_;
+        $q ||= '"';  # provide double quotes if none specified
+        $url = LJ::CleanHTML::https_url( $url );
+        return "url($q$url$q)";
+    };
+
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Hooks/ProxyCSSLinks.pm
```diff
@@ -0,0 +1,35 @@
...
+    return unless ref $textref && $$textref;  # nothing to do
+
+    my $ssl_url = sub {
+        my ( $url, $q ) = @_;
+        $q ||= '"';  # provide double quotes if none specified
+        $url = LJ::CleanHTML::https_url( $url );
+        return "url($q$url$q)";
+    };
+
+    $$textref =~ s/\burl\(\s*(['"]?)(.*?)\1\s*\)/$ssl_url->($2,$1)/geis;
```
zorkian: Is this regex sourced from somewhere (and I can just rubberstamp it) or should I do an actual analysis? Heh.

kareila: I regret to inform you I concocted it myself. 😜

zorkian: Okay; I don't have the brain for it tonight and will come back to it next pass. Sorry!

kareila: That's fine, I think this one needs the most scrutiny of the bunch anyway.  Thanks for powering through all the others. ❤️

alierak: I looked at it some last night but really need a running test instance to attack. So far I figured out that the existing CSS cleaner allows URLs containing backslashed quotes, which this regex would misparse.

zorkian: Just coming back to this, are you saying the regex needs an update, or that the regex is good b/c it prevents some existing bad behavior?

alierak: Hah! I don't remember. I think I was just going to see whether I could exploit the difference in behavior between this and the CSS cleaner to inject some text outside of what it parses as the URL. I don't think I got around to it because dreamhacks were down at that point. No idea if it's actually exploitable, but the difference in quote-matching and honoring backslash escapes is the sort of thing that piques my interest.

zorkian:
Even if this is wrong, I think it only poses a risk to break things not to make them insecure? It's only taking input urls and outputting things passed to our https upgrader, so assuming that's secure (it's out of scope of this review anyway) then I think it's okay.

PR #2092, opened 2017-04-12
[#2089] calculate correct URL for entry import metadata

### LINKED ISSUE #2089: URL link to remote entry not correct in some importer notifications
http://www.dreamwidth.org/support/see_request?id=36241

> A few weeks ago I received a bunch of notifications re: imports, and one
> was a a failure to import a post to a community for reason of "not being
> able to create tags" in said community.
> 
> Along with the notification was a link pointing to said entry. This link
> read (abstracted for form) 'site.xyz/communityname/entryDigits' but the
> actual link was resolved to
> https://www.dreamwidth.org/inbox/site.xyz/communityname/entryDigits.
> 
> The href attribute for that link was 'site.xyz/communityname/entryDigits'.
> it needs to be prepended by 'http(s?)://(www.)?' 

I think I've discovered the responsible statement, in DW/Worker/ContentImporter/LiveJournal/Entries.pm:

`$evt->{key} = $evt->{url} = $data->{hostname} . '/' . ( $data->{usejournal} // $data->{username} ) . '/' . ( $evt->{itemid} * 256 + $evt->{anum} );`

I suspect that `$evt->{key}` should remain unchanged, but `$evt->{url}` should prepend a protocol.

kareila:
Using a new method, `DW::External::Site->entry_url`, which uses itemid and anum to calculate the page link using the journal_url method, which can differ between sites.

There is an assumption here that every import source will have a definition in DW::External::Site, but that seems safe enough.

Fixes #2089.

### DIFF
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -56,9 +56,9 @@ $domaintosite{"pinboard.in"} = DW::External::Site->new("20", "www.pinboard.in", 
 $domaintosite{"fanfiction.net"} = DW::External::Site->new("21", "www.fanfiction.net", "fanfiction.net", "FanFiction", "FanFiction");
 $domaintosite{"pinterest.com"} = DW::External::Site->new("22", "www.pinterest.com", "pinterest.com", "Pinterest", "pinterest");
 $domaintosite{"youtube.com"} = DW::External::Site->new("23", "www.youtube.com", "youtube.com", "YouTube", "yt");
-$domaintosite{"github.com"} = DW::External::Site->new("24", "www.github.com", "github.com", "GitHub", "gh"); 
+$domaintosite{"github.com"} = DW::External::Site->new("24", "www.github.com", "github.com", "GitHub", "gh");
 # three-part domain name
-$domaintosite{"lj.rossia.org"} = DW::External::Site->new("25", "lj.rossia.org", "lj.rossia.org", "LJRossia", "lj"); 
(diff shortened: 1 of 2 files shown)


cgi-bin/DW/External/Site.pm
```diff
@@ -206,6 +206,24 @@ sub journal_url {
...
+        unless $u && ref $u eq 'DW::External::User';
+
+    # override this on a site-by-site basis if needed
+    my $base = $self->journal_url( $u );
+
+    # several possible options for specifying an entry;
+    # for now we just support itemid + anum
+    return unless exists $opts{itemid} && exists $opts{anum};
+
+    my $pagenum = $opts{itemid} * 256 + $opts{anum};
```
alierak: This particular calculation is specific to "lj" type sites, but then, so is the importer thus far. Note in case we ever care, or maybe have the function complain if wrong site type?

alierak: I guess it's unlikely the entry would have itemid and anum if it's another site type though.

kareila: Yeah. If we're adding another type of site to the importer, it will use a different module, not LiveJournal/Entries.pm.  If we ever use this in another context, we can improve it based on what information is available in that context.

zorkian:
Touching this code means you own it now. I really can't tell you how happy this makes me. 🔥 🥔 🔥

rahaeli:
Cracked up at that & was laughing so hard it startled the cat!

PR #2103, opened 2017-04-13
[#2097] raise feed size limit for initial fetch

### LINKED ISSUE #2097: make feed-fetch user agent use the same size limits as synsuck
https://www.dreamwidth.org/support/see_request?id=36264

Trying to create a feed with that feed URL (http://feeds.feedburner.com/ProbablyOverthinkingIt) and putting in the name think_bayes (so it becomes think_bayes_feed) leads to a page on DW that just says "Error: 200 OK". 

The feed does validate as valid Atom, but has a few warnings. I don't think any of them could cause the problem, though?

kareila:
Uses the same limit as the synsuck process.

Also prints a more informative error message if the feed was still too large when fetched.

Fixes #2097.

### DIFF
--- a/cgi-bin/LJ/SynSuck.pm
+++ b/cgi-bin/LJ/SynSuck.pm
@@ -61,6 +61,17 @@ sub delay {
     return undef;
 }
 
+sub max_size {
+    my ( $u ) = @_;  # optional user object for feed
+    my $max_size = $LJ::SYNSUCK_MAX_SIZE || 3000; # in kb
+
+    if ( $u && $u->has_priv( "siteadmin", "largefeedsize" ) ) {
+        $max_size = $LJ::SYNSUCK_LARGE_MAX_SIZE || 6000; # in kb
+    }
+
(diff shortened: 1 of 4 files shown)


cgi-bin/LJ/SynSuck.pm
```diff
@@ -61,6 +61,17 @@ sub delay {
     return undef;
 }
 
+sub max_size {
+    my ( $u ) = @_;  # optional user object for feed
+    my $max_size = $LJ::SYNSUCK_MAX_SIZE || 3000; # in kb
+
+    if ( $u && $u->has_priv( "siteadmin", "largefeedsize" ) ) {
+        $max_size = $LJ::SYNSUCK_LARGE_MAX_SIZE || 6000; # in kb
```
zorkian: If you feel like simplifying I'm pretty sure we can rip this out and just set the default to 6MB. This was written in a time when web servers have 4GB of RAM and such.

PR #2104, opened 2017-04-13
[#2095] temporarily disable LJRossia crossposts

### LINKED ISSUE #2095: remove lj.rossia.org as xpost destination (while keeping it in the user tag)
https://www.dreamwidth.org/support/see_request?id=36290

From the error message, I'm going to guess that the problem is either a) lj.rossia.org doesn't have a working version of the protocol running, or b) they do but the xposter is having trouble handling it because it's a three-level domain and not a two-level domain. (I know that was a problem with adding LJR to the user tag.)

Obviously we can't fix A, but if it's B, that's something we can do something about!

kareila:
This removes LJRossia from the "Add External Account" form and preemptively fails any attempts to crosspost there.

It turns out that failing the job doesn't automatically notify the user, so we also make sure to send a notification to inform the user of the current situation.

Fixes #2095.

### DIFF
--- a/cgi-bin/DW/Worker/XPostWorker.pm
+++ b/cgi-bin/DW/Worker/XPostWorker.pm
@@ -62,6 +62,10 @@ sub work {
     my $acct = DW::External::Account->get_external_account( $u, $acctid )
         or return $job->failed("Unable to load account $acctid for uid $uid");
 
+    # LJRossia is temporarily broken, so skip
+    return $job->permanent_failure( "Crossposts to LJRossia are disabled until the remote site fixes their XMLRPC protocol handler." )
+        if $acct->externalsite && $acct->externalsite->{sitename} eq 'LJRossia';
+
     my $domain = $acct->externalsite ? $acct->externalsite->{domain} : 'unknown';
 
     my $entry = LJ::Entry->new( $u, ditemid => $ditemid );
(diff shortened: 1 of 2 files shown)


cgi-bin/DW/Worker/XPostWorker.pm
```diff
@@ -62,6 +62,22 @@ sub work {
...
+    my $notify_fail = sub {
+        $sclient->insert_jobs(
+            LJ::Event::XPostFailure->new(
+                $u, $acctid, $ditemid, ( $_[0] || 'Unknown error message.' )
+            )->fire_job
+        );
+    };
+
+    # LJRossia is temporarily broken, so skip - but we do want to notify
+    if ( $acct->externalsite && $acct->externalsite->{sitename} eq 'LJRossia' ) {
```
zorkian: Wow, is this really the best way of doing this?! Hardcoding like this feels pretty sad, but I guess the original design of the xpost system didn't take into account that we might want to administratively disable things? I don't recall thinking of this use case :)

kareila:
If we remove it from the external sites list entirely, then all the other functionality provided by DW::External::Site, like userheads, will stop working as well.

zorkian:
We have no way to say 'this is an external site that doesn't allow crossposting'? I could have sworn that was a flag we had to turn on or off?

kareila:
It's dependent on "servicetype" - if we update the list to say it's not an LJ-based site, that will do it, but that's (a) less obviously temporary and (b) might cause other problems.

zorkian:
Makes sense. Thanks for thinking about it!

PR #2106, opened 2017-04-13
[#2096] don't include FAQs with no category in LJ::Faq->load_matching

### LINKED ISSUE #2096: don't return hidden FAQs when someone uses the FAQ search
People using the FAQ search on https://www.dreamwidth.org/support/ shouldn't see results from any FAQ that's hidden (has no faqcat), since instead of deleting FAQs, we drop them in "no category" in case we ever need to resurrect them.

kareila:
This page has the only use of this method, so changing its behavior here won't affect anything else.

Fixes #2096.

### DIFF
--- a/cgi-bin/LJ/Faq.pm
+++ b/cgi-bin/LJ/Faq.pm
@@ -123,7 +123,7 @@ sub load_all {
     my $lang = delete $opts{lang} || $LJ::DEFAULT_LANG;
     my $faqcat = delete $opts{cat};
     my $allow_no_cat = delete $opts{allow_no_cat} || 0;
-    
+
     my $wherecat = "";
     if ( $allow_no_cat ) {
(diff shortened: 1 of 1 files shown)


cgi-bin/LJ/Faq.pm
```diff
@@ -493,34 +493,22 @@ sub load_matching {
...
-	}
-
-	next unless $score;
-
-	$scores{$f->{faqid}} = $score;
-
-	push @results, $f;
+        my $score = 0;
+
+        $score += 3 if $f->question_raw =~ /\Q$term\E/i;
```
zorkian: WHAT IS THIS WE DO FAQ SEARCH WITH MANUAL SCORING?!?!?

omg if I ever knew this I had purged it entirely from my brain

rahaeli: I love when we find things that horrify you, Mark.

IIRC, worst username evar hacked this together for support in a weekend because I was getting increasingly desperate!

PR #2110, opened 2017-04-13
add medium to user tag and profile other-sites block

srukle:
This addresses issue #1997.

Sorry, the past PR was documented incorrectly. Thank you for pointing that out @afuna.


cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -838,6 +838,17 @@ sub external_services {
         };
     }
 
+    if ( my $emedium = $u->prop( 'medium' ) ) {
```
zorkian: This should be `my $medium = ...` reason being you need to have the raw contents so you can escape it for HTML later and for URL both.

cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -838,6 +838,17 @@ sub external_services {
...
+	my $emedium = LJ::eurl( $emedium );
+	push @ret, {
+	    type => 'medium',
+	    text => LJ::ehtml( $emedium ),
```
zorkian: You don't want to double-escape something -- in this case you eurl'd the contents and then you're ehtml'ing the eurl'd comments which won't work out well.

kareila:
I just noticed that this version of the pull request is missing the necessary changes to htdocs/manage/profile/index.bml.

srukle:
Does @kaberett's fork resolve these problems?
https://github.com/kareila/dw-free/commit/3aa01daad73520f755073e8e80d901c3725718b3

zorkian:
Pinging @kareila on the question!

kareila:
Sorry - I'm not sure which fork you mean. The link you pasted was to my work for the user tag.

If you look at #2008 you'll see there are changes there to htdocs/manage/profile/index.bml that were not included here.

Also, this PR now has a merge conflict due to overlapping changes in htdocs/profile.bml.text.

Please address these issues and also @zorkian's requested changes.

zorkian:
This is old -- closing for now! If you ever have time/want to pick it up again, would love to have you!

PR #2115, opened 2017-04-14
move /openid/server to DW::Controller

kareila:
I removed the code block concerning fake `_ljwho` identities and OPENID_DEST_DOMAIN_TRUSTED hashes, because it seemed outdated and unused.

This is a new multiple-use branch to gradually convert pages under htdocs/openid.  I did this one first because it was simplest, but without a functioning dreamhack, I don't have a way to test it right now.


cgi-bin/DW/Controller/OpenID.pm
```diff
@@ -22,10 +23,41 @@ use DW::Routing;
...
+    my $r = DW::Request->get;
+    my $get = $r->get_args;
+
+    my $trust_root = $get->{'openid.trust_root'} // '';
+    my $return_to  = $get->{'openid.return_to'} // '';
+
+    ## Non-OpenID-compliant section: rewrite LiveJournal's trust_root to
+    ## https so that it will match their return_to URL and pass validation.
+    $get->{'openid.trust_root'} = 'https://www.livejournal.com/'
+        if ( $trust_root eq 'http://www.livejournal.com/' &&
```
zorkian: I believe the style guide requests not to use parenthesis when they aren't necessary?

zorkian:
Yeah I forget what that was for but it might have been Fotobilder?

cgi-bin/DW/Controller/OpenID.pm
kareila: I plead copy/paste artifact.

kareila:
Whatever it was, it's gone now.

Issue #2121, opened 2017-04-17
BML::get_query_string not preserved when doing an SSL redirect?


kareila:
This wins the longevity prize for surviving as a bug since the LJ fork:

LJ::make_auth_code returns a string of random letters and digits from a preselected list, but support/act.bml parses the query string with a regular expression that assumes the auth code doesn't contain any digits.

Changing `\w{15}` to `[\w\d]{15}` should fix the problem.

Reported in an antispam request: http://www.dreamwidth.org/support/see_request?id=36360

rahaeli:
Whoa. That might actually win the prize for "oldest bug we've ever fixed", in fact: it was present in the original checkin from 16 years ago. (https://github.com/dreamwidth/livejournal/blame/master/htdocs/support/act.bml)

kareila:
It's old enough that I wonder if the behavior of `\w` changed at some point - surely this worked at some point in the past?

zorkian:
I'm pretty sure `\w` matches numbers?

```
    \w        [3]  Match a "word" character (alphanumeric plus "_", plus
                                      other connector punctuation chars plus Unicode
                                      marks)
```

And:

```
$ perl -e 'print 1 if "123" =~ /^\w+$/;'
1
```

kareila:
Hmm, that wasn't my recollection, but if that's the case, can you figure out why the query string in the support request didn't match the regex?

rahaeli:
Having looked more closely at the error the user was getting -- yeah, that code has worked for a while. Hrrrrmmmm.

zorkian:
Issue is the SSL redirect. The original link in the support request is `http://` and when we do the redirect we're not preserving the GET arguments in this case somehow.

kareila:
*spits curses at BML*

rahaeli:
OHO. Good debugging work!

kareila:
DW::Request->get->query_string is a drop-in replacement that doesn't rely on BML.  See if that has the same problem?

kareila:
@zorkian says that the HTTPS Everywhere redirect is using code that only works for query strings that parse neatly into argument hashes.  Unorthodox query strings are getting dropped in the conversion process.  We'll need to update the LJ::create_url function to allow these to pass through.

PR #2122, opened 2017-04-17
Fix redirect for query strings

zorkian:
Certain codepaths use querystrings that do not contain arguments and
instead are just plaintext. This codepath was only passing through valid
looking arguments. This updates it to always use the raw query string.


kareila:
Fixes #2121.

PR #2132, opened 2017-04-20
Image pages

### LINKED ISSUE #1985: add file hosting quota utilization to manage page
https://www.dreamwidth.org/file/list and https://www.dreamwidth.org/file/edit should probably have "you have used XXX of your 500MB quota".

### LINKED ISSUE #2033: using an apostrophe in an image's short description produces invalid markup
If you put something containing an apostrophe (such as a possessive) in the short description/alt text of an image, the resulting code to copy and paste into your entry provokes an invalid-markup error.

Example:

> <a href='http://synecdochic.dreamwidth.org/file/104311.jpg'> src='http://synecdochic.dreamwidth.org/file/100x100/104311.jpg' 
> title='Thea sprawling on Sarah'  alt='Thea sprawls in Sarah's lap, trusting 
> Sarah to keep her from falling'  /></a>

The HTML cleaner thinks the second ' (in this case, the one in "Sarah's") closes out the alt attribute, and tosses out an invalid-markup error.

I'm guessing the best solution will be to have http://www.dreamwidth.org/file/new escape apostrophes/single-quotes when saving alt text?

### LINKED ISSUE #2036: security on image cannot be set to access list on image upload page
If you try to change the security level from public to access list, it will be set to private security instead.  Changing the security to private or leaving it on public acts as expected.

Repro Steps: 

1. Upload an image 
2. Set the visibility to access list from the upload page 
3. Image is still set to private security despite showing as being changed to access list 

Workaround: Can be fixed by going to the Manage Images page and re-saving all the settings (don't need to change the dropdown), but it'll remain private until either a new image is uploaded or you hit save on the Manage Images page instead of the upload page.

kareila:
Various fixes for the image upload/management pages.

Fixes #1985.  Fixes #2033.  Fixes #2036.

### DIFF
--- a/htdocs/js/jquery.fileupload.js
+++ b/htdocs/js/jquery.fileupload.js
@@ -27,6 +27,10 @@ $(function() {
                 data[file_id] = {};
 
             data[file_id][form_field.name] = form_field.value;
+
+            if (form_field.name == "security" && form_field.value == "usemask")
+                data[file_id]["allowmask"] = 1;
+
         });
 
         $.ajax( Site.siteroot + '/api/v1/file/edit', {
--- a/htdocs/js/jquery.fileupload.js
+++ b/htdocs/js/jquery.fileupload.js
@@ -213,10 +213,13 @@ $(function() {
         $.extend( image, data );
         $field.data( "image-attributes", image );
 
+        var escape_titletext = image.title.replace( /</g, '&lt;' ).replace( /'/g, "&apos;" );
+        var escape_alttext = image.alttext.replace( /</g, '&lt;' ).replace( /'/g, "&apos;" );
+
         var text = [];
         text.push( "<a href='" + image.url + "'><img src='" + image.thumbnail_url + "'" );
-        if ( image.title ) text.push( " title='" + image.title + "' " );
-        if ( image.alttext ) text.push(" alt='" + image.alttext + "' ");
+        if ( escape_titletext ) text.push( " title='" + escape_titletext + "' " );
+        if ( escape_alttext ) text.push(" alt='" + escape_alttext + "' ");
         text.push( " /></a>" );
         $field.val(text.join(""));
     });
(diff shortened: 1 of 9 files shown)


htdocs/js/jquery.fileupload.js
```diff
@@ -209,10 +213,13 @@ $(function() {
         $.extend( image, data );
         $field.data( "image-attributes", image );
 
+        var escape_titletext = image.title.replace( /</g, '&lt;' ).replace( /'/g, "&apos;" );
```
zorkian: It boggles my mind that there isn't a function for this, but I did peek around and didn't see one... just a bunch of people on the Internet saying 'do it yourself'. And I guess we don't have such a function already anywhere?

kareila: I really don't know jQuery well, but this does work.  I'm open to a more elegant solution if anyone has one.

alierak: I've got an idea, but I don't know if it's elegant. Instead of manipulating text, for which the functions don't already exist, you'd have to manipulate the DOM so that the page actually contains the img you're trying to build up, non-displayed. Then ask jQuery for the html source of that img.

kareila:
@alierak and I couldn't figure out a better way to do this in the given context.  I have tweaked my initial approach to fix a couple of issues, but we agree this is probably the best way to do it in this instance.

alierak:
I did eventually find an "existing function", though. In js/6alib/core.js, they define String.encodeHTML for this purpose. However, it may contain a bug in that it does not replace single quotes.

alierak:
Scratch the comment about that function having a purpose, though, since I don't believe it's called anywhere.

kareila:
Thanks for confirming that this is a hard problem to solve. I'll file it next to cache invalidation and naming things.

PR #2133, opened 2017-04-20
[#2131] change /editicons to /manage/icons

### LINKED ISSUE #2131: alias /editicons to /manage/icons
https://www.dreamwidth.org/editicons should also be reachable by https://www.dreamwidth.org/manage/icons, because I forget it isn't *every goddamn time*.

(Make /manage/icons canonical, support redirects from /editicons for at least a few years.)

kareila:
Fixes #2131.


zorkian:
Wait you mean the page doesn't let you edit your icons?!

rahaeli:
More like, "AUUUUGH WHY ARE WE SO INCONSISTENT WITH WHAT'S IN /MANAGE AND WHAT'S NOT"...

PR #2135, opened 2017-04-20
[#2077] allow ?cat as well as ?category for /support/submit

### LINKED ISSUE #2077: allow /support/submit to have URL argument for supportcat
When replying to a user, I often find myself typing out something like "&lt;a href="https://www.dreamwidth.org/support/submit">open a support request&lt;/a> in the Account Payments category" (or "the Anti-Spam category", or "the Terms of Service category", etc) to direct them to the right supportcat. Wouldn't it be nice if I could type "&lt;a href="https://www.dreamwidth.org/support/submit?cat=accounts">open a support request&lt;/a> in the Account Payments category" and specify the category's shortcode (the one used for URL filtering on support/help.bml) to have the form render with the specified category pre-selected in the drop-down box? Yes, yes it would be.

kareila:
Fixes #2077.

### DIFF
--- a/cgi-bin/DW/Controller/Support/Submit.pm
+++ b/cgi-bin/DW/Controller/Support/Submit.pm
@@ -117,8 +117,9 @@ sub submit_handler {
     $vars->{include_email} = ($remote && $remote->email_raw) ? 0 : 1;
 
     my $cats = LJ::Support::load_cats();
+    my $catarg = $r->get_args->{cat} || $r->get_args->{category};
     my $cat;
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Controller/Support/Submit.pm
```diff
@@ -117,8 +117,9 @@ sub submit_handler {
     $vars->{include_email} = ($remote && $remote->email_raw) ? 0 : 1;
 
     my $cats = LJ::Support::load_cats();
+    my $catarg = $r->get_args->{cat} || $r->get_args->{category};
```
zorkian: We'll have to never make a cat called `0`.

PR #2142, opened 2017-04-26
increment dw.opml.used stat when generating output

kareila:
@zorkian: "I've forgotten all about whatever this is: https://www.dreamwidth.org/tools/opml"
@rahaeli: "I think it's a way to export the accounts you read to a feed reader?"
@zorkian: "okay. do we need it? is it even linked anywhere?"
@rahaeli: "looks like some people use it in third-party tools? idk"
@rahaeli: "we should stick a tally counter in the module that generates it and see how many times the code is called over the next month or whatever"

### DIFF
--- a/htdocs/tools/opml.bml
+++ b/htdocs/tools/opml.bml
@@ -34,7 +34,7 @@ _c?><?_code
     # will use watched accounts for personal accounts
     # and identity accounts
     # and members for communities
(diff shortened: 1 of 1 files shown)


zorkian:
@rahaeli and I stand around discussing and @kareila picks up the keyboard and gets things done. <3

rahaeli:
I'm still not 100% convinced she isn't telepathic.

PR #2143, opened 2017-04-26
convert URL space to %20 in get_proxy_url subroutine

### LINKED ISSUE #2134: HTML cleaner should replace space in URLs with %20 before calculating checksum
http://dw-maintenance.dreamwidth.org/75916.html?thread=2769548#cmt276948

This according to @alierak: "The correct checksum according to the proxy log would be 54f3f9186cd9, so I have a fairly strong hunch that the checksum was initially calculated based on there being a space in the URL, but when they come back and try to load the image, the browser has replaced a space in the URL with %20, which yields a different checksum."

kareila:
This will change the link returned by the proxy to use %20 if the
original URL included a space, which is what most browsers will
expect.  Without this transformation, requests that changed
spaces to %20 would fail because the checksums wouldn't match.

Fixes #2134.

### DIFF
--- a/cgi-bin/DW/Proxy.pm
+++ b/cgi-bin/DW/Proxy.pm
@@ -41,6 +41,9 @@ sub get_proxy_url {
     my ( $url, %opts ) = @_;
     return undef unless $LJ::PROXY_URL && substr($url, 0, 7) eq 'http://';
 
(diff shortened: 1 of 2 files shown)


alierak:
Yay, thanks!

PR #2144, opened 2017-04-26
[#1854] improve detection of metadata headers in emailposts

### LINKED ISSUE #1854: Email posting: post-security tag not working (possibly under some conditions)
From support with some useful info and testing:
https://www.dreamwidth.org/support/see_request?id=34146

It seems placing the post-security tag after body text or using it with no body text means it's ignored and security reverts to your default mobile posting one (or entry security?). The user mentions having issues only when attaching images but I was having the same issues with text only.

I'm also having issues with the post-security tag no matter where I place it. It is parsed as body text no matter what I do (text-only; latest Thunderbird).

kareila:
You still can't post an entry with a blank body because the
protocol won't allow it, but you can put metadata headers
in places other than the very beginning of the body text
and have them be properly detected.

Fixes #1854.

### DIFF
--- a/cgi-bin/DW/EmailPost/Base.pm
+++ b/cgi-bin/DW/EmailPost/Base.pm
@@ -227,13 +227,13 @@ sub _extract_post_headers {
     my ( %post_headers, $amask );
 
     # first look for old style lj headers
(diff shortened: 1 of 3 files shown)


zorkian:
Making me reason through regexes. X_X; I'm so rusty at these.

PR #2151, opened 2017-04-28
[#1891] do server-side permission check for comment search

### LINKED ISSUE #1891: comment search needs permission check on form submit as well as form rendering
http://www.dreamwidth.org/support/see_request?id=34652

Should render as a regular entry search regardless of checkbox state if the account can't search comments.

kareila:
If the check fails, just do regular entry search.

Fixes #1891.


zorkian:
Wherps.

PR #2152, opened 2017-04-28
[#1998] make print_global_title a hyperlink to Recent Entries

### LINKED ISSUE #1998: make journal title a clickable link back to the journal
Pulled out of the suggestions queue because it's so obvious I thought we already did this: the journal title (the h1 tag) should be a clickable link back to the recent entries page (username.dreamwidth.org) on all journal-styled journal views (recent, reading, tags, archive, icons, individual entry pages).

kareila:
This should modify the behavior for every page except for reading pages that use a custom title.

The Boxes and Borders layout had a customized print_global_title function that required updating as well.

Fixes #1998.

### DIFF
--- a/styles/boxesandborders/layout.s2
+++ b/styles/boxesandborders/layout.s2
@@ -145,7 +145,7 @@ function Page::print_global_title() {
     }
 
     if ($.global_title) {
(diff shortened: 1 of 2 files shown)


zorkian:
Yay! Titles should definitely be links.

PR #2154, opened 2017-04-28
Cleanup for 4/30 code push

kareila:
The usual miscellany.


cgi-bin/LJ/HTMLControls.pm
```diff
@@ -224,8 +224,10 @@ sub _html_option {
 
     # are there additional data-attributes?
     my $data_attribute = '';
-    if ( $item->{data} ) {
-        $data_attribute .= " data-$_='$item->{data}->{$_}'" foreach keys %{$item->{data} || {}};
+    my %item_data = $item->{data} ? %{$item->{data}} : ();
```
zorkian: I don't know if it matters here, but traditionally if we're about to deref something we do a ref check:

```my %item_data = ref $item->{data} eq 'HASH' ? %{$item->{data}} : ();```

File: cgi-bin/LJ/Userpic.pm [hunk 2]
```diff
@@ -883,10 +883,11 @@ sub make_default {
     $u->{'defaultpicid'} = $self->id;
 }
 
-# returns true if this picture if the default userpic
+# returns true if this picture is the default userpic
 sub is_default {
-    my $self = shift;
+    my $self = $_[0];
     my $u = $self->owner;
+    return unless defined $u->{'defaultpicid'};
```
zorkian: Extraneous single quotes sighted!

PR #2170, opened 2017-05-09
search all language directories for deadphrase files

kareila:
The texttool script was only searching LJHOME for deadphrase data, and there's a deadphrases-local.dat in ext/dw-nonfree.


zorkian:
Oh wow, heh.

PR #2173, opened 2017-05-10
Adds missing space to the .message.explicit.communityreasons string.

### LINKED ISSUE #2172: missing space in text string about discretion-advised communities
`views/journal/adult_content.tt.text:.message.explicit.communityreason=[[journal]] provided the following reason for this community being marked"suitable for 18+": <strong>[[reason]]</strong>.`

Needs an additional space! And is an excellent way for me to break in my Dreamhack.

kaberett:
Fixes #2172.

I *think* I've remembered how to edit strings, but It's Been A While so please tell me if I've screwed up/should be putting stuff in deadphrases/etc.

### DIFF
--- a/views/journal/adult_content.tt.text
+++ b/views/journal/adult_content.tt.text
@@ -39,7 +39,7 @@
 
 .message.explicit.bycommunity=You're about to view content that community administrators on [[journal]] have marked as <strong>inappropriate for anyone under the age of 18</strong>. To continue, you must confirm that you're at least 18 years of age.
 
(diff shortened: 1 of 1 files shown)


alierak:
Per conversation, this is fine without renaming the string or anything because that will keep us from having to complicate the code in cgi-bin/DW/Logic/AdultContent.pm that decides which string to use. This string will also need to be updated in production (and translations, if anybody has them).

rahaeli:
Updated in prod!

PR #2187, opened 2017-05-20
Use the right CSS property in FAQ css (fixes #1886)

### LINKED ISSUE #1886: weird FAQ rendering issue
I tried to add a FAQ just now that contained an unordered list using `<ul>` tags.  When I previewed the FAQ, it rendered normally, but when I looked at it via faqbrowse after saving, the list displayed as an ordered list with numbers.

pinterface:
'outside' is a list-style-position, not a list-style-type.  This seems
to result in certain versions of certain browsers displaying unordered
lists as though they were ordered lists, which is less than ideal.

Hopefully fixes #1886

### DIFF
--- a/htdocs/stc/faq.css
+++ b/htdocs/stc/faq.css
@@ -1,5 +1,5 @@
 #faq ul, #faq ol {
     margin-left: 3em;
     margin-bottom: 2em;
-    list-style-type: outside;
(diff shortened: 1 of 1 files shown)


zorkian:
That does sound.....quite less than ideal. Thank you!

PR #2192, opened 2017-05-30
2718 next prev entry

### LINKED ISSUE #2178: Add next/previous entry shortcuts to reading/entry pages
( from https://dw-suggestions.dreamwidth.org/1504972.html )

As modified from comments in the above suggestion, add a shortcut that will allow for quick access to move to next and previous entries.  

Implementation will be as configurable shortcuts either as key bindings or as touch gesture bindings.

akpdw:
Fixes #2178

This adds a new 'Shortcuts' section to the settings page, with settings for both keyboard and touch shortcuts.  Available shortcuts are for next and previous entry.
 
Shortcuts, if configured, are made available on Entry, Reading, and Recent pages.

This also adds two new javascript libraries: mousetrap.min.js (https://github.com/ccampbell/mousetrap) and jquery.touchSwipe.js (https://github.com/mattbryson/TouchSwipe-Jquery-Plugin), with the latter having been customized some to work with multi-finger gestures on iOS.

There is also a new shortcuts.js file to register configured shortcuts, and jquery.shortcuts.nextentry.js which actually does the next/previous entry actions.

...I was able to test the iOS gestures a lot more than the Android touch gestures, so the latter might need a little bit more attention.


cgi-bin/LJ/Talk.pm
```diff
@@ -2256,6 +2256,40 @@ sub init_s2journal_js {
         ) ) if $opts{lastn};
 }
 
+# convenience function for keyboard shortcuts
+# args: $remote and $p for adding javascript to header
+# returns: nothing, just calls need_res
+sub init_s2journal_shortcut_js {
+    my ( $remote, $p ) = @_;
+
+    if ( $remote ) {
```
zorkian: nit: To prevent overly deep nesting, prefer to just return early without `$remote`.

cgi-bin/LJ/Talk.pm
```diff
@@ -2256,6 +2256,40 @@ sub init_s2journal_js {
...
+        my $connect_string = "";
+        if ( $remote->prop( "opt_shortcuts" ) ||  $remote->prop( "opt_shortcuts") ) {
```
zorkian: Did you mean to look at a different property here? This seems like the same one twice.

nit: Also, same comment about deep nesting: you could just return if these are false.

akpdw:
@zorkian Changes made.

zorkian:
Squashed and merged! Thank you so much for doing this!

PR #2194, opened 2017-06-10
Changes sort order to newest-first on 'upload images' page.

### LINKED ISSUE #2156: When uploading images, display most recent upload at top
At the moment, images are added to the bottom of the list of uploads for the session - it's more useful if they're added at the top, so that users can immediately see if the upload succeeded and add text, without having to scroll back and forth.

From http://dw-suggestions.dreamwidth.org/1505666.html

rshatch:
Fixes #2156


zorkian:
Makes sense!

PR #2203, opened 2017-07-16
Update Delicious profile and user links

### LINKED ISSUE #2184: Update user profile Delicious link
https://www.dreamwidth.org/support/see_request?id=36745:
> The links on a user's profile will go to www.delicious.com/username which worked in the past I think but doesn't now. The new link needs to be del.icio.us/username to work. 

... and in fact integration with Delicious is in general broken by the move to del.icio.us.

kaberett:
Fixes #2184.

I'm not entirely sure *why*, which is a little concerning -- in particular I'm sufficiently hazy on what Site.pm is doing that I'm preeeeeeetty sure I don't want both "delicious.com" and "del.icio.us" set up -- buuuuut it appears, in testing, to work, so I would very gratefully have an explanation, please! :-)


zorkian:
If it works in your testing I'm happy with it. The code in this diff looks fine, and I don't recall the functionality of SIte.pm well enough to know the answer to your question... T_T

PR #2220, opened 2017-09-09
Add S3 support to BlobStore

zorkian:
This implements S3 support for BlobStore using the Paws SDK, which seems
to work fairly well. This also updates the BlobStore initialization code
to not be magical about what fallbacks are valid, it's now just an array
that can be configured directly by the admin.

AWS credentials are by default handled using IAM roles/instance
profiles, which is the secure way to do it if you are running a server
in EC2. This is highly recommended, but fallback support is provided for
specifing the keys directly (for now?).

### DIFF
--- a/bin/checkconfig.pl
+++ b/bin/checkconfig.pl
@@ -296,6 +296,9 @@ my %modules = (
     "Text::Wrap" => {
         ver => '2013.0523', # issue #1447
     },
+    "Moose" => {
+        opt => "Used for BlobStore support of S3",
+    },
 );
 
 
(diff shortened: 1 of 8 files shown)


kareila:
Any suggestions for how to test this on a Dreamhack, for those of us who know nothing about how S3 works?

zorkian:
Okay, it's all configured for the hack server. You should just need to add this to your BlobStore configuration:

```
@BLOBSTORES = (
    s3 => {
        access_key => undef,
        secret_key => undef,
        bucket => 'dw-hack-blobstore',
        region => 'us-west-1',
        prefix => 'kareila',
    },
);
```

PR #2224, opened 2017-09-28
Don't proxy RSS images

zorkian:
When we're cleaning content for display on an external site (such as in
feeds), we don't want to convert links into Dreamwidth proxy URLs. It
won't work very well and it's unnecessary.


zorkian:
From https://www.dreamwidth.org/support/see_request?id=37504

afuna:
Nice.

PR #2226, opened 2017-10-05
advertise OpenID 1.1, remove 1.0 compat option, fixes #2225

### LINKED ISSUE #2225: Advertise http://openid.net/signon/1.1 via Yadis
When acting as a OpenID provider, discovery via the Yadis protocol advertises OpenID 1.1:

https://github.com/dreamwidth/dw-free/blob/f54a0ad3e0ae7cc03219d73c03ffff47c1f61fbc/cgi-bin/LJ/Feed.pm#L798

However, the implementation seems to be something more like OpenID 1.1. There are very few differences between the two protocols, but one difference is that with OpenID 1.1 there is an
is_valid element in the mode.check_authentication response, something that the dreamwidth.org response seems to have.

I just worked around this in the outreachy.org application system:
https://git.gnome.org/browse/opw-web/commit/, which has a fork of code from LightOpenID - see:
https://github.com/iignatov/LightOpenID/blob/master/openid.php

alierak:


alierak:
Fixes #2225

PR #2230, opened 2017-10-26
TT/Foundation the logout page

zorkian:
This slightly changes the behavior in that we don't just send people to
the homepage when they log out. Now they get a nice message confirming
the actions they took.


pauamma:
Shouldn't logout.tt be English-stripped?

zorkian:
Great question.

I was trying to find the source of it, but there was some conversation somewhere about this. We're never going to translate Dreamwidth, and one of our biggest issues is all the hurdles people have to jump through to contribute code. I've been slowly killing systems that are hard to run/use/understand (Perlbal, MogileFS, etc) and so I'm going to extend that to things like English stripping.

Denise's request was that before we wholesale abandon doing so, we should have a way of quickly changing text on the site without a code push. That's totally reasonable, so for today we're still generally asking people to strip things until we have something better.

Sophira:
Abandoning English-stripping would imply that "update text without code push" would need to alter the original source files in some way.

The cleanest way to do *that* without risking overwrites would be having a perpetually unmerged local branch in git that only ever gets rebased onto released code tags. This then ensures that it would always be safe to auto-pull onto the site itself, I believe? It could even be pushed to GitHub for transparency, but you'd need to make certain nobody tries to open a pull request to it.

Unfortunately this method is prone to git conflicts so it would require manual intervention in some cases, which is kind of the opposite of what you want.

I can't think of a better way than that, though - this is what VCS systems such as git are designed for, after all. Any other method would require storing information outside of the source files and would have to rewrite the source files on each update to prevent changes being overwritten, and this doesn't feel like something that could handle conflicts any better than git can.

So I kinda feel that unless you're okay with handling conflicts manually as they occur, English-stripping is probably still the best option.

Maybe the process of English-stripping or replacing text on the site could be refined, but I'm not sure English-stripping could be abandoned completely.

zorkian:
We don't get a lot of patch submissions as it is, so I'm optimizing for
making that easier even if it means we have to do a little more manual
work in the rare occasion people are working on the same things at the
same time. That doesn't happen very often.

pauamma:
Given the turn the discussion took, I believe it would better to have it on dw_dev rather than buried here.

zorkian:
Yeah, let's clarify: this isn't a discussion. I was providing a response
but I've made a decision about the future of this functionality. When
I'm ready to actually enact a policy change I'll post to dw_dev but I'm
not ready yet. Until then, it's business as usual with English
stripping.

zorkian:
I'm not going to block forward progress on English stripping, I'll post to dw_dev today. Merging this for now.

PR #2231, opened 2017-10-29
Fix adult interstitial

zorkian:
This broke when we switched to SSL everywhere. Fix it. Also add a flag
that allows us to just say 'yes, everything is SSL' which is useful.


zorkian:
This is live in prod. Merging.

PR #2234, opened 2017-11-13
Remove AOL information from user profile

srukle:
I removed AOL information from the profile page and also removed the AIM icon.


zorkian:
RIP AIM, we knew ye well.

PR #2237, opened 2017-11-19
Pass ssl option to internal_redir routing call.

### LINKED ISSUE #2233: Trying to access a private entry sends the user into a 302 Redirect loop
Steps to reproduce:

1. Create a private entry in a journal. I created a test post with “This is a test post” as a title and “Show this entry to: Private (Just You)”.
2. Send that link to somebody who doesn’t have journal access. e.g. https://alexwlchan.dreamwidth.org/2896.html
3. Attempt to access the entry.

In this case, I’d expect the returned HTTP status to be one of:

* 404 Not Found – deny any knowledge of the post’s existence. If a user isn’t authorised to see a post, they can’t know anything about it, even whether it exists.
* 401 Unauthorized or 403 Forbidden – with an appropriate error message.

In general I favour the 404 approach, but either one is probably fine. I’m not sure what information is encoded in the post IDs, which look entirely numeric, so admitting the existence of a post I can’t see may not be hugely problematic.

I *don’t* expect the server to return a 302 Redirect to *the same page*, which is the current behaviour. This sends a browser into an infinite redirect loop, and the user doesn’t get a helpful error messages.

Example responses, from a private post, and from a non-existent post (edited for brevity):

```console
$ curl --head "https://alexwlchan.dreamwidth.org/2896.html"
HTTP/1.1 302 Moved Temporarily
Content-Type: text/html; charset=iso-8859-1
Connection: keep-alive
Date: Sun, 05 Nov 2017 23:43:29 GMT
Server: Apache/2.4.7 (Ubuntu) mod_apreq2-20090110/2.8.0 mod_perl/2.0.8 Perl/v5.18.2
Location: https://alexwlchan.dreamwidth.org/2896.html

$ curl --head "https://alexwlchan.dreamwidth.org/404.html"
HTTP/1.1 404 Not Found
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Date: Sun, 05 Nov 2017 23:49:17 GMT
Server: Apache/2.4.7 (Ubuntu) mod_apreq2-20090110/2.8.0 mod_perl/2.0.8 Perl/v5.18.2
```

alierak:
Fixes #2233

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -1339,7 +1339,7 @@ sub journal_content
 
     # check for redirects
     if ( $opts->{internal_redir} ) {
-        my $int_redir = DW::Routing->call( uri => $opts->{internal_redir} );
+        my $int_redir = DW::Routing->call( uri => $opts->{internal_redir}, ssl => $LJ::IS_SSL );
         if ( defined $int_redir ) {
             # we got a match; clear the request cache and return DECLINED.
             LJ::start_request();


alierak:
Tested in production, because why not.

zorkian:
LGTM! Thank you!

Issue #2219, opened 2017-09-02
more accurate text when displaying PM from suspended user


kareila:
I noticed recently, when we had a PM spammer that was widely reported, that after the user was suspended, the message in my inbox read "Reply from suspended user" but the PM was not _in reply to_ anything.  It would be more accurate to print "Message from suspended user" instead.

The strings are hardcoded in `cgi-bin/LJ/Event/UserMessageRecvd.pm`, lines 113 and 177.

PR #2245, opened 2017-12-14
Issue #2219 - Updates text when displaying PM from suspended user

hotlevel4:
Changes text from "Reply from suspended user" to "Message from suspended user"


zorkian:
Fixes #2219

PR #2252, opened 2017-12-26
Add tabindex= to month/day/year in entry form.

### LINKED ISSUE #2247: Incorrect tab order on old Create Entries page
https://www.dreamwidth.org/support/see_request?id=37982

> When selecting the "Edit Date" option on the old Post an Entry page (foo.dreamwidth.org/update) to choose a specific date and time for the entry to be posted with, using the Tab key to move through the fields of date and time allows me to select the month, day, and year, in order. The next press of the Tab key skips to the "View Thumbnails" link associated with icons instead of proceeding to the hour and minute fields, as I would expect it to do.

pauamma:
Fixes #2247

### DIFF
--- a/cgi-bin/LJ/HTMLControls.pm
+++ b/cgi-bin/LJ/HTMLControls.pm
@@ -26,6 +26,7 @@ use strict;
 # des-:
 # returns:
 # </LJFUNC>
+# It's the caller's responsibility to leave appropriate gaps if using tabindex
 sub html_datetime
 {
     my $opts = shift;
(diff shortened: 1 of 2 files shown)


pauamma:
Tested on Chrome (logged in) and Firefox (not logged in). Both give the expected tab navigation order as far as date&time are concerned, but I don't really understand tab navigation behavior expectations, so further testing by someone who does is welcome.

zorkian:
I think the code is good and this makes sense.

PR #2253, opened 2017-12-29
Change userlog.ip to varchar(45). Fixes #1818.

### LINKED ISSUE #1818: The "ip" column in the userlog table should be widened to accommodate IPv6 addresses
It is currently only 15 characters wide.  IPv6 addresses can be up to 45 characters, per discussion in #1393.  I'm seeing some IP mismatch errors in the logs from SecurityAttributeChanged.pm due to the truncation.

pauamma:
Minimally tested by checking the column was indeed set to varchar(45). Can't test further as dreamacks are IPv4-only. Anyone working outside those limits is welcome to test further, as hinted in https://github.com/dreamwidth/dw-free/issues/1818#issuecomment-354332036.

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -1770,7 +1770,7 @@ CREATE TABLE userlog (
     action        VARCHAR(30) NOT NULL,
     actiontarget  INT UNSIGNED,
     remoteid      INT UNSIGNED,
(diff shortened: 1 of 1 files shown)


zorkian:
This is going to be a little painful in prod. The main cluster has 10 million rows and widening a column means rewriting the table. We should expect that deploying this is going to be a couple minutes of table lock probably.

PR #2259, opened 2018-01-01
Make methods return sensibly for expunged users.

### LINKED ISSUE #2200: can't load support requests from D&P users
This bug was introduced when adding media quota usage to support requests.  Loading a request from a D&P user (example: https://www.dreamwidth.org/support/see_request?id=2650) now dies with an error on line 229 of DW/Media.pm.

The subroutines `get_active_for_user`, `get_quota_for_user`, `get_usage_for_user` , and `can_upload_media` should all quietly abort if asked to operate on a D&P user.

pauamma:
- Per bug spec, DW::Media->get_active_for_user, DW::Media->get_quota_for_user,
  DW::Media->get_usage_for_use, and LJ::User->can_upload_media now all return
  sensible values for an expunged user.
- This exposed a division by 0 in htdocs/support/see_request.bml, which I
  also fixed.

Fixes #2200

### DIFF
--- a/cgi-bin/DW/Media.pm
+++ b/cgi-bin/DW/Media.pm
@@ -11,7 +11,7 @@
 # Authors:
 #      Mark Smith <mark@dreamwidth.org>
 #
-# Copyright (c) 2010 by Dreamwidth Studios, LLC.
+# Copyright (c) 2010-2018 by Dreamwidth Studios, LLC.
 #
 # This program is free software; you may redistribute it and/or modify it under
 # the same terms as Perl itself.  For a copy of the license, please reference
@@ -189,6 +189,7 @@ sub get_upload_type {
 sub get_active_for_user {
     my ( $class, $u, %opts ) = @_;
     confess 'Invalid user' unless LJ::isu( $u );
+    return () if $u->is_expunged;
 
     # get all active rows for this user
     my $rows = $u->selectcol_arrayref(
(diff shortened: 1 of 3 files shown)


cgi-bin/DW/Media.pm
```diff
@@ -11,7 +11,7 @@
 # Authors:
 #      Mark Smith <mark@dreamwidth.org>
```
zorkian: You can add your name to the authors list when you're editing files, if you want. Up to you!

htdocs/support/see_request.bml
```diff
@@ -289,7 +289,7 @@ body<=
             my $media_usage = DW::Media->get_usage_for_user( $u );
             my $media_quota = DW::Media->get_quota_for_user( $u );
             my $megabytes  = sprintf( "%0.3f MB", $media_usage / 1024 / 1024 );
-            my $percentage = sprintf( "%0.1f%%", $media_usage / $media_quota * 100 );
+            my $percentage = sprintf( "%0.1f%%", $media_usage / ( $media_quota || 1 ) * 100 );
```
zorkian: In this case since we're just printing it, I'd rather see something like 'if quota == 0, then percentage == `no quota`' or something?

cgi-bin/DW/Media.pm
pauamma: http://wiki.dreamwidth.net/wiki/index.php/File_headers#Authors_list says it's too minor a change for me to. Are you proposing an official policy change?

zorkian: Those are guidelines. If it was getting out of hand I'd push back, I just figured I'd mention it since I hadn't seen you do that in a while so I wasn't sure if you were considering it or not.

pauamma:
OK, this should take care of your change request.

zorkian:
Thank you!

PR #2265, opened 2018-01-04
Basic API + sample routes

rshatch:
This includes the changes requested in the other pull request, except for the inclusion of $class in the construction methods because I'm not clear what that's doing?

The YAML load *should* only be called from within the API endpoint defs, and it's called with a directory path prepended to the file, so it only loads things within api/ rather than any yaml anywhere.


rshatch:
Also going over the code again, I think we may want to merge rest_ok and rest_error into something more generic, or just predefine error text or something.

cgi-bin/DW/Controller/API/REST.pm
```diff
@@ -0,0 +1,229 @@
...
+
+# A generic API method dispatcher, for use in registering API
+# endpoints to the routing table. When called, looks up the handler
+# defined in the resource object for that HTTP action and calls it
+# or returns an error response if it's not implemented.
+
+sub _dispatcher {
+
+    my ( $self, @args ) = @_;
+    warn Dumper(@args);
```
zorkian: debugging

cgi-bin/DW/API/Parameter.pm
```diff
@@ -0,0 +1,79 @@
...
+sub define_parameter {
+    my $args = $_[0];
+    my $parameter = {
+        name => $args->{name},
+        desc => $args->{description},
+        in => $args->{in},
+        type => $args->{type},
+        required => $args->{required}
+    };
+    return bless $parameter;
```
zorkian: I left a comment on the last diff, can we update these to use the preferred style like in DW::Media etc? Basically something like,

```
sub define_parameter {
    my ( $class, $args ) = @_;
    my $parameter = { ... };
    return bless $parameter, $class;
}
```

The reason for this is that it makes it explicit that you're blessing into a certain class and also the calling convention (using the `->` notation) is more idiomatic.

cgi-bin/DW/Controller/API/REST.pm
rshatch: gah, I thought I got them all!

cgi-bin/DW/API/Parameter.pm
rshatch: How would calling the new variant work? That was the part that really confused me - I didn't know how to make it actually work and not just break horribly.

bin/upgrading/update-db-general.pl
```diff
@@ -3056,6 +3056,40 @@
 )
 EOC
 
+register_tablecreate("api_key", <<'EOC');
+CREATE TABLE `api_key` (
+  `userid` int(10) unsigned NOT NULL,
+  `keyid` int(10) unsigned NOT NULL,
+  `hash` varchar(32) UNIQUE NOT NULL,
```
zorkian: char(32)

cgi-bin/DW/API/Key.pm
```diff
@@ -0,0 +1,191 @@
...
+use warnings;
+use Carp;
+
+# Usage: new_for_user ( user ) 
+# Creates a new API key for a given user, saves it to DB, 
+# and returns the new key object.
+sub new_for_user {
+    my ($self, $u) = @_;
+    my $user = LJ::want_user( $u )
+        or die "need a user!\n";
```
zorkian: croak?

cgi-bin/DW/API/Key.pm
```diff
@@ -0,0 +1,191 @@
...
+
+    my $id = LJ::alloc_user_counter( $user, 'B' )
+        or croak 'Unable to allocate user counter for API key.';
+
+    my $key = LJ::rand_chars( 32 );
+    my $dbw = LJ::get_db_writer() or die "Failed to get database";
```
zorkian: Needs to be `LJ::get_cluster_writer()`? Or `$u->writer`? Need to check

cgi-bin/LJ/DB.pm
```diff
@@ -61,7 +61,7 @@ $LJ::DBIRole = new DBI::Role {
                     "logprop_history", "import_status", "externalaccount",
                     "content_filters", "content_filter_data", "userpicmap3",
                     "media", "collections", "collection_items", "logslugs",
-                    "media_versions", "media_props",
+                    "media_versions", "media_props", "api_key", "key_props",
```
zorkian: Remove api_key and key_props from here, let's not cluster these tables

cgi-bin/DW/API/Key.pm
```diff
@@ -0,0 +1,191 @@
...
+    my ($self, $u) = @_;
+    my $user = LJ::want_user( $u )
+        or die "need a user!\n";
+    my $memkey = [ $self->{keyhash}, "api_key:" . $self->{keyhash} ];
+
+    $self->valid_for_user($user) or die "key doesn't belong to user";
+
+    my $dbw = LJ::get_db_writer() or die "Failed to get database";
+    $dbw->do(
+        q{UPDATE api_key SET state = 'D' WHERE hash = ?},
```
zorkian: Today there's only two states, but I'd traditionally like to have the WHERE clause contain the from state we want to be true just to avoid race conditions. So, like `WHERE state = 'A' AND hash = ?`.

cgi-bin/DW/Routing.pm
```diff
@@ -556,6 +576,53 @@ sub register_api_endpoints {
...
+
+=back
+
+=cut
+
+sub register_api_rest_endpoint {
+    my ( $class, $string, $sub, $controller_class, %opts ) = @_;
+
+    # this is dev only for now -akp
+    return unless $LJ::IS_DEV_SERVER;
```
zorkian: do we want this still?

cgi-bin/DW/Routing/CallInfo.pm
```diff
@@ -79,7 +93,15 @@ sub call {
 
     my @args;
     @args = @{$opts->subpatterns} if ( $opts->subpatterns );
-    $opts->{__hash}->{sub}->( $opts, @args );
+    my $hash = $opts->{__hash};
+    # FIXME comment this
```
zorkian: looks like you did

doc/raw/memcache-keys.txt
```diff
@@ -127,3 +127,6 @@ pop_interests == array of interest keywords and counts
 poll:<pollid> = arrayref of poll properties, see cgi-bin/LJ/Poll.pm for details (section under "use base 'LJ::MemCacheable'")
 
 invite_code_try_ip:<ip> = stores a value of 1 for the IP address of the person who tried an invalid invite code (for rate limiting)
+
+<keyhash> api_key:<keyhash> = hashref of keyid, hash, and owning userid
```
zorkian: if you remember, would be curious why you are hashing on `<keyhash>` here

htdocs/manage/emailpost.bml
```diff
@@ -429,9 +454,31 @@ body<=
...
+    $ret .= "<h1>$ML{'.api.header'}</h1>";
+    $ret .= "<p>$ML{'.api.text'}</p>";
+    if (@{ $apikeys }) {
+        $ret .= "<table border=\"0\">";
+        my $count = 0;
+        for my $key (@$apikeys) {
+            $ret .= '<tr><td>';
+            $ret .= $key->{keyhash};
+            $ret .= '</td><td>';
+            # TODO: actually support delete
```
zorkian: # TODO: is delete supported? looks like it is

htdocs/manage/emailpost.bml
```diff
@@ -367,6 +369,28 @@ body<=
                     : "<div class='error-box'>$ML{'.reply.status.error'}</div>";
     }
 
+    if ( LJ::did_post() && $POST{"action:apikey"}) {
+        $ret .= DW::API::Key->new_for_user($u)
```
zorkian: need to check form auth

htdocs/manage/emailpost.bml
```diff
@@ -367,6 +369,28 @@ body<=
...
+                    ? "<div class='highlight-box'>$ML{'.api.status.success'}</div>"
+                    : "<div class='error-box'>$ML{'.api.status.error'}</div>";
+    }
+
+    if ( LJ::did_post() && $POST{"action:delete"}) {
+        my $deleted_keys .= $POST{"delete"};
```
zorkian: need to check form auth

cgi-bin/DW/Controller/API/REST.pm
```diff
@@ -0,0 +1,233 @@
...
+use DW::API::Method;
+use DW::API::Key;
+use JSON;
+use YAML::XS qw'LoadFile';
+
+use Carp qw/ croak /;
+
+our %API_DOCS = ();
+our %TYPE_REGEX = (
+    string => '([^/]*)',
```
zorkian: do we want `+` instead of `*`?

cgi-bin/DW/Controller/API/REST.pm
```diff
@@ -0,0 +1,233 @@
...
+
+sub _dispatcher {
+
+    my ( $self, @args ) = @_;
+
+    my ( $ok, $rv ) = controller( anonymous => 1 );
+    return $rv unless $ok;
+    
+    my $r = $rv->{r};
+    my $apikey = DW::API::Key->get_key($r->header_in('api_key'));
```
zorkian: `X-DW-APIKey`? or something?

cgi-bin/DW/Controller/API/REST/Spec.pm
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/perl
+#
+# DW::Controller::API::REST::Icons
+#
+# API controls for the icon system
```
zorkian: does not seem to be

cgi-bin/DW/Controller/API/REST/Comments.pm
```diff
@@ -0,0 +1,56 @@
...
+
+
+sub get_screening {
+    my $self = $_[0];
+    
+    # we want to handle the not logged in case ourselves
+    my ( $ok, $rv ) = controller( anonymous => 1 );
+    return $rv unless $ok;
+
+    my $r = $rv->{r};
```
zorkian: Doesn't look like we need the ok/rv/controller/r stuff in this endpoint.

cgi-bin/DW/Controller/API/REST/Icons.pm
```diff
@@ -0,0 +1,54 @@
...
+use warnings;
+use DW::Routing;
+use DW::Request;
+use DW::Controller;
+use JSON;
+
+my $icons_all = DW::Controller::API::REST->path('icons_all.yaml', 1, {'get' => \&rest_get});
+
+sub rest_get {
+    my ( $self, $opts, $username, $picid ) = @_;
```
zorkian: $picid? doesn't seem to be used in this endpoint

cgi-bin/DW/Controller/API/REST/Icons.pm
```diff
@@ -0,0 +1,54 @@
...
+
+    my $u = LJ::load_user( $username );
+
+    # we want to handle the not logged in case ourselves
+    my ( $ok, $rv ) = controller( anonymous => 1 );
```
zorkian: move this above the user load, just in case it returns error and we bail -- might as well not load a user in that case

zorkian: or just don't need it at all :)

cgi-bin/DW/Controller/API/REST/Icons.pm
```diff
@@ -0,0 +1,54 @@
...
+    return $rv unless $ok;
+
+    # if we're given a picid, try to load that userpic
+    if ($picid != "") {
```
zorkian: endpoint doesn't seem to take a picid according to the YAML?

cgi-bin/DW/Controller/API/REST/Spec.pm
```diff
@@ -0,0 +1,71 @@
...
+use DW::Request;
+use DW::Controller;
+use JSON;
+
+# Define route and associated params
+my $spec = DW::Controller::API::REST->path('spec.yaml', 1, {'get' => \&rest_get});
+
+sub rest_get {
+    my $self = $_[0];
+    my ( $ok, $rv ) = controller( anonymous => 1 );
```
zorkian: remove controller

cgi-bin/DW/Controller/API/REST/Spec.pm
```diff
@@ -0,0 +1,71 @@
...
+
+}
+
+sub _spec_20 {
+    my $self = $_[0];
+    my $ver = $spec->{ver};
+
+    my @content_types = qw(application/json);
+    my @schemes = qw(https);
+    my $security_defs = { "api_key" => {"type" => "apiKey", "name" => "api_key", "in" => "header" }};
```
zorkian: probably also need to update header here

cgi-bin/DW/Controller/API/REST/Spec.pm
```diff
@@ -0,0 +1,71 @@
...
+ 
+    my %spec = (
+        swagger => '2.0',
+        info => {
+            version => "$ver",
+            title => "$LJ::SITENAME API",
+            description => "An OpenAPI-compatible API for $LJ::SITENAME"
+
+        },
+        host => $ENV{SERVER_NAME} ||  $ENV{HTTP_HOST},
```
zorkian: Hmm, $LJ::DOMAIN? or $LJ::DOMAIN_WEB?

momijizukamori:
@zorkian I think this rube goldberg machine is ready for another look

cgi-bin/DW/API/Key.pm
zorkian: This didn't get done.

cgi-bin/DW/Routing.pm
zorkian: And this one, do we still want this? Or, maybe maybe this some `$LJ::ENABLE_REST_API`?

doc/raw/memcache-keys.txt
zorkian: We should remove the first `<keyhash>`, there's no reason to hash on the hash.

htdocs/manage/emailpost.bml
zorkian: 🆗

htdocs/manage/emailpost.bml
zorkian: 🆗

zorkian: Need to abort the flow if form auth is invalid so we don't do the action anyway.

htdocs/manage/emailpost.bml
zorkian: Same here, abort flow.

rshatch:
@zorkian and we're good to go!

PR #2270, opened 2018-01-06
Add userlinks => 100 to %CAP_PREMIUM. Fixes #2181.

### LINKED ISSUE #2181: Incorrect number of links in Links List for premium/seed accounts
https://www.dreamwidth.org/support/see_request?id=36733

FAQ says 100 but current limit is 50.

pauamma:


pauamma:
Tested by creating 51 links (that was tedious enough).

zorkian:
I think we'll have to make this change in prod too. I've made that change, it should go live next time we push out configs.

PR #2274, opened 2018-01-08
Remove opt_nomodemail and migrate_prop_to_esn.

### LINKED ISSUE #2052: remove opt_nomodemail userprop
Looks like `opt_nomodemail` is a legacy userprop from the days before ESN, which means it was never effectively used on DW.

I believe removing this will also let us remove the `migrate_prop_to_esn` user method.

pauamma:
Fixes #2052.


zorkian:
Dead code is happy code!

PR #2278, opened 2018-01-09
Global backend changes to allow converting /customize to TT

rshatch:
Otherwise known as 'the stuff that might break other parts of the site'.
-Allow template recursion
-Add a TT form element for color pickers
-Allow empty values in forms to make it to controllers (so that options can be set to nothing)


rshatch:
I just realized I need to make a tweak to error_ml too, so don't merge this until get that done, heh.

rshatch:
Okay, this should be good to go now.

PR #2283, opened 2018-01-10
Customize/advanced to tt

rshatch:
The less-breaking changes for /customize/ conversion. Does, however, require the global changes from https://github.com/dreamwidth/dw-free/pull/2278 and thus may need a rebase after those are merged.


zorkian:
boom!

PR #2288, opened 2018-01-13
Spellcheck gmail.com and other common domains.

### LINKED ISSUE #993: Detect misspellings of gmail, etc
Our current check_email sub looks for misspellings of hotmail, aol... and nothing else. We should probably update this ;)

(bonus if we can use some kind of distance algorithm to detect possible misspellings)

pauamma:
Fixes #993.


zorkian:
We can call 3 our magic fuzzy number, which is good. XKCD already proved 4 was random so this all makes sense.

PR #2295, opened 2018-01-16
Cleanup 2017q2

kareila:
Old cleanup branch that was never submitted until now.

We must not forget that cats are special.

Please don't squash this branch as the individual commit messages contain the errors fixed by each commit.


zorkian:
Thanks for the reminder on the squashing. :)

And yes, cats are special. Well, most of them.

PR #2298, opened 2018-01-18
Remove S1 memcache keys. Fixes #2255.

### LINKED ISSUE #2255: Get rid of remaining references to S1 memcache keys
The following S1 memcache keys are referred to in cgi-bin/LJ/User/Data.pm (wipe_major_memcache), htdocs/admin/memcache_view.bml, and or doc/raw/memcache-keys.txt, but not used anywhere else:

- s1overr
- s1uc
- s1pubstyc
- s1styc
- s1style
- s1style_all
- s1stylemap

Let's get rid of them there too.

pauamma:


pauamma:
Tested as far as I could, which isn't a whole lot given that Dreamhacks don't have memcache installed.

Sophira:
The Dreamhack machine does have `memcached` installed and usable, but it's not set up for Dreamhacks by default. I'd link to the wiki page on it, but the wiki seems to be down right now and I don't know the precise setup procedure myself.

I can give you a port number to use with it, though; I'll do that later today, so that when the wiki's back up, one of the steps is already done.

alierak:
The wiki's fine, it's just http-only for some reason. I'm not sure if it ever supported https and can't log into Cloudflare to check.

pauamma:
Will wait for port # assignment, then.

pauamma:
OK, tested a bit more in depth. Seems to work.

zorkian:
I'm impressed these lived this long. Nice axing!

PR #2299, opened 2018-01-18
Remove dogeartime from userprops. Fixes #2223.

### LINKED ISSUE #2223: Consider removing userprop dogeartime from bin/upgrading/proplists.dat
Do we remove props no longer used in the code? If so, that's a candidate, as far as I can tell.

pauamma:


pauamma:
Tested by running commands in http://wiki.dreamwidth.net/wiki/index.php/Dev_Maintenance#Update_your_database

zorkian:
I'm not 100% sure removing props is "safe" in the sense that it might leave them in the database for the future, but it's probably not going to bite us... and if it does it'll be interesting.

PR #2302, opened 2018-01-26
2215/state sort

rshatch:
Bless whoever wrote the proper sort code in Controller/MiscLegacy, it made this a lot easier.


zorkian:
Took me way too long to figure out what 'states' meant. I was thinking like PENDING, ACTIVE, MAINTENANCE which are states from $dayjob and was like...what is happening...help...

rahaeli:
@zorkian so you're saying you were in ... a state of confusion?

I'll see myself out.

PR #2309, opened 2018-02-05
[#2308] tweaks to Markdown username conversion

### LINKED ISSUE #2308: user names in markdown now fail to convert if they happen at the beginning of a line
The application of the fix in #2167 changed the order of execution in `LJ::CleanHTML::clean_as_markdown`, such that the regular expression is checked before running `Text::Markdown::markdown` on the input string.

This has caused one of the tests in t/cleaner-markdown.t to start failing.  When the input was cleaned first, paragraph tags were added and the string being checked was `<p>@system</p>` which converted fine. In the revised code, we are passing in `@system` as a bare string which fails to match.

The obvious approach of adding `^|` to the lookbehind block fails with a "Variable length lookbehind not implemented in regex" error.

kareila:
- Find and convert usernames at the beginning of the string -- fixes #2308.

- Print '\\@user' as '@user' not '\\@user' -- the Markdown conversion wasn't removing the escape character.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1755,12 +1755,18 @@ sub clean_as_markdown {
         }
     };
 
+    # First pass is just to look for an edge case where an unescaped
+    # username that needs to be converted is the first item in the string.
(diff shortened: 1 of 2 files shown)


zorkian:
Thanks for taking the regex 🗡 and making 🍖 🔥 😋

PR #2312, opened 2018-02-12
avoid db replication get/set race in saving community settings

### LINKED ISSUE #2014: Membership option won't save in Account Settings
https://www.dreamwidth.org/support/see_request?id=35720
http://www.dreamwidth.org/support/see_request?id=35631

If you go to Account Settings to change the membership option of your community, it just won't save. It appears to, but if you refresh the page it hasn't and the setting doesn't seem to have been applied (the about section profile page didn't change either).

alierak:
by supporting UPDATE of an individual setting in set_comm_settings.
fixes #2014

### DIFF
--- a/cgi-bin/DW/Setting/CommunityMembership.pm
+++ b/cgi-bin/DW/Setting/CommunityMembership.pm
@@ -74,9 +74,8 @@ sub save {
 
     my $remote = LJ::get_remote();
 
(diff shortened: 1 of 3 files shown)


alierak:
Note: please review for general idea! Not tested at all.

zorkian:
I believe this will work and I looked at it long enough that I'm willing to merge it and see how it goes.

Certainly the approach will fix the race condition between get/set we were seeing.

PR #2321, opened 2018-03-23
Remove obsolete text from create entries success

### LINKED ISSUE #1665: "View all entries from the same date" displayed perplexingly on Create Entry success landing page
(Semi-tested using Beta update page.)

When updating with "Use the current time" and "Don't show on Reading pages" checked, the second item in the "Now you can..." list was "View all entries from the same date".

When updating with "Use the current time" but _not_ "Don't show on Reading pages" checked, "View all entries from the same date" does not display.

When editing an entry with "Don't show on the Reading page" but _not_ "Use the current time" checked, "View all entries from the same date" does not display.

It's not clear to me why this might be intentional, or what is gained from this inconsistency due to these specific options.

kaberett:
Fixes #1665.

Somewhat skimpy testing makes it look as though this works for both the old update page and the new one.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -1043,16 +1043,6 @@ sub _do_post {
                 ml_string => ".new.links.view" }
         );
 
-        if ( $form_req->{props}->{opt_backdated} ) {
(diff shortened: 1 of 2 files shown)


zorkian:
Boom!

PR #2324, opened 2018-03-23
2316 support notifs https

kaberett:
This is literally a one-character change and I haven't tested it because I can't quite believe it's right and I'm dubious about my ability to get support notif e-mails working again in zero time.


kaberett:
NO. BAD GITHUB. TOO MANY COMMITS. NO.

kaberett:
FORCE PUSH MAKES EVERYTHING BETTER.

kaberett:
I can *in principle* put protocol checks in if we *want* to keep supporting HTTP...

zorkian:
I agree this would probably fix this case, but I'm worried about edge cases. `SITEROOT` is the kind of variable that is used by literally everybody everywhere.

I _think_ that changing this to https would be fine for us...

let me ponder this a bit and see if I can think of anything it would break.

zorkian:
Let's do it.

PR #2325, opened 2018-03-23
Make Quick Update respect minsecurity

chrisboyle:
Fixes: #1180

Once more with JavaScript. This leans heavily on data attributes to convey the minsecurity data and alternate translated labels (for community entries' access) into JS.

In the absence of JS, security will initially be set to match minsecurity, but you can still pick any security level (it will still be overriden on submit if necessary). This is because you might want to post to a community at a more public level than your own journal's minsecurity.

Worth noting that browsers (Chrome, at least) seem to canonicalise the names of data attributes to lowercase, so trying to camel-case them would only confuse matters. :-)


cgi-bin/LJ/Web.pm
```diff
@@ -1997,6 +1997,15 @@ sub entry_form_entry_widget {
 }
 
 
+sub minsec_for_user {
+    my $user = LJ::load_user( shift );
```
zorkian: For performance and style, we usually use `$_[0]` instead of `shift`. It ends up being slightly faster since we don't need to manipulate the underlying array that was used to pass the parameters in.

cgi-bin/LJ/Web.pm
```diff
@@ -1997,6 +1997,15 @@ sub entry_form_entry_widget {
...
+    if ( ! $user ) {
+        return undef;
+    }
```
zorkian: The preferred Dreamwidth style would be:

```perl
my $user = LJ::load_user( $_[0] )
    or return;
```

Almost always, `if !something` should be `unless something`. Also a single-line if statement should be a postfix conditional `return unless something` (dropping the `undef` which is the default return value).

cgi-bin/LJ/Web.pm
```diff
@@ -2017,26 +2026,52 @@ sub entry_form_postto_widget {
...
     return undef unless $res;
 
-    my @journals = map { $_, $_ } @{$res->{'usejournals'}};
+    LJ::need_res( { group => 'jquery' }, 'js/quickupdate.js' );
+
+    my @journals = map { {
+            value => $_,
+            text => $_,
+            data => { minsecurity => minsec_for_user( $_ ), iscomm => 1 }
+        } } @{$res->{'usejournals'}};
```
zorkian: Style: no optional quotes (such as around most keys), so `usejournals` alone is fine.

zorkian:
Merged. :D

PR #2329, opened 2018-03-25
1661 banned behaviour

swaldman3:
revision of PR 1693, incorporating requested style changes.

This *partially* fixes bug 1661, in that banned users can no longer comment through the main talkpost_do.bml Reply screen. They still can through quickreply, or probably email, because that appears to use totally separate logic and I don't understand where/how. But perhaps it's a start.

Please unassign me from 1661 once this is merged.


zorkian:
Issue #1661

zorkian:
Thank you for doing this! Will unassign you from the bug.

PR #2343, opened 2018-06-16
Removes awkward beta update whitespace if no icons

### LINKED ISSUE #2323: Beta update page userpic widget appears blank/confusing with no userpics
As a user with no icons, go to the beta update page:

userpic-picker-empty

The tiny square area can be scrolled down to show
"""
[blank lines]
Upload a
[more blank]
userpic
[further emptiness]
"""

Removing the class "icon" from the "upload a userpic" link fixes things, so we could just do that:

userpics-better

kaberett:
Fixes #2323.

... as far as I can tell, but I am definitely open to being told there's a less awful way of doing this.

### DIFF
--- a/views/entry/module-icons.tt
+++ b/views/entry/module-icons.tt
@@ -19,7 +19,7 @@ the same terms as Perl itself.  For a copy of the license, please reference
 
     <div class="row">
     <div class="columns">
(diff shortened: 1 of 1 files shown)


zorkian:
I think that seems reasonable to me, actually!

PR #2345, opened 2018-06-17
Allow email spell-checking to be overridden

### LINKED ISSUE #2304: Make the email spellcheck test a soft reject/hint instead of the hard reject it is currently.
See discussion in https://github.com/dreamwidth/dw-free/issues/993#issuecomment-360951407 and in the https://dw-dev.dreamwidth.org/203374.html?thread=1579886#cmt1579886 thread.

chrisboyle:
Fixes #2304, in the many, many places where we check email addresses.

Where possible, this gives you a tickbox near the input saying "Yes I'm sure this is correct".

The **display** email, set on the Edit Profile page, is simply excluded from spell-checking because that page doesn't have a way to hold unsaved edits while showing you a checkbox, and reworking it to do so looked too complicated, see 972d7b437ad365077ec3ed40c200a68c0dd6dee9. That page will also now tell you if your email address is more seriously invalid, rather than silently dropping that part of the change.

The sysbans page has similar issues. For that one I've made the ticky always visible up-front (if the type is email).

For the admin console command `email_alias`, you can append `!` to the address to achieve the override, and will be prompted to do so.

Also added `mail.com`, `ymail.com` to known-good domains as per original discussion on #993.

### DIFF
--- a/cgi-bin/DW/Controller/Admin/SupportCat.pm
+++ b/cgi-bin/DW/Controller/Admin/SupportCat.pm
@@ -113,7 +113,7 @@ sub category_controller {
             if $cat->{basepoints} < 0 || $cat->{basepoints} > 255;
         if ( $cat->{replyaddress} ) {
             my @errors = ();
-            LJ::check_email( $cat->{replyaddress}, \@errors );
+            LJ::check_email( $cat->{replyaddress}, \@errors, $post_args, \( $vars->{email_checkbox} ) );
             $errors->add_string( 'replyaddress', $_ ) foreach @errors;
         }
 
--- a/cgi-bin/LJ/User/Message.pm
+++ b/cgi-bin/LJ/User/Message.pm
@@ -489,6 +489,8 @@ sub check_email
 
     my $use_errcode = $opts{errcode};
 
+    my $force_spelling = $opts{force_spelling};
+
     # Trim off whitespace and force to lowercase.
     $email =~ s/^\s+//;
     $email =~ s/\s+$//;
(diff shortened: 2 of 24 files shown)


cgi-bin/DW/Controller/Admin/SupportCat.pm
```diff
@@ -113,7 +113,7 @@ sub category_controller {
             if $cat->{basepoints} < 0 || $cat->{basepoints} > 255;
         if ( $cat->{replyaddress} ) {
             my @errors = ();
-            LJ::check_email( $cat->{replyaddress}, \@errors );
+            LJ::check_email( $cat->{replyaddress}, \@errors, $post_args, \( $vars->{email_checkbox} ) );
```
zorkian: Maybe I haven't been doing enough Perl but what does the `\( ... )` syntax do?

zorkian: Oh this just takes a reference to the scalar inside? I see!

chrisboyle: Yep, that :-)

cgi-bin/LJ/User/Message.pm
```diff
@@ -497,18 +497,15 @@ sub check_email
     my $reject = sub {
         my $errcode = shift;
         my $errmsg = shift;
-        # TODO: add $opts to end of check_email and make option
-        #       to either return error codes, or let caller supply
-        #       a subref to resolve error codes into native language
-        #       error messages (probably via BML::ML hash, or something)
-        push @$errors, $use_errcode ? $errcode : $errmsg;
+        push @$errors, $errmsg if ref( $errors );
```
zorkian: nit: Try to avoid including unnecessary parenthesis.

zorkian:
Thank you for doing this, and especially for being so thorough to each of the places we touch emails. That's appreciated! Also some cleanup in date presentation in some places?

Anyway I read through the code and it seems sane, so I'm merging!

chrisboyle:
@zorkian Oh, yes, the date presentation bit was: when choosing a delivery date in the shop, and re-rendering the form (with previous inputs) to show an error, if the month or the day was a single digit, we were passing in e.g. `"2018-7-1"` to the datetime widget which was parsing it with `/\d{4}-\d{2}-\d{2}/` and therefore showed up blank. Now it keeps your input. :-)

zorkian:
Nice!!! Thank you for fixing that :)

PR #2347, opened 2018-06-17
Prevent creation of access filter names with commas

### LINKED ISSUE #2262: Enforce new "no commas in trust group names" rule
Per discussion in #2201, commas (and other punctuation?) should no longer be permitted in trust group names. What should happen to existing groups?

chrisboyle:
Fixes #2262 (does nothing about existing filter names).

### DIFF
--- a/htdocs/manage/circle/editfilters.bml
+++ b/htdocs/manage/circle/editfilters.bml
@@ -59,6 +59,16 @@ _c?>
             return;
         }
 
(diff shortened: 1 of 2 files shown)


zorkian:
Thank you for tackling this!

It's almost a shame we made it so permissive to begin with. Being able to squash it down to a whitelist of letters/numbers/spaces would be nice, but would break so many use cases. Oh well.

PR #2349, opened 2018-06-17
Clarify Support "No HTML allowed" messages

### LINKED ISSUE #2209: Better wording for 'no html allowed' on support request page
https://www.dreamwidth.org/support/see_request?id=37019

/support/submit says not to use HTML in your submission, making it confusing for users who need to paste some code to explain to us what their pb is or to provide embed codes. As ysobel suggested we could say that HTML will show up as bare code or something simple like that.

kaberett:
Fixes #2209.

(NB the line on submit.tt ends up quite lengthy, rather than wrapping at ~textbox width, but I'm... not dealing with that unless anyone is particularly keen on it.)


zorkian:
I think that's okay if they don't wrap together.

Issue #2227, opened 2017-10-06
Remove AIM from user profiles


kareila:
AOL Instant Messenger is being retired in December: https://techcrunch.com/2017/10/06/aol-instant-messenger-shut-down/

srukle:
Claimed at #2234.

pauamma:
Since the associated PR was merged, can this be closed?

PR #2355, opened 2018-07-18
[#2227] remove remaining references to AIM

kareila:
The profile display was removed in #2234, but the Edit Profile page
still contained the AIM field in the form, and I found a few other
references to the defunct userprop sprinkled throughout odd corners
of the codebase.


zorkian:
Werps.

Issue #2358, opened 2018-07-22
add Discord to "Other Sites" list on user profile


kareila:
Requested: https://dw-maintenance.dreamwidth.org/77357.html?thread=2874157#cmt2874157

zorkian:
Yessssssss

kareila:
Also: https://dw-suggestions.dreamwidth.org/1499325.html

sklei4:
claiming

kareila:
This was resolved in #2363.

PR #2363, opened 2018-10-10
add discord to profile #2358

srukle:
Like [https://dw-suggestions.dreamwidth.org/1499325.html] recommended, I left out the URL so that the username should appear black without a link leading the user to a new account. We could setup a way to message the account directly, but I'm not sure if the behavior is always deterministic. Finding your own user identity number too could be a challenge. This is the easiest way I can imagine this working.


zorkian:
thank you for doing this!!!!

PR #2368, opened 2018-12-07
Attempt to make Travis run compiler

zorkian:
Ignore pls for now


alierak:
Failing, but nifty feature if we can use it.

PR #2370, opened 2018-12-08
Use display_name in place of username when banning/unbanning a user

### LINKED ISSUE #1691: Banning an OpenID user exposes their ext_?? username.
To reproduce:
- Log in as a user
- Hover over an OpenID user's usericon, and select "Ban User"
- Note that the resulting popup, which asks "Are you sure you want to ban $user", uses the ext_?? username for the OpenID user rather than their friendly one. It probably shouldn't.

Only tested on dreamhack.

Darael:
Fixes #1691.

May require testing; I do not presently have an installation or a dreamhack
set up.


zorkian:
thank you!

PR #2372, opened 2018-12-08
Convert /latest to TT and spruce it up a little.

rshatch:
Two quick notes - first, I opted not to move the existing DW/LatestFeed.pm file because a bunch of other stuff references it and it seemed too easy to miss updating references. Second, I'd really like someone to double-check that the memcache functionality is working properly before this is pushed - I'm getting blank pages back and I don't know if it's something wrong in the code, or something wrong with memcache.


rshatch:
Did some clean-up and moved to using our already-written template caching function. Unfortunately there seems to be something weird with it where it's losing the CSS include. I did some work to debug, and the only difference I could find between the cached and non-cached render paths was that the cached one seemed to be using a different $scheme_engine, but I'm not sure why or how it affects things.

cgi-bin/DW/Controller/LatestFeed.pm
```diff
@@ -0,0 +1,176 @@
...
+    # which case it is going to be in memcache
+    my $mckey = "latest_src:$type:$max:$fmt" . ( $feed ? ":$feed" : '' ) . ( $tag ? ":$tag" : '' );
+
+    my $cache_opts = {
+        expire => 60,
+    };
+
+
+    my $ret = DW::Template->render_cached_template( $mckey, 'latest/index.tt', \&generate_vars, $cache_opts );
+    #my $ret = DW::Template->render_template('latest/index.tt', $vars);
```
zorkian: debugging?

cgi-bin/DW/Controller/LatestFeed.pm
```diff
@@ -0,0 +1,176 @@
...
+    my $truncated;
+    my $evt = $entry->event_html_summary( 2000, { cuturl => $url, preformatted => $entry->prop( 'opt_preformatted' ) }, \$truncated );
+    # put a "(Read more)" link at the end of the text if the entry had to be shortened
+    $evt .= ' <a href="' . $url . '">(Read more)</a>' if $truncated;
+    return $evt;
+}
+
+sub generate_vars {
+
+   my ( $ok, $rv ) = controller( anonymous => 1 );
```
zorkian: nit: three spaces?

zorkian:
thanks for converting this! :)

PR #2373, opened 2018-12-09
Move stc include to the controller so it's loaded for cached version.…

rshatch:
… Also, remove debug code.


zorkian:
You'll have to rebase, I already removed the debug code.

PR #2382, opened 2018-12-13
Remove fastclick.js and references to it. Fixes #2381

### LINKED ISSUE #2381: remove fastclick.js
Located in `htdocs/js/foundation/vendor/fastclick.js`

@momijizukamori says: "Turns out the issue with dropdowns not opening in mobile FF is that they require double-clicks... I'm seeing scattered reports online that it's an issue with the FastClick library which Foundation vendors in... and it looks like FastClick is in fact the culprit. Comment it out of the foundation resource group = dropdowns work normally again both in Chrome and FF."

> Note: As of late 2015 most mobile browsers - notably Chrome and Safari - no longer have a 300ms touch delay, so fastclick offers no benefit on newer browsers, and risks introducing bugs into your application. Consider carefully whether you really need to use it.

rshatch:


rshatch:
In addition to the thread linked in the original issue, I poked around some at our fancier Foundation pages (beta create entry, community management) on a few different browsers and I don't see any errors or bugs.

zorkian:
Thanks for investigating this and fixing it :)

PR #2388, opened 2019-01-02
Remove Delicious from 'Other Services' on profile. Fixes #2376

### LINKED ISSUE #2376: more "Other Sites" upkeep: remove Delicious
Delicious has been gone for a while and most of their hardcore users moved to Pinboard.  I'm not suggesting we automatically migrate usernames, but the URL displays should be similar if not identical.

rshatch:


kareila:
No, that code renders user tags for accounts on the site. Separate from profile data.

PR #2389, opened 2019-01-02
Remove Yahoo Messanger from user profiles. Fixes #2356

### LINKED ISSUE #2356: remove Yahoo Messenger from user profiles
Another one bites the dust: Yahoo Messenger shut down today.

Elements to be removed should be similar to what was done for #2227.

rshatch:


bin/upgrading/deadphrases.dat
```diff
@@ -1852,3 +1852,10 @@ general email.newacct5.body
 general widget.createaccount.error.birthdate.invalid
 
 general /profile.bml.details.userpics
+
+general /manage/profile/index.bml.services.delicious
+general /profile.bml.service.delicious
```
Darael: These two lines seem to be duplicated from #2388

PR #2394, opened 2019-01-26
Replace old S2 layer editor with CodeMirror

### LINKED ISSUE #1913: Replace S2 layer editor with something slightly more modern
Because right now, it's a very thin BML wrapper around some 10+ year old Javascript we inherited from LJ, and it's always had some weird behavior (ie, pressing tab sends focus to the compiler output window but *also* inserts two tab characters, etc). And it's only slightly more work to put in something new than try and convert the existing stuff to TT and modern JS. 

We had a bug for this but it was lost to the great Bugzilla deletion abyss. https://ace.c9.io looks promising but I vaguely recall there was a reason against either this or one of it's parent projects originally?

rshatch:
-Convert /customize/advanced/layeredit to TT
-Fix typos in customize/advanced controller
-Fix layeredit sidebar reference links
-Change layeredit text area to CodeMirror editor
-Add highlighting and autocomplete implementations for S2.
-Make bin/gens2lib.pl actually work again and generate the library off core2 instead of core1.

Fixes #1913


zorkian:
Code looks good but @rshatch is going to investigate CSS highlighting issue and an issue with saving.

zorkian:
Can you also add `List::Util==1.45` in the dependencies file and see if that resolves the build issue?

rshatch:
currently returning valid compile for invalid layers, need to investigate that. Also, looks like comment highlighting may have borked CSS highlighting?

zorkian:
Hmm @momijizukamori why is this build failing? it looks like YAML::XS is in the dependencies but it's saying it's not? D:

momijizukamori:
Actually it looks like it got removed? idk if that's in my base branch or part of the merge from develop though

PR #2409, opened 2019-03-24
Bug 2018 Fix for Abstractia themes background images

wahiaronkwas:
Fixes bug 2018. Ads the CSS property background-size: cover to themes that require the image background to cover the entire screen area. On larger screens, the image ended just short of the window.


wahiaronkwas:
I have a problem with my dreamhack.

PR #2414, opened 2019-04-10
[#2410] Make less-quick-reply stop trashing the layout on mobile

### LINKED ISSUE #2410: "Less-quick reply" form blows out the layout on mobile
Quick-reply forms are built by [quickreply.tt](https://github.com/dreamwidth/dw-free/blob/develop/views/journal/quickreply.tt), and for whatever historical reason, there's two versions available: 

- "quicker reply," added in #1078, which is used on multi-entry pages ("reading" and "recent"). This is fine.
- "less-quick reply," which is used on single-entry pages. This is pretty bad! It's a big `<table>`, and it always overflows and trashes the page layout on mobile. It's even wide enough that it's hard to type in, because of the way iOS Safari handles textarea focus. 

<a href='https://roadrunnertwice.dreamwidth.org/file/9961.png'></a>

First question: Can I just refactor the markup in quickreply.tt (and wherever the CSS for that comes from) and lay it out more like the quicker reply? Or is there some hidden requirement that would make that a no-go?

Second question: If the markup does have to stay put, is there some way I can add some CSS on top of every layout except the specifically problematic ones? 

Here's how I fixed this in my own journal style:

```css
#qrform > table,
#qrform > table > tbody {
    /* "the de-table-izing ray! I can dodge this!" */
    display: block;
    /* NARRATOR: but he couldn't. */
}

#qrform > table tr {
    display: flex;
    flex-direction: row;
    flex-wrap: wrap;
    justify-content: space-between;
    align-items: center;
}

#qrform > table td {
    /* dissolve <td> elements, treat their children as direct children of <tr>s */
    display: contents;
}

#qrform > table #subject {
    max-width: 90%;
}

#qrform > table tr > td > * {
    margin-top: .3rem;
}

/* tweak for misaligned quote button */
#qrform > table span#quotebuttonspan {
    display: contents;
}
#qrform > table span#quotebuttonspan > input {
    margin-top: .3rem;
}

#qrform > table textarea#body {
    margin-bottom: .3rem; /* extra room before "post comment" */
}
```

<a href='https://roadrunnertwice.dreamwidth.org/file/10492.png'></a>

nfagerlund:
Resolves #2410.
Fixes one instance of #2421.

This PR rewrites the template for the "less-quick" quick-reply so that it uses a flexible layout similar to the "quicker" quick-reply. To keep things sleek, it moves the "quote" and "random icon" buttons below the body textarea.

This tweaks some tiny amounts of perl (to make the required CSS available on single-entry pages) and javascript (to update the icon preview when the random button fires). 

It also tweaks some less-tiny amounts of SCSS to make sure things display nicely as the page moves up and down the range of sizes.

#### Gif

resize2

#### Stills

quick-er reply (this branch): 

image

less-quick reply (this branch):

image

less-quick reply (mobile, this branch):

### DIFF
--- a/htdocs/js/jquery.quickreply.js
+++ b/htdocs/js/jquery.quickreply.js
@@ -208,6 +208,7 @@ jQuery(document).ready(function($) {
         // take a random number, ignoring the "(default)" option
         var randomnumber = Math.floor(Math.random() * (iconslist.length-1) ) + 1;
         iconslist.selectedIndex = randomnumber;
+        $( iconslist ).trigger("change");
     });
 });
 
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -2232,12 +2232,8 @@ sub init_s2journal_js {
             js/jquery/jquery.ui.widget.js
             js/jquery.quickreply.js
             js/jquery.threadexpander.js
-        ) ) unless $opts{noqr};
-
-    # this is only used on lastn-type pages (DayPage, RecentPage, FriendsPage)
-    LJ::need_res( { group => "jquery" }, qw(
             stc/css/components/quick-reply.css
-        ) ) if $opts{lastn};
+        ) ) unless $opts{noqr};
 
     # load for userpicselect
     LJ::need_res( init_iconbrowser_js( 1 ) ) if $opts{iconbrowser};
--- a/htdocs/js/jquery.iconselector.js
+++ b/htdocs/js/jquery.iconselector.js
@@ -15,7 +15,7 @@
                 });
         }
 
-        return $(this).wrap("<span class='iconselect_trigger_wrapper'></span>");
+        return $(this);
     };
 
     // selected icon
--- a/htdocs/scss/components/quick-reply.scss
+++ b/htdocs/scss/components/quick-reply.scss
@@ -65,10 +65,42 @@
 
     .qr-meta {
         margin-left: 55px;
+        display: -webkit-flex;
+        -webkit-flex-direction: row;
+        -webkit-flex-wrap: wrap;
+
+        display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
+
+        &>* {
+            -webkit-flex-grow: 1;
+            -webkit-flex-shrink: 1;
+
+            flex-grow: 1;
+            flex-shrink: 1;
+        }
+    }
+
+    .qr-footer {
+        display: -webkit-flex;
+        -webkit-flex-direction: row;
+        -webkit-flex-wrap: wrap;
+
+        display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
+    }
+
+    span#quotebuttonspan {
+        // This only exists to get something appended to it by jQuery. Its child
+        // input is the real flex item.
+        display: contents;
     }
 
     textarea, select, input[type="text"] {
-        width: 100%;
+        max-width: 100%;
+        flex-basis: 20%;
     }
 
     select, input, textarea {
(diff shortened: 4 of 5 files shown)


htdocs/js/jquery.quickreply.js
```diff
@@ -208,6 +208,7 @@ jQuery(document).ready(function($) {
         // take a random number, ignoring the "(default)" option
         var randomnumber = Math.floor(Math.random() * (iconslist.length-1) ) + 1;
         iconslist.selectedIndex = randomnumber;
+        $("#prop_picture_keyword").trigger("change");
```
nfagerlund: on second thought, let's not re-select.

```suggestion
        $( iconslist ).trigger("change");
```

nfagerlund:
This is ready for review.

nfagerlund:
Hmm, I hadn't tested it on the site scheme entry pages yet, and it DEFINITELY is doing something goofy there. TBH it looks mostly fine just bc the bar is so low, but if you're typing in the text fields in Firefox and you type an apostrophe, it'll trigger the browser shortcut for quick-find (links only). I have literally never seen that happen in a textarea or input before. Some help with that one would be good.

nfagerlund:
NEVER MIND, that's [a known firefox bug that only happens when you're in "responsive design mode."](https://bugzilla.mozilla.org/show_bug.cgi?id=1322385) (It happens on user-styled pages too, it's just that I somehow hadn't typed while that was active yet.) 

So in that case the only problem I noticed with qr in site scheme is that the icon dropdown is too narrow, which should be an easy CSS tweak.

nfagerlund:
Screen Shot 2019-04-15 at Apr 15, 7 43PM

...?!#@$?

lmaoooo no WAY:

Screen Shot 2019-04-15 at Apr 15, 7 43PM 1

nfagerlund:
K. Nice.

rahaeli:
.invisible being position: absoluted to the cornfield is part of how we show things only to people using screenreaders and keyboard-focus, btw. The class could probably stand to be more descriptive, though!

nfagerlund:
@rahaeli Nah, `.invisible` makes perfect sense to me! What I _wasn't_ expecting was for it to get `relative`d by something and then take the entire flex container with it. Extremely daffy_duck_stretch_frame.png.

nfagerlund:
Note for reviewers, there's a visual glitch with paid accounts as this stands because jquery.iconselector.js wraps the icon dropdown with a `<span class='iconselect_trigger_wrapper'>`, usurping the flex item. (I finally made a paid account while I was going in on that responsive-sitescheme-entrypage project.)

I acked the code, and we don't seem to ever again refer to `iconselect_trigger_wrapper` or a span parent of the select in question. Am I missing something, or is it just entirely useless?

kareila:
> I acked the code, and we don't seem to ever again refer to `iconselect_trigger_wrapper` or a span parent of the select in question. Am I missing something, or is it just entirely useless?

I checked the history of that file and I don't think that element was ever targeted by anything. Should be fine to remove it.

momijizukamori:
This moooostly looks good, but I worry we're going to get a bunch of complaints about moving the quote/icon buttons. Which isn't necessarily a reason not to do it! But I'd prefer to get @rahaeli or @zorkian to sign off on it first.

zorkian:
I'm fine with that change. If you rebase this, we can get it merged, if you don't mind. (And if you do I can take a whack at it, LMK.)

zorkian:
I went ahead and did the rebase since I was doing so many this evening!

nfagerlund:
Thanks!! I was actually in the process of that on a WIP branch, but got tied up in some bigger redesign stuff (you saw all that traffic about webfonts). I'll do a quick check for weirdness...

...looks like you handled the main merge conflict in Talk.pm the same way I did, and nothing else goofy happens because I mostly stayed out of perl-land on this one. 👍🏼 !

PR #2416, opened 2019-04-11
Fix compounding blowout on new entry page [#2415]

### LINKED ISSUE #2415: Beta "create entry" page has outrageous compounding blowout on mobile
(`/entry/new`, not `/update`.)

You've probably already experienced this slapstick for yourself, but just in case:

user-blowout

The page starts ultra-hosed, and then every time you change the journal you're posting to it gets another 300+px wider. 

I have fixed this. Patch incoming. It's an ensemble comedy with multiple participants.

nfagerlund:
Resolves #2415.
Fixes one instance of #2421.

Last time, on etc. etc. etc.:

the blowout gif

Should be fixed:

**Mobile width:**

image

**Desktop width:**

image

**An in-between sorta thing:**

image

### DIFF
--- a/htdocs/js/components/jquery.fancy-select.js
+++ b/htdocs/js/components/jquery.fancy-select.js
@@ -39,19 +39,23 @@ function updateSelected(e) {
     $(e.target).next().find("output").html(displayHTML);
 }
 
-function setMinWidth($ele, newMin, padding) {
-    newMin = parseInt(newMin) || 0;
-    padding = parseInt(padding) || 0;
-    var mw = parseInt( $ele.css("min-width") ) || 0;
-
-    $ele.css("min-width", Math.max(mw, newMin) + padding + "px")
+// Don't shrink below the current width the next time the select is changed (but
+// if it blew past the window size at some point, allow it to come back inside).
+function setMinWidth(e) {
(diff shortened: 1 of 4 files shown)


nfagerlund:
Hey, by the way....... do we HAVE to manually babysit these vendor prefixes? Or is there an autoprefixer available somewhere in the SCSS build toolchain?

nfagerlund:
Mmmmm I thought I had this one but some stuff about it is still bedeviling me. Consider this a draft for the next little bit.

nfagerlund:
I think this one is also finally ready for review. Debugging those jquery components is certainly an adventure.

nfagerlund:
After sleeping on it, I think it's better to not bend over backwards to preserve that weird +30px behavior. It looks perfectly good without it, and the component behaves much more predictably.

Once someone gives some feedback on this, I can rebase these commits so they make less of a mess.

momijizukamori:
Yeah, I think we're okay ditching that 30px of padding if it's going to be messy to reimplement it - it's pretty minor aesthetics, and this is still technically in beta. This should be good to merge with a clean rebase!

nfagerlund:
@momijizukamori Sweet! Rebased, and squashed it down to three commits so it reads a little more focused-like.

PR #2418, opened 2019-04-27
[#844, #2417, others] Modernize control strip

### LINKED ISSUE #827: Don't apply page margins to nav strip
It seems some styling applies to the nav strip which can be problematic and make it look very odd and/or cause readability issues (I think someone mentioned font size). One in particular is page margins also applying to the nav strip. This didn't use to be the case. I remember coding margins specifically so they wouldn't do that actually.

### LINKED ISSUE #833: Control strip: take up the entire width of the screen if possible
We want to have the control strip take up the entire width of the screen as much as possible, but then break onto a newline if not

So, brainstorming:

1) display: flex
 I know this will make it able to fill up the remaining space, but I don't know if that will let it fall onto the next line

Note: must keep in mind older browsers don't do flex so need a fallback

2) display: table on larger screens, display: inline-block as we have it on smaller screens

3) ??

### LINKED ISSUE #835: Control Strip: Don't disallow line wrap on smaller screens
Possibly related to #833 

We disallow line wrap on the middle section so that the links don't go onto the next line. This is what was needed when navstrip had to be all on one line, but now that control strip wraps, this might be causing too much wrapping (that is, the middle section, instead of being on line 1 and line 2, takes up its own line so there are lines 1-3)

### LINKED ISSUE #844: Use a more logical HTML structure for the control strip
It no longer needs to be a table, therefore don't make it one.

(it's half-broken appearing with the current table forced to inline hybrid appearance is so let's might as well go all the way and actually make it reasonably easy to style)

### LINKED ISSUE #858: Control Strip login form goes onto two lines
IIRC this used to only take up one line.

(This may very well be fixed by de-tablifying the control strip)

### LINKED ISSUE #1335: can we remove nowrap from navstrip?
The center cell of the navstrip, that contains the "you and so-and-so have mutual subscriptions and access" part, is set to nowrap:

<td id='lj_controlstrip_actionlinks' nowrap='nowrap'>

That was a relic of the days before we responsived the navstrip (yes that is a verb) to keep things from getting funky, I think. Can we remove that nowrap now? It forces the third cell onto a second line at smaller screen widths with longer relationships ("you and verylongusername have mutual subscriptions, and verylongusername grants access to you", etc), and I don't see a problem with making longer relationships spill over onto a second line.

('No, we can't remove that nowrap because reasons' is a valid answer here, I just don't think we rethought the nowrap when we responsived it.)

Brought up in http://www.dreamwidth.org/support/see_request?id=29510.

### LINKED ISSUE #2417: TT-ified /logout controller breaks control strip "log out" button
This can't be experienced in prod yet because the change was never deployed there, but it looks like `cgi-bin/DW/Controller/Auth.pm` does a `... = controller( anonymous => 1, form_auth => 1 )`, and the control strip logout form does not supply the necessary hidden input for form_auth. 

To reproduce: 

- Make a dev server running upstream `develop`, get some users w/ journals, yadda yadda. 
- Go to a user-styled S2 page like a journal or entry.
- Click the logout button in the control strip. 

Expected: my user logs out. Got: form submission error.

### LINKED ISSUE #2420: Goldleaf layout: sideways slop due to footer overflow
If you use the defaults for the footer width, it ends up at 100%, and in content-box sizing that results in sideways slop for the whole page. 

Can be fixed by changing the width to auto (draconian way) or setting `box-sizing: border-box;` (friendly way).

It looks like someone might have previously fixed this issue for the header by setting border-box, but missed the footer, possibly because they weren't using the default width + units anymore.

nfagerlund:
Fixes #844 
Fixes #2420 
Fixes #833 
Fixes #835
Fixes #1335
Fixes #858
Fixes #827

There are several longstanding bugs that require minor, straightforward, and
novice-friendly edits to the markup and styling of the control strip, all of
which have languished because it was entirely impossible to make a minor edit to
the markup of the control strip. 

#2424 extracted the control strip into a template; this PR fixes all of said issues that I could find. Plus a random one that had nothing to do with the control strip but which risked making me briefly look bad if someone ran into it while testing. :]

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -3294,7 +3294,6 @@ sub control_strip {
     my $template_args = {
         'view'          => $view,
         'userpic_html'  => '',
-        'userpic_class' => 'lj_controlstrip_userpic',
         'logo_html'     => ( LJ::Hooks::run_hook( 'control_strip_logo', $remote, $journal ) || '' ),
         'show_login_form' => 0,
         'login_chal'      => '',
@@ -3598,7 +3597,6 @@ sub control_strip {
 
     }
     else {
-        $template_args->{'userpic_class'} = 'lj_controlstrip_loggedout_userpic';
         $template_args->{'userpic_html'} = LJ::Hooks::run_hook('control_strip_loggedout_userpic_contents', $euri) || "";
 
 
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -3598,22 +3598,17 @@ sub control_strip {
 
     }
     else {
+        $template_args->{'userpic_class'} = 'lj_controlstrip_loggedout_userpic';
+        $template_args->{'userpic_html'} = LJ::Hooks::run_hook('control_strip_loggedout_userpic_contents', $euri) || "";
 
-        my $show_login_form = LJ::Hooks::run_hook( "show_control_strip_login_form", $journal );
+
+        my $show_login_form = LJ::Hooks::run_hook("show_control_strip_login_form", $journal);
         $show_login_form = 1 if !defined $show_login_form;
 
         $template_args->{'show_login_form'} = $show_login_form;
 
         if ($show_login_form) {
-            $template_args->{'userpic_html'} =
-                LJ::Hooks::run_hook( 'control_strip_userpic_contents', $euri ) || "&nbsp;";
             $template_args->{'login_chal'} = LJ::challenge_generate(300);
-        }
-        else {
-            $template_args->{'userpic_class'} = 'lj_controlstrip_loggedout_userpic'; # NF: ugh, WHY.
-            $template_args->{'userpic_html'} =
-                LJ::Hooks::run_hook( 'control_strip_loggedout_userpic_contents', $euri )
-                || "&nbsp;";
         }
 
         if ( $journal->is_personal || $journal->is_identity ) {
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -3302,16 +3302,16 @@ sub control_strip {
         'statustext'      => '',
 
         # remote # only set if logged in
-        # .user
-        # .sessid
-        # .display
-        # .is_validated
-        # .is_identity
+        #     .user => "plainname"
+        #     .sessid => integer
+        #     .display => "<span class="ljuser">..."
+        #     .is_validated => bool
+        #     .is_identity => bool
         'actionlinks' => [],
 
         # filters # only set if viewing reading or network page
-        # .all => []
-        # .selected => ""
+        #     .all => []
+        #     .selected => ""
         'viewoptions' => [],
         'search_html' => LJ::Widget::Search->render,
     };
--- a/htdocs/login.bml.text
+++ b/htdocs/login.bml.text
@@ -59,7 +59,7 @@
 
 .login.otheropts=Other Options:
 
-.login.password=Password:
+.login.password=Password
 
 .login.remember=Remember me
 
@@ -69,7 +69,7 @@
 
 .login.title=Log in
 
-.login.username=Account name:
+.login.username=Username
 
 .login.welcome=Welcome to [[sitename]]!
 
--- a/htdocs/stc/controlstrip-dark.css
+++ b/htdocs/stc/controlstrip-dark.css
@@ -2,10 +2,6 @@
 {
     background-color: #2e2c2c;
     background-image: url(/img/controlstrip/bg-dark.gif);
-}
-
-#lj_controlstrip td
-{
     color: #989898;
 }
 
@@ -16,7 +12,6 @@
 
 #lj_controlstrip_user,
 #lj_controlstrip_login,
-#lj_controlstrip_loggedout_userpic,
 #lj_controlstrip_actionlinks
 {
     border-right: 1px solid #878080;
--- a/htdocs/stc/controlstrip-light.css
+++ b/htdocs/stc/controlstrip-light.css
@@ -2,10 +2,6 @@
 {
     background-color: #c7c6c6;
     background-image: url(/img/controlstrip/bg-light.gif);
-}
-
-#lj_controlstrip td
-{
     color: #7c7c7c;
 }
 
@@ -16,7 +12,6 @@
 
 #lj_controlstrip_user,
 #lj_controlstrip_login,
-#lj_controlstrip_loggedout_userpic,
 #lj_controlstrip_actionlinks
 {
     border-right: 1px solid #fbfbfb;
--- a/htdocs/stc/controlstrip.css
+++ b/htdocs/stc/controlstrip.css
@@ -15,27 +15,30 @@ html body
     background-repeat: repeat-x;
     background-position: bottom;
     text-align: left;
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap;
+    justify-content: flex-start;
+    align-items: center;
 }
 
-#lj_controlstrip table
-{
-    margin: 0;
-    padding: 0;
-    width: auto;
-}
-
-#lj_controlstrip td
+#lj_controlstrip > div
 {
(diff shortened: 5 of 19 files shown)


nfagerlund:
Screnshoops:

**The new control strip on mobile:**

<a href="https://user-images.githubusercontent.com/484309/56975768-5d508c80-6b26-11e9-8c79-c100523e9ea5.png">IMG_5414</a> <a href="https://user-images.githubusercontent.com/484309/56975773-5d508c80-6b26-11e9-9c5b-0ae6328b6464.png">IMG_5410</a> <a href="https://user-images.githubusercontent.com/484309/56975775-5de92300-6b26-11e9-9b8d-f05c5e5d04fa.png">IMG_5409</a> <a href="https://user-images.githubusercontent.com/484309/56975776-5de92300-6b26-11e9-815b-aeced35fb0b9.png">IMG_5406</a> <a href="https://user-images.githubusercontent.com/484309/56975777-5de92300-6b26-11e9-8d91-f3d023397654.png">IMG_5407</a> <a href="https://user-images.githubusercontent.com/484309/56975778-5de92300-6b26-11e9-92b9-a3e691b263e2.png">IMG_5408</a>

**Comparison shots of old control strip in torture-test conditions on mobile:**

<a href="https://user-images.githubusercontent.com/484309/56975769-5d508c80-6b26-11e9-9679-9c178e42f34f.png">IMG_5413</a>
<a href="https://user-images.githubusercontent.com/484309/56975771-5d508c80-6b26-11e9-8377-9133dde563de.png">IMG_5411</a>
<a href="https://user-images.githubusercontent.com/484309/56975774-5d508c80-6b26-11e9-8cfc-67aea7dae9d3.png">IMG_5412</a>

**The new control strip on desktop:** (easy enough to find your own comparisons)

<a href="https://user-images.githubusercontent.com/484309/56975785-5f1a5000-6b26-11e9-8d4b-85e904658fb6.png">Screenshot_2019-04-30 betagal Reading</a> <a href="https://user-images.githubusercontent.com/484309/56975782-5e81b980-6b26-11e9-8c70-ee203df2a07b.png">Screenshot_2019-04-30 betagal Recent Entries</a> <a href="https://user-images.githubusercontent.com/484309/56975784-5e81b980-6b26-11e9-8cbb-6044c095b7ed.png">Screenshot_2019-04-30 sheworks4themoney Recent Entries</a> <a href="https://user-images.githubusercontent.com/484309/56975783-5e81b980-6b26-11e9-967d-bfc877fe8e62.png">Screenshot_2019-04-30 sheworks4themoney Recent Entries(1)</a> <a href="https://user-images.githubusercontent.com/484309/56975780-5de92300-6b26-11e9-96ce-2aa0db6602b6.png">Screenshot_2019-04-30 betagal Recent Entries(2)</a> <a href="https://user-images.githubusercontent.com/484309/56975781-5e81b980-6b26-11e9-9ebe-2ba9486fd362.png">Screenshot_2019-04-30 betagal Recent Entries(1)</a> <a href="https://user-images.githubusercontent.com/484309/56975779-5de92300-6b26-11e9-9668-58aa3141ba09.png">Screenshot_2019-04-30 sheworks4themoney Recent Entries(2)</a>

nfagerlund:
Oh, note for reviewers, right now this PR doesn't do any vendor prefixing on the flexbox properties because it's 2019. LMK if we have a policy of it being 2015 instead.

cgi-bin/LJ/Web.pm
```diff
@@ -2738,19 +2738,47 @@ sub control_strip
                     'trusted_by'        => BML::ml('web.controlstrip.status.trustedby', {'user' => $journal_display}),
                     'watched_by'        => BML::ml('web.controlstrip.status.watchedby', {'user' => $journal_display}),
                     );
-    # Style the status text
-    foreach my $key (keys %statustext) {
-        $statustext{$key} = "<span id='lj_controlstrip_statustext'>" . $statustext{$key} . "</span>";
```
kareila: It looks like this `foreach` loop is intended to allow multiple possible statustext lines which are appended to the output in sequence, but in your rewrite, you only assign one line value to `$template_args->{'statustext'}`. While a single line is definitely the most common case, I believe it is possible to for multiple lines to be returned here, and we should continue to support that.

cgi-bin/LJ/Web.pm
```diff
@@ -2738,19 +2738,47 @@ sub control_strip
...
+    my $actionlinks = $template_args->{'actionlinks'};
+    my $viewoptions = $template_args->{'viewoptions'};
+
     if ($remote) {
-        my $remote_display  = LJ::ljuser($remote);
         my $userpic = $remote->userpic;
+        $template_args->{'remote'} = {
+            'sessid' => $remote->{'_session'}->{'sessid'} || 0,
+            'user' => $remote->{'user'},
+            'display' => LJ::ljuser($remote),
```
kareila: The `LJ::ljuser` function is deprecated in favor of the equivalent user method `$remote->ljuser_display`, so we should update that here.

We also discourage treating the user object as a hash, so `$remote->{'user'}` and `$remote->{'_session'}->{'sessid'}` should be updated to `$remote->user` and `$remote->session->{'sessid'}` going forward.  (Or even `$remote->session->id` if you want to make sure the session is defined first -- attempting to call methods on undefined objects is the most frequent problem revealed when doing this conversion.)

cgi-bin/LJ/Web.pm
nfagerlund: I analyzed the logic pretty carefully, and I'm like 98% sure it's intended as one-and-only-one. 

Because of how tangled it is, you pretty much have to look for every instance of `$ret .= $statustext{'<SOMETHING>'};` and keep track of the nested if/elsif blocks until your eyes cross, but I did that several times and I _think_ there are no gaps. Also, the _extremely_ granular statustext options are another argument for that interpretation — if we had multiple chances to add to the status text, we probably wouldn't be making a distinction between `mutualtrust_watch` and `mutualtrust_mutualwatch`. 

I think the foreach loop is just laziness — they might only use one statustext per pageload, but within the logic they end up calling `$ret .= $statustext{'<ETC>'};` like 35 times or something, and wrapping the span around it each time would have been demented. So I just moved it to the template.

Anyway, if y'all decide it's necessary we can keep an array of statustexts (like with actionlinks etc.) and just do a foreach in the template instead. But if my reading of the conditionals is right, this is a chance to clearly express an intent that was impossible to express in the old format.

cgi-bin/LJ/Web.pm
nfagerlund: Oh!! Haha, I totally thought it WAS a hash. Or, I guess I had to switch the `is_verified`/`is_identity` things to method syntax before they'd work, so probably I thought it was something like a javascript object where it has both properties and methods. 

What's the difference between `$remote->session->{'sessid'}` and `$remote->session->id`?

cgi-bin/LJ/Web.pm
kareila: Sure, that sounds fair. We can always adjust it later if needed.

cgi-bin/LJ/Web.pm
kareila: Basically equivalent, just expressed as a hash value versus using an object method.  Objects in Perl are weird, they’re fundamentally “blessed” hashes that you can call methods on.  But accessing a value of an undefined hash complains quietly, while calling a method on an undefined object throws a tantrum that will stop the execution flow.

nfagerlund:
Behold, it looks fine in IE8 now:

Screen Shot 2019-05-24 at May 24, 4 30PM
Screen Shot 2019-05-24 at May 24, 4 32PM 1
Screen Shot 2019-05-24 at May 24, 4 31PM
Screen Shot 2019-05-24 at May 24, 4 32PM

zorkian:
I am not qualified to review CSS so if it looks good then let's ship it. Looks like `cgi-bin/LJ/Web.pm` needs tidying?

zorkian:
Tidied and merged!

nfagerlund:
Thanks!!! 072f3ae8c looks like my eyes were just crossed when I was resolving those merge conflicts, or something, and I picked the untidied instance of those lines.

PR #2423, opened 2019-05-01
Keys on reference is experimental, so we should properly dereference them

rshatch:


kareila:
This fixes the failing compile test on my hack.

PR #2431, opened 2019-05-01
API release hotfixes

rshatch:
-Fix SQL statement for deleting API keys
-Actually return picid as an integer, like we claim we do
-Capture all path parameters, not just the first one
-Loosen JSON validation for incoming data, because Perl doesn't care

### DIFF
--- a/cgi-bin/DW/API/Key.pm
+++ b/cgi-bin/DW/API/Key.pm
@@ -170,7 +170,7 @@ sub delete {
 
     my $dbw = LJ::get_db_writer() or croak "Failed to get database";
     $dbw->do(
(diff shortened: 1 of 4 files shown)


cgi-bin/DW/API/Key.pm
```diff
@@ -170,7 +170,7 @@ sub delete {
 
     my $dbw = LJ::get_db_writer() or croak "Failed to get database";
     $dbw->do(
-        q{UPDATE api_key SET state = 'D' WHERE state = 'A' hash = ?},
+        q{UPDATE api_key SET state = 'D' WHERE state = 'A' AND hash = ?},
```
zorkian: bad reviewer no biscuit, mark

PR #2440, opened 2019-05-04
[#2430] fix error_ml arguments

### LINKED ISSUE #2430: Fix variable substitutions in error messages
In #2278 error_ml was changed to pass only one argument to LJ::Lang::ml, so variable substitutions in error messages aren't happening. For example, while viewing the most recent entry in a journal, attempting to navigate to the next entry (/go?dir=next) displays:

There is no entry following this one. Go back to [[journal]].

If this were the only case where a variable substitution is needed, I think we could update error.tt to pipe the message thru ml and do the substitution, but it isn't, there are a bunch of different errors that need different inputs to LJ::Lang::ml, so I think this still needs to be done in error_ml the way it was before.

kareila:
The previous change to error_ml dropped any message arguments for LJ::Lang::ml
on the floor, because the expected inputs weren't well documented.

This changes error_ml to accept an input format similar to success_ml, where
the message arguments are in the first hashref, and any page variables are
in a subsequent hashref.

The two places where page variables had begun being passed into error_ml are
updated to pass in undef for the message arguments, so everything should line
up correctly.

Fixes #2430.

### DIFF
--- a/cgi-bin/DW/Controller.pm
+++ b/cgi-bin/DW/Controller.pm
@@ -42,9 +42,8 @@ sub needlogin {
 
 # returns an error page using a language string
 sub error_ml {
-    my ($string, $opts) = @_;
     return DW::Template->render_template(
-        'error.tt', { message => LJ::Lang::ml( $string ), opts => $opts }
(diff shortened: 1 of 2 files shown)


zorkian:
LGTM, thank you!

PR #2442, opened 2019-05-06
[#2434] Don't insert jQuery script tags into main body content

### LINKED ISSUE #2434: Don't insert script tags that invoke jquery at random points in the page content
I ran into this because it was a blocker on #2427, but it also seems worth doing just for the sake of consistency and hygiene. 

Conventional practice says script tags go either in the head or at the very bottom of the body. DW's `need_res` deferred javascript loader does exactly that. By which I mean it does both thats: if the active resource group (bear with me here) is `"jquery"`, they go in the head, but if it's `"foundation"` then they go at the bottom of the body.

But DW also has a few spots where script tags just get squirted at random into the body text instead of deferred, because the code is older than heck and that's just what happened sometimes. These are mostly in S2 pages, I think, and I don't believe there's very many of them left. 

In most cases, that's "annoying" or "gross" instead of "bad." But if you're interested in converting the S2 site scheme pages to use the Foundation version of site scheme (#2427), then you've got a problem, because some of those random script tags try to call jQuery, and in Foundation land jQuery isn't loaded yet. It's just down there at the end of the body element, a chasm of content between it and the caller, watching helplessly as the footgun goes off in slow motion. 

The fix is pretty tractable tho. As far as I could tell, it's only the qrdiv-adder, the quotebutton, and the iconbrowser; the other rogue script tags use ancient vanilla JS and can wait their turn.

I've got a branch, but it depends on that quickreply PR, so I'm gonna wait until that's merged.

nfagerlund:
Fixes #2434. 

Currently, several parts of the reply forms will only work if all deferred javascript resources are loaded during the page's `<head>`. This PR decouples those features from the position of the global script tags, so they'll continue to work if the deferred JS gets moved to the bottom of the `<body>`. These changes should have zero effect on existing behavior, but will enable some future feature work.

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -2176,7 +2176,7 @@ sub talkform {
         my $hidebutton = ( $opts->{errors} && @{ $opts->{errors} } );
         unless ( $hidebutton ) {
             my $alerttext = LJ::Lang::ml( 'talk.error.quickquote' );
-            $ret .= "<span id='quotebuttonspan' data-quote-error=\"$alerttext\"></span>";
+            $ret .= qq{<span id="quotebuttonspan" data-quote-error="$alerttext"></span>};
         }
     }
 
@@ -2343,7 +2343,7 @@ sub icon_dropdown {
         if ( $remote && $remote->can_use_userpic_select ) {
             my $metatext = $remote->iconbrowser_metatext ? "true" : "false";
             my $smallicons = $remote->iconbrowser_smallicons ? "true" : "false";
-            $ret .= '<input type="button" id="lj_userpicselect" value="Browse" data-iconbrowser-metatext="' . $metatext . '" data-iconbrowser-smallicons="' . $smallicons . '"/>';
+            $ret .= qq{<input type="button" id="lj_userpicselect" value="Browse" data-iconbrowser-metatext="$metatext" data-iconbrowser-smallicons="$smallicons"/>};
         }
 
         # random icon button - hidden for non-JS
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -2174,11 +2174,9 @@ sub talkform {
 
         # only show quick quote button on initial composition
         my $hidebutton = ( $opts->{errors} && @{ $opts->{errors} } );
-        unless ($hidebutton) {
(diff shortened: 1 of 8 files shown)


htdocs/js/jquery.iconselector.js
```diff
@@ -311,4 +311,13 @@
         }
     };
 
+    var browseButton = $("#lj_userpicselect");
```
nfagerlund: This worked in my foundation branch, but in retrospect it's the only bit that I'm not 100% sure isn't broken when the deferred script tags are in the head. It should probably be outside this main function body, in a separate `jQuery( function() {...} );` callback, to make sure it's only called once the DOM is ready.

cgi-bin/LJ/Talk.pm
```diff
@@ -2153,8 +2153,9 @@ sub icon_dropdown {
 
         # userpic browse button
         if ( $remote && $remote->can_use_userpic_select ) {
-            $ret .= '<input type="button" id="lj_userpicselect" value="Browse" />';
-            $ret .= LJ::Talk::js_iconbrowser_button();
+            my $metatext = $remote->iconbrowser_metatext ? "true" : "false";
+            my $smallicons = $remote->iconbrowser_smallicons ? "true" : "false";
+            $ret .= '<input type="button" id="lj_userpicselect" value="Browse" data-iconbrowser-metatext="' . $metatext . '" data-iconbrowser-smallicons="' . $smallicons . '"/>';
```
zorkian: nit: Use `qq{...}` to avoid having to concatenate quoted strings with mixed quotes and variables. Something like this:

```perl
$ret .= qq{<input type="button" id="lj_userpicselect" value="Browse" data-iconbrowser-metatext="$metatext" data-iconbrowser-smallicons="$smallicons"/>};
```

zorkian:
I looked at the last three commits, they LGTM. Whenever you've rebased the other and it's landed, happy to merge this.

zorkian:
(Marking requested changes just so I know I've looked at it.)

cgi-bin/LJ/Talk.pm
nfagerlund: good call, thanks!

nfagerlund:
@zorkian This is rebased/tidied and ready for merge!

zorkian:
LGTM, merged up!

PR #2448, opened 2019-05-11
[#2254] Remove some unused methods as listed

### LINKED ISSUE #2254: Consider removing LJ::User functions and methods unused since at least August 2009
The following LJ::User functions and methods, which were marked as unused in August 2009 (see https://github.com/dreamwidth/dw-free/commit/a397a0cc33c090c4531b8bb85f1045aa3697158c) are still unused after 8 years and change and should perhaps be removed:

- note_transition (now in cgi-bin/LJ/User/Data.pm)
- last_transition (now in cgi-bin/LJ/User/Data.pm) only used in note_transition which also goes away
- transition_list (now in cgi-bin/LJ/User/Data.pm) only used in last_transition which also goes away
- format_time (now in cgi-bin/LJ/User/Display.pm)

alexwlchan:
Closes #2254.

I grepped the codebase to double-check these functions weren’t in use before deleting them.


zorkian:
Thank you for helping clean up the codebase!

PR #2451, opened 2019-05-11
[#2275] Remove an unused variable in Protocol.pm

### LINKED ISSUE #2275: Useless variable in LJ::Protocol lines 1494&1496
This is probably a remnant of old code that used to limit how many moderators are emailed when a new moderated community entry is queued. Since that variable is never incremented, testing it and declaring it are pointless and should be removed. Found while working on bug #2052.

alexwlchan:
Per the issue notes, there's no limit on community moderators anywhere else and this `last` statement will never execute, so just remove it.

Fixes #2275.


zorkian:
Merged, thank you!

PR #2452, opened 2019-05-11
[#1852] Include both absolute and relative timestamps in the inbox

### LINKED ISSUE #1852: add absolute time to inbox timestamps
Suggestion from Feedback (modified slightly): there are cases, especially with high-volume discussion/writing-meme communities, where being able to tell the absolute time a comment was left, at a glance, from notifications in the inbox is helpful. Relative time ("X hours ago") is also useful for other use cases, so let's do both. Where the inbox timestamp is currently:

> 2 hours ago

It should be:

> 2016-08-09 3:41 AM EST 
> 2 hours ago

Timestamp should be in user's timezone if they've set it in preferences, otherwise UTC.

alexwlchan:
Closes #1852. This displays both the absolute time and the relative time in inbox notifications.

Screenshot:

Screenshot 2019-05-11 at 21 44 31

Or if you have a timezone set:

Screenshot 2019-05-11 at 21 24 33

I’ve checked, and this respects your timezone and 12h/24h time format preferences. Huzzah for reusable code!

I could only find one use of the `sitescheme_secs_to_iso` routine elsewhere – see https://github.com/dreamwidth/dw-free/blob/2fb06b1291a99409c61e91f701f8cbbe6e8645f1/cgi-bin/LJ/Talk.pm#L3801 – and it doesn't pass any extra options, so I think changing the hash/hash reference here is fine.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2638,7 +2638,7 @@ sub nth_entry_seen {
 }
 
 sub sitescheme_secs_to_iso {
-    my ( $secs, %opts ) = @_;
+    my ( $secs, $opts ) = @_;
     my $remote = LJ::get_remote();
     my @ret;
 
@@ -2653,7 +2653,7 @@ sub sitescheme_secs_to_iso {
     my $has_tz = '';  # don't display timezone unless requested below
 
     # if opts has a true tz key, get the remote user's timezone if possible
-    if ( $opts{tz} ) {
+    if ( $opts->{tz} ) {
         $s2_datetime = DateTime_tz( $secs, $remote );
         $has_tz = defined $s2_datetime ? "(local)" : "UTC";
     }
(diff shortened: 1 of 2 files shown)


alexwlchan:
While working on this, I had a quick poke at whether you could do this elsewhere, and I think the answer is “probably not without lots more work”.

Comments are already displayed this way – see screenshot below. This patch means the timestamp in notifications and on the comment itself are consistent.

Screenshot 2019-05-11 at 21 55 28

But in the post title, every part of the date is a link to a filter page:

Screenshot 2019-05-11 at 21 56 20

Here “May”, “11” and “2019” are all links to date pages. Those dates will shift depending on your timezone, and you’d have to build those pages relative to the logged-in user’s timezone, or you could click from a post dated “May 10” and not find it on the “posts from May 10” page. Which is confusing!

rahaeli:
Yes, post times are user-supplied and not to be treated as anything other than "when the person wanted you to think the entry was posted" (you can backdate or forward-date an entry to any time you want, to say nothing of timezones). Comments are server-timestamped but entries have time provided by the user and could be anytime from 1970 on.

PR #2454, opened 2019-05-11
Upgrade to reCAPTCHA v2 API. Fixes #2331

### LINKED ISSUE #2331: Upgrade to recaptcha 2
First gen API being discontinued, etc

rshatch:
BIG IMPORTANT NOTE: the API keys may have to be regenerated for v2! I do not have existing v1 keys so I couldn't test if they're forwards compatible, but I had to specify which version I wanted keys for when I generated the v2 ones.


zorkian:
Alright, let's add this to the shopping list for next code push!

Issue #1568, opened 2015-09-06
username maxlength not consistent everywhere, should probably be named constant


kareila:
So as most of you probably know, when DW forked from LJ, LJ had a maximum username length of 15 characters, and we upped that to 25 characters.  Unfortunately, we didn't change it uniformly everywhere, and there are still a few cobweb-filled corners of the code that limit usernames to 15 characters.  I just found one in LJ::Session, and it looks like htdocs/tools/tellafriend.bml and htdocs/tools/textmessage.bml also have the problem; there may be others.

At a minimum, these should be increased, but ideally I'd also like to find all the places that have hardcoded 25 and have them reference a site constant, $LJ::USERNAME_LENGTH.  I realize that increasing it again would also require columns to be widened in the database, but that would still be two places to change as opposed to dozens.

zorkian:
Site constant is a good idea, even given the caveat. :)

--
Mark Smith
mark@qq.is
 
 
On Sun, Sep 6, 2015, at 11:32 AM, Jen wrote:
> So as most of you probably know, when DW forked from LJ, LJ had a maximum username length of 15 characters, and we upped that to 25 characters.  Unfortunately, we didn't change it uniformly everywhere, and there are still a few cobweb-filled corners of the code that limit usernames to 15 characters.  I just found one in LJ::Session, and it looks like htdocs/tools/tellafriend.bml and htdocs/tools/textmessage.bml also have the problem; there may be others.


> At a minimum, these should be increased, but ideally I'd also like to find all the places that have hardcoded 25 and have them reference a site constant, $LJ::USERNAME_LENGTH.  I realize that increasing it again would also require columns to be widened in the database, but that would still be two places to change as opposed to dozens.


> —
> Reply to this email directly or view it on GitHub[1]. 


>  
 


Links:

  1. https://github.com/dreamwidth/dw-free/issues/1568

swaldman3:
claim

swaldman3:
Unclaim / unassign.

Anybody looking to pick this up - note that it is nearly done, at least for some cases, in swaldman3/dw-free. There were some further style changes requested. See PR 1610.

alexwlchan:
I’m going to poke at #1610, try rebasing against develop and see how far I can get through the PR comments.

PR #2456, opened 2019-05-11
[#1568] Use a named constant for username maxlength

alexwlchan:
A partial fix for #1568. This just takes @swaldman3’s PR #1610, rebases it against the current tip of develop, and addresses the outstanding comments.

As noted in that PR, there are probably other cases that have been missed.

e.g. https://github.com/dreamwidth/dw-free/blob/4ca92d214e6a92e94ad19f3058633fa4e4201f2c/views/directory/index.tt#L92-L97 looks plausibly in need of using-a-constant

But going through and evaluating all of those takes time, and it’s nearly my bedtime here. 😴 

(I count ~16.5k occurrences of the strings “15” or “25” somewhere in the codebase. Heuristics will cut that down somewhat, but some of them need inspecting manually.)

I figured better to get _some_ of this code into the main branch, and other occurrences can be added as and when, than try to get something perfect.

### DIFF
--- a/cgi-bin/LJ/User/Account.pm
+++ b/cgi-bin/LJ/User/Account.pm
@@ -1283,7 +1283,7 @@ sub canonical_username {
     my $input = lc( $_[0] // '' );
     my $user = "";
     my $exp = '^\s*([a-z0-9_\-]{1,' . $LJ::USERNAME_MAXLENGTH . '})\s*$';
-    if ( $input =~ $exp ) {  # good username
+    if ( $input =~ /$exp/ ) {  # good username
         $user = $1;
         $user =~ s/-/_/g;
     }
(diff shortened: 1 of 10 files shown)


cgi-bin/LJ/User/Account.pm
```diff
@@ -1282,7 +1282,8 @@ use Carp;
 sub canonical_username {
     my $input = lc( $_[0] // '' );
     my $user = "";
-    if ( $input =~ /^\s*([a-z0-9_\-]{1,25})\s*$/ ) {  # good username
+    my $exp = '^\s*([a-z0-9_\-]{1,' . $LJ::USERNAME_MAXLENGTH . '})\s*$';
```
zorkian: This still confuses me, not sure what this line does :)

I'll delete it to get this merged..

zorkian:
Merged, and thank you!!

PR #2469, opened 2019-05-31
[#2468] Make quote button work better on mobile

### LINKED ISSUE #2468: Quote button in reply form doesn't work on iOS
I am shocked, _shocked_ that this ancient javascript wasn't written with mobile in mind. Anyway, looks like it might be grabbing focus away from the text selection before it has a chance to get the selection.

nfagerlund:
Fixes #2468.

The comment form quote button uses javascript to grab the current text selection
at the time the button was pressed. But touch-based mobile browsers smear the
timing of "pressed," such that the selection has usually been cleared by the
time the button's click handler fires. The user experience of that is basically
:trollface:. It sometimes works! Technically!

You can reliably fix this by retaining the last non-zero text selection and
quoting _that_ if the current selection is zero.

On mouse platforms, where text selection is an easy fidget behavior instead of a
slow and intentional command, there's a medium chance of doing the wrong thing
if someone fidget-selects for a while and then uses the quote button for the
first time. Luckily, the last two years of phone browsers all support CSS4 media
interaction features, so it's easy to only target platforms with slow-select.

Since every browser from the last seven years supports the W3C draft of the
selection API (including IE9!), this also seems like a good time to stop it with
that IE8 thing, especially since Microsoft themselves deleted their docs for it
and redirected the page to MDN's Selection docs. (#wow)

### DIFF
--- a/htdocs/js/jquery.quotebutton.js
+++ b/htdocs/js/jquery.quotebutton.js
@@ -1,45 +1,52 @@
 // On pages with a reply form, create a quote button that pastes in selected
-// page text wrapped in a quote element. #body is what the quickreply form uses,
-// and #commenttext is like some old talkpost_do shenanigans that will take some
-// extra effort to dispose of.
+// page text wrapped in a quote element.
 jQuery(function($){
-    var helped = 0; var pasted = 0;
+    var showHelp = true;
+    var lastSelection = '';
+    var quoteTarget = $('textarea#body'); // quickreply
+    if (quoteTarget.length === 0) {
+        quoteTarget = $('textarea#commenttext'); // talkform
+        if (quoteTarget.length === 0) {
+            return; // Nowhere to paste, skip all other quote button setup.
+        }
+    }
+    quoteTarget = quoteTarget.get(0);
+
+    // Touch-based browsers like to collapse the selection too early, so
+    // retain the last real selection.
+    if (window.matchMedia("(any-hover: none)").matches) {
+        document.addEventListener('selectionchange', function() {
+            newSelection = document.getSelection().toString();
+            if (newSelection.length > 0) {
+                lastSelection = newSelection;
+            }
+        });
+    }
+
+    // Return current selection or last intentional selection
+    function getSelection() {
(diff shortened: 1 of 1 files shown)


nfagerlund:
I haven't tested this on a dev instance yet, but http://www.sedumphotos.net/nfagerlund/quoter2.html is a demo that uses this same JS.

nfagerlund:
References: 

- https://caniuse.com/#feat=selection-api
- https://caniuse.com/#search=media%20features

zorkian:
I played around with the demo, it even quotes if you select some text in the textarea. Is that intentional?

nfagerlund:
@zorkian It is not intentional! Huh!

Well, that seemed like a surprising thing for this code to introduce, so I went and checked the desktop browsers against the real-life quote button on prod:

- Chrome does the same thing on dreamwidth.org. 
- Firefox and Safari don't.

So, while I think that's kind of a misbehavior, it's at least not making anything _worse_ — it takes Mobile Safari from "hella broken" to "same as desktop Chrome," and I don't think anyone's complained about the quote button on Chrome, yeah? It's also not really confusing — it's obvious what happened, and at least my reaction would be "lol whoops, cmd-Z".

In the mobile case, I can see how to prevent it — since we're updating the text in the background, we can just compare `document.activeElement` to our `quoteTarget` and skip the update if the textarea is focused.

In the case of desktop Chrome, that wouldn't work because we get the text in response to a click on the quote button, which blurs the textarea. The only other workaround I can think of is to compare the current selection to `quoteTarget.value.substring( quoteTarget.selectionStart, quoteTarget.selectionEnd )` and bail if they're identical, but the problem with that is that textareas retain their personal selection properties when they blur, so if you selected text in the field and then _just happened to select the exact same text in the page,_ it would give a false failure, even though your intention was clear.

So for Chrome, I don't think a workaround is worth the complexity. Should I go ahead and check the activeElement on mobile?

nfagerlund:
(also, have now tested this on dev and it works as expected.)

zorkian:
SGTM, thanks for confirming!

PR #2473, opened 2019-06-04
Talkpost extraction: Easier-to-review version

nfagerlund:
**This is not going to be the mergeable version of the PR.**  This is only to help reviewers view the extraction of the logic as bite-sized chunks. 

I'm about to rebase the real branch onto the upstream and PR it for real, but I think there are going to be some merge conflicts. Since the code involved is so bad, it'll probably be very hard to fix those conflicts incrementally while preserving the order of the part-by-part extraction, and if I run into any roadblocks I'm just going to play it safe and squash all. But that'll be pretty hell to review, so I'm helping you out here. :) 

Targeting code-tour for a more legible diff, since that's about where this branch diverged from develop.

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1548,6 +1548,8 @@ sub talkform {
         # Partial form was submitted; pick up where we left off.
         $default_usertype = $form->{usertype};
     } elsif ($remote) {
+        # If logged-in user isn't allowed to comment we end up with no default,
+        # which seems correct.
         if ($remote->openid_identity) {
             $default_usertype = 'openid_cookie';
         } else {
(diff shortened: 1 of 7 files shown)


cgi-bin/LJ/Web.pm
```diff
@@ -777,6 +777,54 @@ sub bad_password_redirect {
...
+#          If no userpics or if remote is undefined, an empty array.
+# </LJFUNC>
+sub icons_for_remote {
+    my ($remote) = @_;
+    my @pics;
+
+    {
+        my %res;
+        if ($remote) {
+            LJ::do_request(
```
zorkian: Just a note for future spelunkers... this is such a weird design to call the protocol handler and "log in" to get the userpics.

nfagerlund: man, I was like, I don't even want to figure out what's happening in here, just isolate the infection and g e t  o u t

cgi-bin/LJ/Talk.pm
```diff
@@ -1516,11 +1516,20 @@ sub talkform {
         'form_url' => LJ::create_url( '/talkpost_do', host => $LJ::DOMAIN_WEB ),
         'errors' => $opts->{errors},
         'create_link' => '',
+
         'editid' => $editid,
         'editreason' => $comment ? $comment->edit_reason : '',
+        'comment_body' => $form->{body} || '',
```
zorkian: Do you really mean `||` (truthy-or) or would `//` (defined-or) be better here?

cgi-bin/LJ/Talk.pm
```diff
@@ -1540,6 +1541,25 @@ sub talkform {
         $basesubject = "Re: $basesubject";
     }
 
+    my $screening = LJ::Talk::screening_level( $journalu, $opts->{ditemid} >> 8 ) || '';
+    $screening = 'A' if $journalu->has_autoscreen($remote);
```
zorkian: nit: This check should be first since it always overwrites the value of whatever `LJ::Talk::screening_level` does. Although now that I look at it, it definitely predates this diff... oh well.

cgi-bin/LJ/Talk.pm
```diff
@@ -1603,7 +1606,14 @@ sub talkform {
             ljuser => $remote->ljuser_display,
             openid_identity => $remote->openid_identity,
 
-            # allowed => ,
+            allowed => !$journalu->has_banned($remote) && (
```
zorkian: Feels like a further cleanup would be to pull a lot of this logic out to the user object, instead of having these many-fold conditionals in the middle of the talkform rendering. Then we could add tests around them and such...

zorkian:
I made a few comments (most of which we don't need to address here, but can be thought of as future improvements if someone desires), but nothing major. I tried to follow as much of the critical path logic as I could and didn't see anything that felt worrisome or wrong, the diffs being broken down like this really helped -- thank you!

cgi-bin/LJ/Talk.pm
nfagerlund: Now that I look closer, I think I don't actually want to bother with any "or," because TT will just output an empty string if I render an undefined value, which is what I want. (I dropped that unnecessary logic for a couple of later pieces, but not this one.)

(P. sure I just copied this from the existing logic, which I think predated `//` being reliably available.)

cgi-bin/LJ/Talk.pm
nfagerlund: Oh huh, and actually I think that's not necessary anymore because I'm duplicating that check down in the `remote.screened` template argument. (In the commit-by-commit version, I couldn't have removed that until all the from fields were dealt with.)

cgi-bin/LJ/Talk.pm
nfagerlund: Huh. Should this all be rolled into the `does_not_allow_comments_from` method, or a new one?

zorkian: I would invert it probably, I always find it easier to reason about positive sounding conditions with if statements. Since most of the time we'd be writing something like,

```perl
return $bad_request->( 403 ) unless $u->allow_comments_from( $remote );
```

etc. FWIW, I don't think this is necessary for this patch, I'm just commenting for future work. I don't want to block merging the work already done!

zorkian: also I said if statements and then used an unless, go me

zorkian:
I'm closing this since the main one is merged, but feel free to reopen if it's necessary/useful!

PR #2477, opened 2019-06-04
Let's see if we can figure out wtf tidyall thinks it's doing

nfagerlund:
I'm filing this as a for-real PR because that's the only way to get Travis to run it, but it's not intended for actual merge.


nfagerlund:
I think I've figured it out, so https://github.com/dreamwidth/dw-free/pull/2478

PR #2478, opened 2019-06-04
Resolve mysterious tidyall Travis failures (hopefully!)

nfagerlund:
Perltidy's vertical alignment behavior changed drastically somewhere between 20181120 and 20190601, probably somewhere around commit https://github.com/perltidy/perltidy/commit/e15a1df. Travis is using the newest version it can get, I think, and it looks like we've all been running something older.

### DIFF
--- a/bin/ljsysban.pl
+++ b/bin/ljsysban.pl
@@ -174,7 +174,7 @@ if ($modify) {
         || ( $status && $status ne $ban->{'status'} && $status eq 'active' ) )
     {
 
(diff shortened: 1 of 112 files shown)


nfagerlund:
Well this totally worked, but Travis still fails because of the compile errors that are addressed in https://github.com/dreamwidth/dw-free/pull/2475 and https://github.com/dreamwidth/dw-free/pull/2474 😅

Issue #2437, opened 2019-05-02
Extract talkform into template


nfagerlund:
This monster: 

https://github.com/dreamwidth/dw-free/blob/2fb06b1291a99409c61e91f701f8cbbe6e8645f1/cgi-bin/LJ/Talk.pm#L1425

It offends me and should therefore die.

nfagerlund:
WIP branch, on pause for probably the next week: 

https://github.com/nfagerlund/dw-free/tree/may19_talkform

nfagerlund:
claiming

(oops, forgot to do that earlier. Not that anyone else is in a hurry to touch this garbage!)

nfagerlund:
WHOOPS, forgot to include the "fixed" line in #2479. Closing.

PR #2479, opened 2019-06-05
[#2437] Extract talkform into template

nfagerlund:
This is the real, mergeable version of #2473. 

Anyone actually willing to review this probably already saw this note, but #2473 has most of this changeset broken out into more digestable commits — the "files changed" view is just as bad, but the individual commit view shows the row-by-row progress of butchering this table, and IMO that's an easier way to understand just what went down over here. 

The commits after the big one are changes that came after the reviewable draft. So far, it's a bug I spotted and an unnecessary complexity that @zorkian removed while I was elbow-deep in it.

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1589,7 +1589,7 @@ sub talkform {
         form_url             => LJ::create_url( '/talkpost_do', host => $LJ::DOMAIN_WEB ),
         errors               => $opts->{errors},
         create_link          => '',
-        subject_icons        => $subjecticons,
+        subjecticons        => $subjecticons,
 
         public_entry     => $entry->security eq 'public',
         default_usertype => $default_usertype,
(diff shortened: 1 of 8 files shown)


cgi-bin/LJ/Talk.pm
```diff
@@ -1505,11 +1546,147 @@ sub talkform {
...
+            editid      => $editid,
+            editreason  => $comment ? $comment->edit_reason : '',
+            oidurl      => $form->{oidurl},
+            oiddo_login => $form->{oiddo_login},
+            user        => $form->{userpost}
+                || $form->{cookieuser},
+            body        => $form->{body},
+            subject     => $basesubject,
+            subjecticon => $form->{subjecticon}
+                || 'none',    # a subjecticon ID
```
nfagerlund: This will get sent in the form submission. The reason that's okay is because `LJ::Talk::init()` special-cases both `'none'` and `''` as a blank subjecticon field.

zorkian:
The review on the other split-out version did not generate any blocking items, so I'm going to merge this monstrosity (and I mean that in the most happy, encouraging way).

PR #2486, opened 2019-06-07
Styles: Modernize image constraints

nfagerlund:
Within the content of an entry or comment:

- Large images should never overflow their container boundaries, EVER. 
- Portrait-orientation images should be viewable in toto on a single screen without scrolling. 
- Images should never get their aspect ratios messed up. 
- If a user used width and/or height attributes in an `img` tag to make it smaller or larger than the native pixel size, we should respect that (_as long as it's within the container boundaries_). 

Well, CSS lets you do all that now. [Demo page](http://www.sedumphotos.net/nfagerlund/imgtest.html) — try zooming up or down, and try opening the web inspector and messing with the div width and/or the img height/width attributes.

### DIFF
--- a/styles/abstractia/layout.s2
+++ b/styles/abstractia/layout.s2
@@ -1227,14 +1227,20 @@ q {
 }
 .comment-content { border-top: 1px transparent solid; } /* for firefox */
 
-/* To constrain the width and prevent layout breaking */
+/* Constrain image dimensions.
+    Job 1: Don't trash the layout sideways.
+    Job 2: Limit height to fit inside the viewport. Having to scroll to see a
+      portrait of someone is nonsense.
+    Job 3: Defend the native aspect ratio.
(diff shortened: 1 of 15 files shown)


nfagerlund:
Question for @rshatch or anyone else who's dealt with the weird way these Rasa-scions are somehow layout-children-of-a-layout: Can we just, like, delete the reiteration of these limits on those, and let them inherit from Rasa? Or are they doing something additionally wack.

rshatch:
If they're the same across all the children (which they probably are). Basically all the TR children have two stylesheets instead of one, with one of those being the TR stylesheet.

nfagerlund:
OK, sweet. I went back through that second commit and made sure I wasn't adding that CSS to any Rasa-babies. (There were a few that were adding more img shenanigans, which I just removed so's to not interfere with the TR CSS.) 

This should be ready to go.

zorkian:
TIL about `vh`. That's pretty cool.

PR #2489, opened 2019-06-11
Styles: Wrong keyword value in image fix

nfagerlund:
I used a Firefox-ism that isn't broadly supported! When an image with height and
width attributes gets shrunk to fit, this can cause some white space above and
below it in other browsers. `height: auto;` is equivalent to what I meant here.
(It overrides the height attribute, but sets it to match the native aspect ratio
for the width we ended up using. 99.5% of the time this is correct; the other
.5%, you can override it with an inline `min-height`.)

### DIFF
--- a/styles/abstractia/layout.s2
+++ b/styles/abstractia/layout.s2
@@ -1236,7 +1236,7 @@ q {
       (within the limits of the container), but if they conflict with the aspect
       ratio, treat them as maximums and let the aspect ratio win. */
 .entry-content img, .comment-content img {
-    height: max-content;
+    height: auto;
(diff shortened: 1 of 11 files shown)


zorkian:
Out of curiosity, is there anything like a CSS-browser-linter where you can create a file and say 'we must support browsers X>40, Y>3.0, Z>27' and it validates that the CSS conforms to those requirements?

rahaeli:
A bit of searching seems to indicate https://caniuse.com has browser based (http://doiuse.herokuapp.com) and command line (https://github.com/sgentle/caniuse-cmd) tools!

PR #2490, opened 2019-06-12
Fix thick-bottomed userpic boxes

nfagerlund:
A normal `<a>` is an inline element whose height is derived from font metrics and
line-height, rather than from the height of the box-things in it. The `<a>` around
a userpic thus adds some extra height to the surrounding div, making styling
difficult and often adding visual glitches. If you make it a block element
and remove the line-height, it will fit snugly to its `` contents instead.

Before:

Screen Shot 2019-06-12 at 10 50 46 AM

After:

Screen Shot 2019-06-12 at 10 51 11 AM

Before:

Screen Shot 2019-06-12 at 11 11 28 AM

After:

Screen Shot 2019-06-12 at 11 11 00 AM

### DIFF
--- a/htdocs/stc/jquery.contextualhover.css
+++ b/htdocs/stc/jquery.contextualhover.css
@@ -32,6 +32,12 @@
     padding: 2px;
 }
 
+/* Don't add extra height at bottom of userpic box */
+.ContextualPopup .Userpic a {
+    display: block;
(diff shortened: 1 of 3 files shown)


zorkian:
And thanks for the earworm. ;)

nfagerlund:
😂

PR #2491, opened 2019-06-12
Change several JS setup handlers to DOMContentLoaded

nfagerlund:
The `DOMContentLoaded` event on `document` is the correct target for most setup
stuff, but a lot of ancient code uses the `load` event on `window` instead,
because support for `DOMContentLoaded` was sparse.

Nowadays everything supports `DOMContentLoaded`.

This will break support for some features in IE8.

This MIGHT fix the issue where the inbox buttons refuse to respond to anything until a massive fanfiction.net image timeout finishes happening. 

References:

- https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event
- https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event


kaberett:
For my own future reference, this was discussed and characterised over in https://dw-support-training.dreamwidth.org/33538.html

PR #2494, opened 2019-06-21
Add encoding checks to comment importer

rshatch:


alierak:
Are you sure this is the right / only section that would need modifying to handle other encodings? It looks to me like parser->parse appears several other places, including one that tries to special-case ISO-8859-1, and the place you've changed probably has to do with parsing the comment "meta" structure rather than the comment bodies.

rshatch:
....I think you're right, looking at this, though I am too fried tonight to sort it out.

zorkian:
This fails tidy, please tidy up and I'll merge!

zorkian:
@rshatch any chance for that tidy? :D

zorkian:
Tidied and merged.

zorkian:
https://github.com/dreamwidth/dw-free/compare/5b84b6b4f0e0...43d44066d2b0

PR #2498, opened 2019-06-22
pre-code push fixes

kareila:
Found these running the test suite for the first time since the last push.


nfagerlund:
TIL about Defaults.pm! Thank you for catching that.

PR #2516, opened 2019-06-24
fix image overflow CSS issue with poll code

kareila:
There are two commits here. The first one puts the HTML string in one place instead of two, and the second one adds the inline CSS that @nfagerlund thinks will fix the problem displaying poll results on styles with his image overflow fix. I haven't tested this myself.


nfagerlund:
The CSS is 👍🏼 . 

What I'm not sure about is what'll happen when the bar's requested width is wider than the container. I think they'll end up constrained, which'll make the graph inaccurate. But the alternative is allowing a nasty sideways overflow. (Really, shouldn't these bars be using percentages to fit within their container? There's probably something that makes that hard that I haven't thought of yet.)

zorkian:
Heh concatencruft, good one...

PR #2517, opened 2019-06-24
Quickreply Re-re-re-design: Brutalist Ziggurat Edition 

nfagerlund:
In this re-adjustment, I'm giving up entirely on trying to make two rows of
controls align cleanly with the icon preview, because:

1. In the best case scenario, it was looking better but working worse (due to
   adding unnecessary length to the icon selector dropdown, etc.),
2. In the more common scenario, it was also looking worse, due to stretching the
   buttons in a way that didn't make sense when compared to the buttons below,
3. TBH it wasn't even working a lot of the time, because of how different
   platforms style their form controls! Rounded corners or weird user-agent
   !important rules about height and font size were making things not line up
   squarely anyway, like 40% of the time!!!

Farewell you beautiful dream (of perfectly straightened picture frames). In its
place, we get a brutalist-in-the-good-sense rethink of the icon controls:

- Consistent and platform-appropriate button styling.
- Ragged-right margin.
- The dropdown only sticks out as far as it needs to, and doesn't get weird
  vertical stretch.
- A cool triangular ziggurat kind of shape. 📐🏕🏔

In summary: Looks worse, works better.

Also included:

- Bring back the click-on-icon browse button (as a bonus, not the sole instance). 
- Switch position of random and browse buttons (by popular demand).

### DIFF
--- a/htdocs/js/jquery.iconselector.js
+++ b/htdocs/js/jquery.iconselector.js
@@ -319,7 +319,7 @@ jQuery(function($) {
     var browseButton = $("#lj_userpicselect");
     if (browseButton.length > 0) {
         $("#prop_picture_keyword").iconselector({
-            selectorButtons: "#lj_userpicselect",
+            selectorButtons: "#lj_userpicselect, .lj_userpicselect_extra",
             metatext: browseButton.data("iconbrowserMetatext"),
             smallicons: browseButton.data("iconbrowserSmallicons")
         });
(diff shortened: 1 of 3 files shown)


nfagerlund:
Normal smallness on mobile:



Extra smallness on mobile: 



Normal journal style on desktop:

Screenshot_2019-06-23 betagal It's me again, what's up (2)

Site scheme with nothing special going on:

Screenshot_2019-06-23 betagal It's me again, what's up (3)

In responsive design mode with the "no woodchuck zoom" font bump enabled for touch:

Screenshot_2019-06-23 betagal It's me again, what's up (4)

zorkian:
Ziggurats were quite popular a few years ago. Time for a comeback!

PR #2518, opened 2019-06-24
userpicfactory.js: Revert to using an on-load handler

nfagerlund:
SOMETHING breaks if you try to do this setup as soon as the DOM is ready. Either
it's relying on image dimensions that aren't available until the image content
is loaded (been there), or there's some kind of existing race condition that
isn't exposed until this code gets faster. Hard to say, tbh.

### DIFF
--- a/htdocs/js/userpicfactory.js
+++ b/htdocs/js/userpicfactory.js
@@ -76,7 +76,8 @@ function toggleBorder (evt) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


alierak:
Sounds like you're right. I'm no expert but I've got the broken version under debugging in firefox and it does report that the image width and height were zero at the time it initialized the ImageRegionSelect.

PR #2533, opened 2019-06-27
Rewrite talkform javascript

### LINKED ISSUE #2532: Talkform: site user fields are hidden if no one's logged in, and also for openid
Screen_Shot_2019-06-25_at_4 29 05_PM

NUTS. I bet this is in the javascript somewhere, and I bet it started being a problem when I fixed that code that was never being run because it was attached to a nonexistent event. :|

nfagerlund:
Fixes #2532.

This javascript was a miserable pile of secrets.


views/journal/talkform.tt
```diff
@@ -219,98 +219,83 @@ the same terms as Perl itself.  For a copy of the license, please reference
...
-        </tr>
+        <div class="from-login" id="from-user-more">
+
+          <div>
+            [%- form.textbox(
+              name = 'userpost'
+              id = 'username'
+              label = dw.ml('Username')
+              size = 13
+              maxlength = 25
```
zorkian: We recently landed a diff to pull these "magic constants" out and use the $LJ::MAX_USERNAME_LENGTH or whatever the constant is. Would be nice to do that, might need to add it to the LJ::Web thingy too?

kareila: #2456 is the PR being referenced.

nfagerlund: oh hey good call

nfagerlund: Can we put that here? https://github.com/dreamwidth/dw-free/blob/master/cgi-bin/DW/Template.pm#L37 Or should it be wired through the template arguments?

htdocs/js/jquery.quickreply.js
```diff
@@ -156,23 +159,19 @@ jQuery(function($) {
...
         submitform(e);
     });
-    $("#submitpost").live("click", function(e){
-        var maxlength = 16000;
-        var length = $("#body").val().length;
-        if ( length > maxlength ) {
-            alert("Sorry, but your comment of " + length + " characters exceeds the maximum character length of " + maxlength + ". Please try shortening it and then post again");
-
+    $("#submitpost").on("click", function(e){
+        if ( checkCommentLength() ) {
```
nfagerlund: In the cold light of day I’m disliking this. I don’t think the helper code should export a global function. It should probably be a jquery extension, or better yet just attach it to the form submission event during the helper script. 

Is there a reason why the submit event might be too late for that, and we’d need to do it on click instead?

htdocs/js/jquery.quickreply.js
```diff
@@ -156,23 +159,19 @@ jQuery(function($) {
...
+            submitform(e);
+        } else {
             e.stopPropagation();
             e.preventDefault();
-        } else {
-            submitform(e);
         }
     });
-    $("#submitmoreopts").live("click", function(e) {
+    $("#submitmoreopts").on("click", function(e) {
```
nfagerlund: Oh, I should explain this. “Live” means it keeps an eye out for any new instances of the desired selection that get created for the lifetime of the page, and attach the listener to them as well. But we don’t spawn new reply forms willy nilly. 

Remember how the qrform used to get appended to the page with mid-stream js? It was probably related to that. That, or just paranoia maybe.

htdocs/js/jquery.talkform.js
```diff
@@ -0,0 +1,96 @@
...
+
+    // Reset text inputs and checkboxes to their initial values. Inputs can
+    // specify their initial value with the data-revert-value attribute,
+    // otherwise we assume empty/false.
+    jQuery.fn.extend({
+        resetFormFields: function() {
+            this.find('input').each(function(i, elm){
+                var type = elm.getAttribute('type');
+                if (type === 'checkbox' || type === 'radio') {
+                    elm.checked = ($(elm).data('revertValue') || false);
```
nfagerlund: Wait... this might be totally unnecessary. The parent form element’s reset() method might do this automatically. Need to test to make sure it does what I’m thinking, tho.

htdocs/scss/components/quick-reply.scss
```diff
@@ -102,16 +102,8 @@ $icon-control-height: 2.2em;
...
-        // This only exists to get something appended to it by jQuery. Its child
-        // input is the real flex item.
-        display: contents;
-
-        // Separate the "submit" buttons from the "misc" ones, if there's space.
-        // Quote is the first "misc" button.
-        input {
-            margin-left: auto;
-        }
+    #comment-text-quote {
```
nfagerlund: This id is anticipating a bunch of other markup helper buttons. Comment-text-em, -link, etc.

views/journal/talkform.tt
```diff
@@ -219,98 +219,83 @@ the same terms as Perl itself.  For a copy of the license, please reference
...
+              style = "background: url('${site.imgroot}/silk/identity/user.png') no-repeat; background-color: #fff; background-position: 0px 1px; padding-left: 18px; color: #00C; font-weight: bold;"
+            ) -%]
+          </div>
+
+          <div>
+            [%- form.password(
+              name = 'password'
+              id= 'password'
+              label = dw.ml('Password')
+              maxlength = 30
```
nfagerlund: Does this have a constant too?

htdocs/js/jquery.talkform.js
nfagerlund: Past Nick is half-right. reset()'s off the table because these aren't sub-forms and we don't want to whack the whole comment, but input.defaultValue and input.defaultChecked still let me skip that data element BS. nice.

views/journal/talkform.tt
nfagerlund: - wired through template is the practice in the previous PR. 
- lj::web thingy -- quickreply doesn't need a username, and idk if there's any javascript that has a use for it.

htdocs/js/jquery.quickreply.js
nfagerlund: Yes: on last-n pages, the quickreply submits as an ajax POST request, not a real form submission, and it does it in a goofy function called from the submit buttons instead of a submit handler. well, that's fixable.

views/journal/talkform.tt
nfagerlund: Doesn't appear to. 30 is what's used in most places, but I do think I saw a 50 somewhere random.

nfagerlund:
OK, I think this is in decent shape. Here's a pic of the one visual thing I changed. 

Screenshot_2019-06-27 sheworks4themoney Bumping up the friends page (Reply)

views/journal/talkform.tt
zorkian: I don't believe it has a constant, no.

zorkian:
Honestly I'm really bad at knowing exactly how things look today since there are so many permutations. Do you have a before-and-after screenshot?

nfagerlund:
Oh, one thing I forgot to mention — moving the submit behavior off the buttons in quickreply also patches that hole where you could skip the ajax submit and guard-rail checks by hitting enter in the subject field.

nfagerlund:
yeah! here's two "before" images:

image

image

#### Spotter's Guide:

- In the old one, the controls went _icon, radio button, label_. Now they go _radio button, icon, label_. 
- In the old one, if you're currently logged in, it hides the "Dreamwidth account" option; there was a magic "`[other]`" link that would hide the logged-in user option and replace it with the arbitrary user option. In the new one, it shows both options, which removes the mystery of "`[other]`" (\~the other\~) and also lets you switch back to your logged-in user without reloading the whole page. 
- In the old one, disabled options are represented by empty ascii parentheses. Now they're disabled radio buttons, in the platform's default styles.

Hmm, looks like I missed a class on the disabled labels, and the site scheme was greying text for that class. I'll catch that.

nfagerlund:
there we go, here's an updated "after": 

image

zorkian:
Thanks for the pictures! Very helpful, and nobody has commented going "oh no!" so we're going to go ahead.

rahaeli:
Belatedly, the "after" looks so much better, excellent job!

PR #2537, opened 2019-06-29
don't apply convert_user_mentions to imported and syndicated content

kareila:
NOTE: I haven't even tested to see if this compiles!  It involves adding new methods and memcache keys related to whether a given entry or comment has been imported.  The SQL is based off my reading of the erase-imported-content worker.

Just throwing this up for review and as a possible base for subsequent work.


zorkian:
We already have cached means of getting the properties, but I took inspiration from this diff and wrote #2540 which incorporates `is_syndicated` and `is_imported` flags, would appreciate your thoughts if you have a few.

zorkian:
Going to close this as it's superseded by #2540.

Issue #664, opened 2014-04-02
Implement RTE in new update page


afuna:
Perhaps not actually an RTE, but something that does preview / helps insert formatting without needing to know HTML / help with inserting an image or embed.

I'd like to experiment with a Markdown editor here.

afuna:
![screen shot 2015-03-13 at 4 27 55 pm](https://cloud.githubusercontent.com/assets/364169/6648947/0e3b6828-c99e-11e4-8df3-b182c12f560e.png)

^- So this is roughly what I'm thinking it'll look like. 

I initially considered using [Editor](http://lab.lepture.com/editor/markdown) but realized that it's basically a very thin layer over CodeMirror. We have to modify that layer for accessibility, and also would need to extend the underlying code editor for custom tags (cut, user head, etc), so I'm going to try to go the direct route.

I'm currently looking at both [Code Mirror](http://codemirror.net/doc/manual.html) and [ACE Editor](http://ace.c9.io/) to see which is easier to customize. Both are editors that can have the appearance of the input text tweaked (see the bolded/italicized text in the screenshot), without adding a lot of extra hidden HTML code the way RTEs do.

afuna:
One of the major concerns was that on DW, we do: one newline in input == <br> in output.

Markdown, however, ignores single newlines, only making an actual newline in the output when you have two newlines.

I think that there's enough precedence to continue supporting this behavior. We're not the only site to have to deal with this...

Options are either to silently insert two blank spaces where there's only a single newline followed by a non-blank space, so that the backend just parses it as regular markdown, or else to treat things entered via the web interface to be different from that entered for, say, replies to comment emails.

afuna:
The other thing I'm thinking of is our use of asterisks colliding with RPers.

Options:

* big old asterisk button that inserts an escaped asterisk
* use __ instead of * for emphasis (and make sure that's what we insert with the editor) 

Most (all?) markdown libraries will support "_", so it wouldn't be too weird using that, but it would a bit weird to *not* use the "*".

zorkian:
When you say precedent, do you mean on other sites? How do they handle the single newline?

I generally think we should adhere as much as possible to "Internet standard". I.e., what Facebook, Tumblr, Plurk, and Blogger do are probably what we should look at. Since most people bounce around to multiple sites, and we're a small fish in the pond, it makes sense to reduce friction as much as possible so people are more like to go "Yeah, DW is cool, it works for me!"

afuna:
[...]

PR #2540, opened 2019-07-01
Better Markdown, maybe

zorkian:
This moves Markdown and @-user conversions down into the HTML cleaner,
but only when we're processing text nodes and only when we're not in the
middle of HTML.

This means that for a user that is only using Markdown, things will work
great and everything will be "owned" by Markdown.

While using Markdown, you can also "drop to HTML" by opening any tag.
This will suspend Markdown processing and @-user mentions until you
close the tag. This enables you to drop in a textarea or whatever and
not have to stress about it.

For HTML users, we of course disable both Markdown and @-user mentions.

This also makes syndicated and imported content disable both of the
above to make sure we don't frotz them inappropriately.

I'm not convinced this is right yet, I slapped it together this
afternoon after some thinking and would like to write some more tests
before committing it.


zorkian:
There are two caveats here:

1) This deletes the 'fast path', which is probably OK.

2) I'm pretty sure (haven't tested) but it also means that the output of Markdown is not run through the cleaner. Which is probably OK?

zorkian:
Added tests and fixed the Twitter case (which uses `<blockquote>` to embed). I think this is better than where we're at today and also addresses most of the concerns?

cgi-bin/LJ/Entry.pm
```diff
@@ -906,9 +906,19 @@ sub event_html {
...
+    $opts->{is_imported}         = defined $self->{props}{import_source};
+    $opts->{editor}              = $self->{props}{editor};
 
     $self->_load_text unless $self->{_loaded_text};
     my $event = $self->{event};
+
+    # Old-style: if the entry has a special prefix then remove it and also turn on
+    # markdown formatting.
+    if ( $event =~ s/^\s*!markdown\s*\r?\n//s ) {
+        $opts->{editor} = 'markdown';
```
nfagerlund: 🙌🏼 hallelujah

nfagerlund:
I raised this in the channel, but I want to get it down here too. 

Instead of the current combinatorial approach with different settings affecting different parts of the behavior, I think we should lean more heavily on the `editor` property, and use it as the central record of all our content markup dialects. I think I see the following dialects in here:

- `markdown` — explicitly set, or detected when first line is `!markdown`. It’s markdown, w/ @-mentions, and can use other DW-specific tags. 
- `html-external` — explicitly set, or detected. For content from feeds, etc. Is plain html, no special linebreak behavior, don’t convert @-mentions, don’t process DW/LJ tags. 
- `html-dw` — can only be explicitly set. html w/ DW tags and @-mentions. No linebreak munging. 
- `html-dw-linebreaks` — Can only be explicitly set. html w/ DW tags and @-mentions. Treat linebreaks as br tags. 
- `html-legacy-dw` — Detected when internal content, no editor set, and don’t-autoformat is enabled. Html w/ DW tags, but no @-mentions. 
- `html-legacy-dw-linebreaks` — detected when internal content, no editor set, and don’t-autoformat disabled. Same as above, but with linebreak munging. 

Then we tweak all the comment and entry forms to specify their editor dialect. All new content gets one of the new behaviors; old content has @-mention amnesty. If we want, we can auto-detect new content from the last month, too. 

I think being able to say we’re always in one and only one dialect would be very freeing, especially if we ever need to change markup behavior in a disruptive way again (I bet we will).

Alyndra:
This is exciting! Only thing I would say to consider is that a lot of external feeds are from LJ and _do_ want you to process LJ tags. Not sure if anybody puts DW account feeds on their reading list, but I might not wholly rule it out. And of course quote text might be dw/lj text, or might not. I don't think it's a big deal if quoted text drops back into an @person.site format, especially if it can be fixed by changing it to <user name=person site=site>, but if _that_ code shows as raw code instead, it's a little uglier.

nfagerlund:
Good points. (But do lj feeds pre-process their special tags into plain spans and images before we get them?) 

Also, this might not have been clear in my comment above, but I _don’t_ think we should let users choose to _use_ the legacy editors! Other than the current three-or-so, the other editors should just be a backend construct to simplify presenting a mix of current and historical content. 

Raises the q of what should happen if you edit an old post or comment. I think the forms should handle it. The control for picking your formatting displays as locked with the old value. Leave it alone and you can keep it, but if you choose to unlock it, you can only choose a current format. And like @wohali was saying, RTE should be extra finicky about this.

wohali:
Just for posterity, because not everyone has access to Discord: the proposal I'm making is that if RTE 2.0 gets a legacy "newline = <br/>" post, it will refuse to switch into RTE mode and you'll only be able to edit it in unassisted text editing mode.

For new posts, the RTE will always write valid raw HTML, converting any newlines to <br/>s. And we may have the possibility for a Markdown mode in the RTE as well, though it is not clear yet.

That said I'm not sure we need `html-dw-linebreaks` for long, if the RTE gets done quickly. What do you think?

nfagerlund:
Mmm. I think old habits die hard, and that a lot of people actually like that weird hybrid html plus linebreaks thing. I don’t think they overlap much with RTE users, but I bet they’re an equal or greater peer to the markdown users, and I think it probably makes sense to treat that dialect like a peer of markdown. 

I have some thoughts about UI for that, but this ticket probably isn’t the place for that, and I’m not wholly up to speed on how ckeditor wants to handle source mode yet either

Alyndra:
Anecdotally, some fic authors are in the habit of writing with ^i^ tags and ^hr^ tags but not ^p^ tags, and copy/pasting into the entry page. Would the RTE would handle italics and other formatting being pasted in from Word or Docs?

kaberett:
Ducking in to say: you can't create feed accounts from the RSS feeds DW provides. If people really wanted to subscribe to feed accounts pulling from a DW RSS source they'd have to first feed it through a different aggregator, and frankly at that point I think it is legit to eclare them as On Their Own.

wohali:
Hi @Alyndra , let's keep RTE comments and questions over on this issue: #664 .

To answer you: the (non-Markdown-mode, good' ol) RTE absolutely handles rich text being pasted in from Word, Docs, WordPad, etc. That's one reason for the upgrade to a newer version of CKEditor 5. And my proposal is: without `<p>` tags, the RTE would insert those on save; a future edit would show the `<p>` or `<br/>` tags if you went into raw HTML mode.

rahaeli:
Belatedly, yes, both LJ and DW preprocess the LJ/DW specific tags (user, cut, etc) into standard HTML in RSS feeds, so that shouldn't be an issue.

Alyndra:
Okay, that’s good to know! 😅 My point about not disabling those tags (just the @ trick) for the external content option still stands, for imported entries and quoted content, though, right?

zorkian:
@nfagerlund so I think that I agree w/you around "`editor` shall be the thing that defines how we edit/interpret content", and I believe this diff takes us in that direction. Does it do anything we _don't_ want?

I guess basically I'm asking, are there objections to me merging this. I think it fixes some of the Markdown-in-weird-places issues. And is just More Correct since we're actually inserting Markdown handling into the HTML Cleaner.

nfagerlund (approved):
@zorkian I looked back through, and I _think_ this doesn't do anything we'd specifically need to un-do. I'm 👍🏼 to merge!

The use of `$formatting` instead of `$editor` inside the cleaner gave me pause earlier because it's an extra piece of indirection, but on further consideration that's probably good — most old content lacks an editor prop, and having an explicit conversion step will help make the mapping more clear.

nfagerlund:
(Well... the `local_content` thing might need to be refactored away later, but that’s not certain, and it’s pretty contained anyhow.)

zorkian:
Thank you! :D

PR #2543, opened 2019-07-02
Styles: Tabula Rasa and children: Break long words in headers

nfagerlund:
I already did this on a few specific headers, but really you never want
ANY header to blow out the page just because someone had to do a keyboard
smash. And I missed the page title previously.

### DIFF
--- a/styles/core2base/layout.s2
+++ b/styles/core2base/layout.s2
@@ -373,6 +373,10 @@ H1, H2, H3 {
     padding: .25em 0;
 }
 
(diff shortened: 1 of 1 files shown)


zorkian:
Holy shit this will bet rid of some stupid code!

https://github.com/dreamwidth/dw-free/blob/master/cgi-bin/LJ/CleanHTML.pm#L1275

nfagerlund:
Oh wow that is _criminal._

Issue #2522, opened 2019-06-24
QR: Link preview to icons page if no browse; move random into drop-down and spawn re-roll button


nfagerlund:
https://dw-maintenance.dreamwidth.org/78131.html?thread=2943539#cmt2943539

- Put it on the icon preview whenever we don't turn it into a browse button? 
    - But then paid users don't have the link. 
    - But do they even need it if they have the browser? (Assume we're going to fix #2411 at some point in the near future.)
- Put it somewhere else? 
- Leave it out, bc there are other ways to get there and we don't need to add links that navigate you away from the comment form?

rahaeli:
My vote goes to "leave it out because there are other ways to get there and adding the user tag somewhere will make /icons one click away anyway"

Alyndra:
I was making pro/con lists in my head for a couple options here even before this issue opened but...you make a very good point, once I realized from seeing this comment and running to check that the icon in the standard hoverbox was, in fact, clickable. Maybe it needs a "hover over the icon and it says "ICONS" where the other hover says "BROWSE," haha? I wouldn't have realized it was clickable without looking at the bottom of my web browser for an address.

rahaeli:
Ha, I was actually thinking of "click on the userhead icon to get to the profile, click on the icon link to get to the profile page" as the one click. Is there another way?

Alyndra:
If you hover on the userhead icon it has a short list of links and the user's default icon, and the icon is clickable (I JUST DISCOVERED A THING)

nfagerlund:
Yeah, that’s how I always get to my icons page. I like the idea of doing a “view all” mask that’s like the “browse” mask, tho!

rahaeli:
Ohhh, yeah, the contextual hover popup! (I have that disabled because I hate things flying around just because I moused over something, so I didn't think of it.)

Alyndra:
Thought about this some more. I am prepared to make a case that clicking on the preview icon should do _something_ icon-related instead of nothing, since for new or infrequent users who want to look at all their icons, it's much more intuitive to click the icon than to go the janky route of the userhead icon/profile.

At this point I wondered what current non-paid account usage looked like, so I made an account I'd been meaning to create anyway and played around with it. HIGHLY SCIENTIFIC RESULTS:

[...]

Issue #2521, opened 2019-06-24
QR: Bring back username indicator and move "more options" up top


nfagerlund:
cf. https://dw-maintenance.dreamwidth.org/78131.html?thread=2995763#cmt2995763 — some people with many accounts miss having the reply form display their logged-in username.

rahaeli:
It could fit pretty nicely just under the icon.

nfagerlund:
I was kind of thinking of having it off on the right-hand side, since that way it wouldn't increase the total height until everything gets super narrow. Is there anything especially good about having it under the icon?

rahaeli:
Both username and user icon are visual-markers-of-identity rather than settings, metadata, etc, for the comment being composed, so under the icon picker would mean someone could verify that they were logged in with the right account at the same time they're choosing an icon. (Checking which account you're logged into and checking that you used the right icon are also two common last-checks-before-posting, so IMO it's logical to have them together.)

(I also feel like putting username off to the right side would also look kinda detached-head-floating-in-space, but my reason for wanting it to the left under the icon is primarily function-grouping.)

Alyndra:
Whoops I was talking about this here: https://dw-maintenance.dreamwidth.org/78131.html?thread=3015731#cmt3015731

I think having it at the top right next to the icon (where Browse currently sits) would be closest to familiar for people, and would group by function pretty well too. (I admit part of my motivation is wanting the icon stuff all on one line, hee!)

nfagerlund:
@Alyndra You've also gotten me thinking about the "more options" button. If you're right about "more options" really only being about "more _user identity_ options" (and the more I think about it, the more I think you're right; subjecticon \[doomed\] and "don't autoformat" \[coming to quickreply at some point\] are the only other things in the talkform), then it probably belongs next to the username indicator, right? 

(By the way, did we ever fix the issue where "more options" throws away your comment text if you click it from your reading page and you don't yet have an up-to-date cookie from the subdomain of the journal you're replying to?)

kareila:
No, it’s #1484.

Alyndra:
@nfagerlund if it helps, this is what I'm picturing (note preservation of ziggurat shape, pls ignore username spelling fail):
![IMG_2767](https://user-images.githubusercontent.com/51764699/60057887-caf2e200-96ab-11e9-8463-7de82333f02d.JPG)
I wouldn't be at all opposed to fitting all three lines next to the icon instead of the subject line being under it, ftr. 

[...]

PR #2544, opened 2019-07-02
Quick-reply, 4th of July edition: don't squish, smarter button positions

### LINKED ISSUE #2528: QR: change order of post buttons
New qr has the main “post comment” button first, on the left. 

That came from the “quicker reply,” as with many things. The old less-quick reply had it on the right instead. 

Which place is correct? Left is nicer on mobile, but actually it’s “first” that’s nicer on mobile and flexbox lets you lay buttons out backwards if necessary, so it could be right and still first. That might be unintuitive to left-to-right language readers tho. 

Could also move everything right and bump the auxiliary buttons down a row. 

Personally I’m liking left well enough.

### LINKED ISSUE #2535: QR: Don't smoosh down if there's plenty of room to the left
Getting this right could be slightly annoying, but: if the QR is smaller than a certain minimum width, it should leave its right edge in place and kick leftwards to take up some of the extra padding that comes from comment indents. No more 200px chiclet forms. 

The minimum width should be 640px (40em @ 16px) OR the max width it would have if you were replying to a top-level comment, whichever is smaller. 

Because it has to escape its container AND keep its right margin in place (unnatural behaviors for HTML elements), this requires position: relative and some actual measurements in the javascript. 

Also, I'll note that this bears a family resemblance to another project that involved measuring pixel widths with jQuery, and that one turned out to be impossible (or at least impossible within the constraints I set for myself, i.e. no FOUC). But I've got some advantage here, because the math only has to happen on a user interaction (click reply), so... we'll see. There's still a chance that browsers' layout loops will have some nasty timing traps, and the trick might be in making sure I only get measurements from things that I won't be manipulating.

### LINKED ISSUE #2545: QR: Tighten up no-icon and no-default-icon behavior
- In Web.pm, add the no_icon image source to the default item, so it will come back if a no-defaulter switches away from and then back to the default. 
- In the template, when checking whether to wrap the browse button around the icon preview, check browser-allowed status AND the icon count; omit if there's no icons. 

other than those, I think this is in good shape.

nfagerlund:
Fixes #2528 
Fixes #2535 
Fixes #2545

I'm still poking at issues #2521 and #2522, because they bring up a bunch of sticky layout questions. But those really only mess with the uppermost section of the ziggurat, so there's no reason they should delay this batch of improvements. 

These should be pretty uncontroversial **(FAMOUS LAST WORDS)** -- 

- No one likes it when the reply form squishes down to a little chiclet.
- There's pretty strong consensus that the post button should go on the right. (I changed my mind after using it on the wrong side for a few days -- turns out I DO actually have a strong opinion about this.)
- It sounds like a lot of people were confused by the quote button's previous locations (like, on talkform it looks like part of the auto-format decision group). Having it by the subject makes some amount of consistent sense, and at least shouldn't be any _more_ confusing. It also shouldn't shrink the subject by too much, now that we aren't shrinking the form to chiclet size. 

## Screamshots

Desktop, site scheme, deep thread: 

Screenshot_2019-07-02 betagal It's me again, what's up (2)

Desktop, journal style, deep thread: 

Screenshot_2019-07-02 betagal It's me again, what's up

Mobile-size, journal style: 

And just for good measure, here's what it looks like in a _really old_ browser that hasn't been updated since 2012:

Screen Shot 2019-07-02 at Jul 2, 10 24AM

This is the final release of Camino (RIP), and it shows what the form looks like if you have NO flexbox support. It degrades relatively gracefully — the icon controls go below the icon preview instead of beside, and the submit buttons are in the wrong order/wrong alignment, because I'm using flexbox to make the post button go on the far right. (This is the right thing to do IMO, because 1. post should be first in the tab order, and 2. if the buttons wrap on mobile, post should be _closest_ to the body field while remaining on the right. Re-ordering the buttons would sacrifice at least point number 2, even though we could get number 1 back by using positive values in `taborder` (which is Considered Harmful).)

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -786,11 +786,15 @@ sub icons_for_remote {
                 push @pics, [ $res{"pickw_$i"}, $res{"pickwurl_$i"} ];
             }
             @pics = sort { lc( $a->[0] ) cmp lc( $b->[0] ) } @pics;
+            my $defaultpicurl =
+                  $res{defaultpicurl}
+                ? $res{defaultpicurl}
+                : $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src};
             @pics = (
                 {
                     value => "",
                     text  => LJ::Lang::ml('/talkpost.bml.opt.defpic'),
-                    data  => { url => $res{defaultpicurl} }
+                    data  => { url => $defaultpicurl }
                 },
                 map { { value => $_->[0], text => $_->[0], data => { url => $_->[1] } } } @pics
             );
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -786,10 +786,8 @@ sub icons_for_remote {
                 push @pics, [ $res{"pickw_$i"}, $res{"pickwurl_$i"} ];
             }
             @pics = sort { lc( $a->[0] ) cmp lc( $b->[0] ) } @pics;
-            my $defaultpicurl =
-                  $res{defaultpicurl}
-                ? $res{defaultpicurl}
-                : $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src};
+            my $defaultpicurl = $res{defaultpicurl}
+                || ( $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src} );
             @pics = (
                 {
                     value => "",
--- a/htdocs/js/jquery.quickreply.js
+++ b/htdocs/js/jquery.quickreply.js
@@ -2,9 +2,9 @@
 // See also jquery.talkform.js, jquery.replyforms.js
 (function($) {
 var _ok;
-var qrdiv;
 var customsubject;
 var previous;
+var firstCommentWidgetParent;
 
 function can_continue() {
     if ( _ok == undefined )
@@ -14,6 +14,10 @@ function can_continue() {
 }
 
 function update(data,widget) {
+    // There's three target patterns:
+    // - "entry-dw_dev-215060-reply" (lastn page, reply to entry)
+    // - "topcomment", "bottomcomment" (entry page, reply to entry)
+    // - "1236487" (entry page, reply to comment)
     var targetParts = data.target.split("-");
     if ( targetParts.length === 1 ) {
         data.dtid = data.target;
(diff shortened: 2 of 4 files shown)


nfagerlund:
Hmm. I wonder if a fallback of `float: right` would get the submit button into the right position on crap browsers? Worth trying, or don't worry about it?

nfagerlund:
image

Yeah, sure enough: if you just float all three of them right, the post button wins the first slot because it comes first. Could then do `@supports (display: flex) { #qrform .qr-footer input {float: none;} }`. Just a question of whether we care, I think.

nfagerlund:
Yeah, there we go.

Screen Shot 2019-07-02 at Jul 2, 5 15PM

nfagerlund:
This branch is currently deployed on http://www.rr-thrice.hack.dreamwidth.net/, and I pasted a bunch of invite codes in the `#dreamwidth` discord channel at around 3:30 PDT on Wednesday. So far it looks pretty solid; @alyndra turned up some unrelated bugs, and both @alyndra and ilyena_sylph turned up some config errors on my instance, but the quickreply is looking OK. Could probably still use a couple more people banging on it.

nfagerlund:
@Alyndra and Ilyena_sylph and @kaberett all poked at this a bit, and I don't think they turned up any showstoppers. Here's what they did turn up: 

#### The annoying one

iOS 11 and lower (Aly: 11.2.6. me: the final 10.x) make it REALLY HARD to select and edit text that has scrolled out of view in an `<input type="text">`. In fact, it's basically impossible. Textareas like the message body work OK, but for one-line inputs like the subject, you're SOL and basically have to select-all and start over. 

BUT: it's not a regression and it's not just us (try moving the cursor around in the github search field on an older iOS version). It's just a bad Safari bug, and they fixed it if you upgrade to iOS 12, and there's basically nothing we can do about it (IMO turning the subject into a textarea or something would be a mistake). So I don't believe this is a showstopper.

_However,_ because I'm moving the quote button to the subject line, that _does_ mean there's less visible room in the subject field on mobile, which makes it more likely to run into the problem. **Question:** Do we care about that enough to make the design funkier at mobile widths and move the quote button somewhere random? 

#### Some config errors on my dreamhack

- Captcha services not configured
- Free / paid feature caps don't match what we expect from prod

Could update the config templates, maybe, so new dreamhacks are closer to prod and have all features available. 

#### Bugs with unrelated features

- #2548 -- interactions between talkform and preview, not sure whether I introduced or not. Needs more investigation. 
- #2547 (merged)
- (no issue yet) -- a thing I was doing in the mobile comment threads prototype turns out to have fundamental flaws. (The bit with adding go-behind lines indefinitely and just hiding the ones that eventually go off the page.) Needs a rethink. 
- Gross cosmetic issues in Opera Mini -- working as expected 😂 If you turn on the "max savings" thing, it seems to still make the quickreply wider than the viewport, which is what this whole exercise was meant to prevent in the first place. But because it's doing a weird combination of serverside / clientside HTML rendering and since it's literally best known for making sites Get Weird, I don't think it's urgent. (Except when combined with the responsive threads prototype, see above.)

zorkian:
Re: iOS 11, I think that's okay. iOS 12 goes back to the iPhone 5S which was 6 years ago, so I think we're probably okay saying if you happen to be on an iPhone older than that, this one particular thing will be a bit weird sometimes.

I think the subset of people on that OS, who use subjects, who run into this, will be small enough that it's okay.

cgi-bin/LJ/Web.pm
```diff
@@ -786,11 +786,15 @@ sub icons_for_remote {
                 push @pics, [ $res{"pickw_$i"}, $res{"pickwurl_$i"} ];
             }
             @pics = sort { lc( $a->[0] ) cmp lc( $b->[0] ) } @pics;
+            my $defaultpicurl =
+                  $res{defaultpicurl}
+                ? $res{defaultpicurl}
+                : $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src};
```
zorkian: nit: Simpler and more canonical to write,

```perl
my $defaultpicurl = $res{defaultpicurl} || $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src};
```

htdocs/js/jquery.quickreply.js
```diff
@@ -45,6 +49,49 @@ function update(data,widget) {
     if ( ! customsubject || cur_subject == "" )
         subject.val(data.subject);
 
+    // If we previously munged the layout of #qrformdiv, reset it.
+    $('#qrformdiv').removeAttr('style').removeClass('width-adjusted');
+
+    // If replying to a comment, sort out the width. If the reply form would be
+    // super-small but there's plenty of whitespace to the left due to comment
+    // indentation, extend the form out to the left.
+    if ( data.target.match(/^\d+$/) ) {
```
zorkian: All this makes me wonder if a better UI is just to pop up a modal or something for comment entry when you click to reply. Is it actually important to us that the reply form end up embedded in the middle of the entry form?

cgi-bin/LJ/Web.pm
nfagerlund: ah, good call, thx.

zorkian: although make sure the precedence doesn't bite you ... I forget the precedence of `.` heh

htdocs/js/jquery.quickreply.js
nfagerlund: I think so. Or more specifically, the important part is being able to scroll up and down and remind yourself of the surrounding context as you're composing. Obscuring any part of the page would be iffy for that, and being a true modal that blocks scrolling would probably make people flip OUT. (I'd prolly flip out, at least.)

cgi-bin/LJ/Web.pm
nfagerlund: haha, I try to always use parens if there are boolean operators in the mix. Even if you get the precedence right, it's a mental DOS attack on whoever ends up reading the code next.

htdocs/js/jquery.quickreply.js
zorkian: Yeah, I don't mean a modal-modal, more just meant a frame of some kind that doesn't change the flow of the page. Definitely not worth touching right now but feels like it would be a more sustainable solution with all the different style madness things people can do. Food for thought I suppose.

Issue #2549, opened 2019-07-08
Return to same page on logout


nfagerlund:
The controlstrip logout button used to return you to the same page, but #2230 broke that by removing the `?ret=1` parameter. RPers find that inconvenient, and after spending some time testing stuff with different accounts on dev, it's been bothering me as well. I think we should try to find some way to restore that behavior.

This is probably only necessary for the controlstrip logout button?

nfagerlund:
https://github.com/dreamwidth/dw-free/pull/2230/files#diff-b8e87a4762715497c6fc11427a35f336L59

It looks like the old implementation maybe had some optional goofy behavior where you could post an arbitrary destination, but never mind that. The basic `ret=1` version just checked whether the Referrer header came from any of the site's subdomains, and OKed you to go back there if so.

PR #2550, opened 2019-07-08
[#2549] Return to journal pages on logout

nfagerlund:
This restores the ability to click the log out button in the controlstrip and
get returned to the same page, much beloved by RPers and other multi-account
users.

Unlike the original implementation, I'm using a hidden form input instead of a
URL parameter (`?ret=1`) -- since this is a POST action, it seems like the
arguments should all arrive via POST data.

This also fixes an inconsequential copypasta glitch in the controller's name.

----

I only tested this briefly. 

- Works when logging out from control strip (returns you to page).
- Works when logging out elsewhere (goes to logout page). 
- Properly sends you to the logout page with error displayed if you, e.g., delete the control strip's form auth input with the web inspector on a journal page. 
- Seems to properly reject bad referers. (Had to test that with just dumping debug data to the page; guess I could also steal a live form_auth and try to post from somewhere random, but I didn't do that.) 

I'm passing the referer directly to that `LJ::check_referer()` function, because I didn't trust the fallback of grabbing it out of BML. (Maybe that works fine on non-BML pages? Couldn't tell.)

### DIFF
--- a/cgi-bin/DW/Controller/Auth.pm
+++ b/cgi-bin/DW/Controller/Auth.pm
@@ -15,7 +15,7 @@
 # 'perldoc perlartistic' or 'perldoc perlgpl'.
 #
 
-package DW::Controller::Rename;
+package DW::Controller::Auth;
 
 use strict;
 use v5.10;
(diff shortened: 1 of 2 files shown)


cgi-bin/DW/Controller/Auth.pm
```diff
@@ -15,7 +15,7 @@
 # 'perldoc perlartistic' or 'perldoc perlgpl'.
 #
 
-package DW::Controller::Rename;
```
zorkian: lol oops

PR #2554, opened 2019-07-11
[#2549] Return to same page when logging out from site header

nfagerlund:
Same as #2550 but for site header. This also stops hiding the header username and logout button on /logout, because all that extra complexity bought us was a glitched appearance and worse UX.

### DIFF
--- a/cgi-bin/DW/Template/Plugin/SiteScheme.pm
+++ b/cgi-bin/DW/Template/Plugin/SiteScheme.pm
@@ -84,10 +84,6 @@ sub challenge_generate {
     return LJ::challenge_generate(@_);
 }
 
-sub show_logout_button {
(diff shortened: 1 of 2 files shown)


nfagerlund:
While testing, I also checked the in-page logout buttons at /logout, to make sure re-adding the header button didn't interfere with them. They both worked like they're supposed to, logging out and displaying the result page with the appropriate message. 

Although you can get a humorous error if you browse directly to /logout without being logged in and click one of the in-page logout buttons. (That's an existing thing on prod too, and I don't think it's worth any special handling. Anyone doing that already knows what they did wrong.)

zorkian:
Going back to the page you were on is definitely the better behavior IMO, thanks!

PR #2556, opened 2019-07-12
[#2555] Upgrade Foundation to 5.5.3

### LINKED ISSUE #2555: Upgrade Foundation to 5.5.3
Current version (5.4.7?) locks us into Compass 0.12 and Sass 3.2, both of which are ancient and deprecated. Also, the dreamhack box is using the newer versions and it's messing stuff up pretty badly. (10000 line CSS files with 15 copies of the same rule.)

I've got this partially completed in a branch; it's working, but there are some style issues to clear up and a few logistical complexities.

- Logistics: In order to deploy the update, we'll need to simultaneously deploy patches to dw-free, patches to dw-nonfree, and upgrades to the Compass (at least 1.0) and Sass (at least 3.4) gems on production servers. Skip any one of those, and it breaks. 
- All:
    - [x] The box-shadow style to indicate focused text inputs is gone. (A default value for a setting changed.) 
    - [ ] The close "x" in most modals is positioned a little further to the left, now. It's possible for it to overlap text in the content area of the modal, just by a character or two. 
    - [x] At mobile widths, the account links modal doesn't slide in anymore; it just fades in.
    - [x] Existing bug: The close "x" in that modal is impossible to tap on real mobile. Btw, just what the heck was this _supposed_ to look like? Can it just be one of those icon buttons instead of whatever _that_ is?

        
    - [x] `<code><pre>` blocks got bigger text (you can see some on the style guide page). I don't know that we care about this! 
- Tropo:
    - [x] The "go" button in the top bar is funky. They changed how they calculate its height and offset.
    - [x] At small widths, the masthead and the action links userpic get messed up. (Either height calculation changed, or ordering of some rules got messed up. Needs more investigation.)
- Gradation: 
    - [x] Existing bug: the logo image's background doesn't match the background gradient at mobile widths.
    - [x] New bug: the userpic in the header doesn't disappear at mobile widths, and it's doing something SUPER goofy position-wise to boot. Probably a new effect of whatever hit the masthead height on tropo.
- Celerity: 
    - [x] Has the same masthead height issue at mobile widths as tropo, with probably the same cause.
    - [x] Existing bug: The content zone has a min-height based on the length of the sidebar nav, and it's too short; the extra items added to the nav on prod make the sidebar longer than expected, so it overlaps the footer and hides the stripey-stutter-out thing. 
- Gradation Vertical: 
    - [x] Combines Gradation's issues and Celerity's issues. The sidebar overlap thing's even worse here, because I don't think it actually sets a min-height on content, or else it's WAY small. 
- Lynx: 
    - Seems fine!!

nfagerlund:
Fixes #2555 

🚨 **This is not a normal pull request.** 🚨

- These changes and https://github.com/dreamwidth/dw-nonfree/pull/127 **must** be deployed at the same time. Otherwise a tiny bit of Tropo is incompatible with the new version, and compilation will fail. nonfree 127 **cannot** be deployed ahead of time. 
- At the time these changes are deployed, the Sass and Compass gems on any prod machines **must** be upgraded to 3.4.x and 1.0.x, respectively. They **cannot** be upgraded ahead of time, as they're incompatible with the current version of Foundation. At the time of upgrade, you should probably also uninstall the previous gem versions. 
- We should probably ask @afuna very nicely to review it, because I'm not seeing anyone else's name on any of the previous Foundation upgrades. @kareila might know some things too, tho.

Note that there will probably never be another Foundation 5 release, so we're now into "you break it, you bought it" territory w/r/t any buggy upstream behavior.


zorkian:
I pinged @afuna to ask if she has time, but if she doesn't, I'll take a stab.

htdocs/js/foundation/foundation/foundation.abide.js
```diff
@@ -4,19 +4,24 @@
   Foundation.libs.abide = {
     name : 'abide',
 
-    version : '5.4.7',
+    version : '{{VERSION}}',
```
afuna: {{VERSION}}? (same question for all the other files)

htdocs/scss/foundation/_settings.scss
```diff
@@ -636,6 +660,7 @@ $input-font-size: rem-calc(16);
 // $input-disabled-bg: $gainsboro;
 // $input-disabled-cursor: $cursor-default-value;
 // $input-box-shadow: inset 0 1px 2px rgba(0,0,0,0.1);
+$input-include-glowing-effect: true;
```
afuna: are we deliberately enabling this?

htdocs/js/foundation/foundation/foundation.reveal.js
```diff
@@ -489,7 +489,7 @@
     },
 
     cache_offset : function (modal) {
-      var offset = modal.show().height() + parseInt(modal.css('top'), 10) + modal.scrollY;
+      var offset = modal.show().height() + parseInt(modal.css('top'), 10);
```
afuna: the comment for this commit is amazing

afuna:
I looked at things commit by commit and it all seems to make sense!

htdocs/js/foundation/foundation/foundation.abide.js
nfagerlund: OH, uhhhhhhh... good catch? I'm not actually sure what this is! I SAW it, and assumed they'd changed something about how they store their version numbers, but now that I check the old v5.4.7 tag in the upstream repo, that has `{{VERSION}}` also. So, this must be an artifact of pasting from the source rather than a build. I'll go do some rebasing and see what I can learn.

nfagerlund: Yeah, I grabbed a build to double-check, and sure enough. Pushed a commit with the fix.

htdocs/scss/foundation/_settings.scss
nfagerlund: I think you already saw this in the per-commit review, but yeah, this is one where we relied on the default and it changed out from under us.

zorkian:
Thanks for the review! I'll merge this and the nonfree, and plan to update these gems next code push.

Issue #2427, opened 2019-05-01
Make S2 site scheme pages responsive/Foundation-ized


nfagerlund:
Most site scheme pages are either BML (old, nonresponsive) or controller+template (new, usually converted to Foundation styling). To make a "normal" site page responsive, you convert its implementation to controller+TT, throw in a one-liner, and clean up any resulting shrapnel. (At least, that's what I'm led to believe.) 

HOWEVER, there's another class of site scheme pages that _pass through S2 instead of BML_ (styles/siteviews/layout.s2) and which act like a weird hybrid between user-styled journal pages and "normal" site scheme pages. The ones I know of are: 

- Entry page (click "reload in: site" in control strip, or disable "entry pages in my style")
- Reply page AKA slowreply (same as above)
- Icons page (always, unless the option to use your style is enabled. ...Does anyone use that? I always forget about it.)
- Tags, year-archive, month-subjects (can't ever be reached via links, but if you paste `?style=site` in the url bar, they appear to work perfectly fine)
- Reading, Network, Recent (same as previous, except they actually do look suuuuper broken)
- ~Profile??? Or does that go through BML???? Hard to tell!~ The profile page seems to be normal BML. IMO it should be high-priority because it's high-profile, but it's not part of this issue!

None of these pages use Foundation, and the usual conversion workflow doesn't apply. But the entry page and icons page in particular are probably in at least the top five high-profile nonresponsive site scheme pages remaining, so there's a big win to be had by making them work right on a dang telephone. 

The terrain is weird, but I've made some tentative progress on this. More to come. 

### WIP branch

https://github.com/nfagerlund/dw-free/tree/apr19_sitescheme_hell

### Dupes

- #1698 is a special sub-case of this issue. 

### Known blockers or semi-blockers

- [x] #2434 (midstream jquery: actual blocker, qr breaks if you don't do this)
- [ ] #2432 (adjustable comment indents: if we have to redesign the entry page from scratch anyway, let's not do anything we'll immediately want to undo)
- [x] #2411 (the "old new" userpic browser loses all its styling once you go to Foundation land, and I don't want to waste time re-styling it if its days are numbered anyway)
- [x] #2437 (extract talkform: I know this is going to get in the way at some point, I just KNOW it)

rahaeli:
You are amazing for even considering tackling this -- we kind of all looked at it and went "arrrgh".

[...]

Issue #2411, opened 2019-03-26
Icon browser / `iconselector` pop-up is useless on mobile


nfagerlund:
The "new-new" icon browser (first appeared on beta new entry page) is superior to the "old-new" one, so we should use it on journal pages. 

This requires making Foundation components available in S2 pages (which I've already done), and it'll be gated behind the same beta as the site-scheme comment pages refresh. 

Miscellaneous to-dos:

- [x] there are like four different ways to get the list of icons, and also the new entry page uses a different approach for the random icon and update-the-preview-icon functions. Can we standardize all this? (It looks like the icon list in the template doesn’t matter for the browser maybe? Like it’s getting its data via xhr from somewhere.)
- [x] Gotta wire the existing controls up correctly. 
- [x] Gotta clean up some of the foundationizing stuff where I was sloppy because I didn’t know if it was even gonna work. 
- [x] Get approval for the general approach of bringing foundation components into journal S2 land. 
- [x] Figure out what was going on with need_res getting ignored in .tt templates, but only under certain conditions. 
- [x] Clean up resource groups shenanigans in general. 
- [x] When ther're multiple keywords, you're supposed to style the currently selected one in the browser. Most s2 styles don't have anything for that; gotta figure out a way to do a fallback for it. 
- [x] Selected icon should be styled with a highlight-y border, but s2 styles don't have an affordance for that. And picking an arbitrary one will look bad with many styles. 
- [x] Something I did to fix that breakage in the modal slide animation might have caused the page to jump to the top when it first does its zwoop-zwoop thingie sometimes. Wonder wtf's up with that. Need reliable repro; only happens rarely. 
- [x] Enable new-new browser in talkform, since the jquery one's disabled now if you're in the beta. 
- [x] existing issue: keywords look bad if they wrap. 
- [x] “select” button too small and out of the way. (Less of an issue if there's alternate ways to finalize on mobile.)
- [x] no double-click finalize on mobile. (Less of an issue if select button is easier to get to.)
- [x] on mobile, if you scroll a mile to find your icon, you're way far away from the comment form when you finish and close it. Should be able to bounce back to home? 
- [x] foundation "block grid" is a crap replacement for flexbox. huck it. 
- [x] Figure out how to style it on journal style pages. (oh heck yeah, just shim into the thing core2 does for contextual popups!) 
- [x] Make second-click-confirms more discoverable. (I'm just putting a note at the top, screw it.)
[...]

PR #2558, opened 2019-07-15
[#2427] Responsive site-scheme comment pages (minimal)

nfagerlund:
OK, this is probably at a stage where it'd be useful to have someone look at it.

This is the minimal form of #2427. It is gated behind a beta feature. It is deployed on my dreamhack at http://rr-thrice.hack.dreamwidth.net.

### Goals: 

- _Mostly_ reproduce the current site-scheme comment page layout. (It's going to look somewhat different; the radically different base CSS just makes that unavoidable. But all the essentials ought to be in place.) 
- Make site-scheme comments not completely broken on mobile. Among other things, it should help with the outrageous layout crimes we keep seeing from those Android browsers. 

### Specifically _not_ the goal:

- Redesign the site-scheme comment pages to be a _good reading experience_ on mobile. 

This is just a patch job, and it deliberately doesn't do anything about the outlandish amounts of sideways slop, the various wastes of precious horizontal space, or the fact that indentation is a completely useless way to represent comment threading once you're on a screen where a two-paragraph comment will take up 100% of the available vertical room. It also doesn't do anything about #2411.
 
What it _does_ do is make sure the full width of a comment or entry can fit on the screen at a reasonable font size, and that the various controls have usable dimensions. 

### Things that aren't done or that I'm not satisfied with

- The icons page gets affected by this switch too, so it also needs some new CSS. It's simpler than the entry page, at least. Might even be able to punt and use Foundation's built-in grid stuff, if it's available as mixins and not just as markup classes. 
- The reply page is affected too... but actually, it's not in too bad a shape. I'm as amazed as you are. The talkform doesn't look _the same,_ but it also doesn't look _worse._
- The CSS is kind of messy. Needs a cleanup before it's ready for primetime.
- IDK if there's actually anything that can be done about it, but 80% of the effort on this was fighting with Foundation's global styles. Especially on the typographical basics. (The full 16px seemed way too fluffy, so I set the base font size to .85rem, and Foundation REALLY didn't want me doing that.) 
- Should probably go through that thread Momiji linked me about the last time she tried to re-create this page's design from scratch, and make sure I have an answer to all of the notable complaints from it. (I feel like a lot of those got answered by changes to the markup, which should mean I inherit the win without any extra effort. But we'll see.) 
- Light style might be less lite in Foundation land. I think at least an extra 100kb of CSS... maybe more images as well? I'm not quite sure how to measure this, especially the JS (since it's not getting minified in this dev env). But otoh, most of that should be cached if you've ever visited a main site page. Only time it might hose you is if you're coming from /mobile? Just too much about this bit that I'm unsure about.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -215,8 +215,6 @@ sub make_journal {
             )
     );
 
-    # Include any head stc or js head content
-    LJ::Hooks::run_hooks( "need_res_for_journals", $u );
     my $extra_js = LJ::statusvis_message_js($u);
 
     # this will cause double-JS and likely cause issues if called during siteviews
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -4988,7 +4988,10 @@ sub Siteviews__need_res {
 sub Siteviews__active_resource_group {
     my ( $ctx, $this, $grp ) = @_;
     die "Siteviews doesn't work standalone" unless $ctx->[S2::SCRATCH]->{siteviews_enabled};
-    LJ::set_active_resource_group($grp);
+    my $remote = LJ::get_remote();
+    if ( LJ::BetaFeatures->user_in_beta( $remote => "mobilesiteschemecomments") ) {
+        LJ::set_active_resource_group($grp);
+    }
 }
 
 sub Siteviews__start_capture {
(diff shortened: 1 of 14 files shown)


nfagerlund:
Work on this has progressed in a different branch, and ended up incorporating the new-new icon browser into all journal pages (not just siteskin ones). Closing this expo PR, will revive later with a real PR. 

FYI, current blockers on filing that PR are:

- [ ] Confirm the best-for-now layout for the upper area of the quickreply. I need to rearrange the talkform to make reply pages not look nasty, so I'm going to base it on the quickreply to reduce the amount of work that takes. (Without throwing away any talkform-specific stuff for the time being.) 
- [x] I think that's it, actually!

PR #2561, opened 2019-07-24
Fix some miscellaneous siteskin bugs

nfagerlund:
My Foundation upgrade PR (#2556) had several commits that fixed existing bugs and weren't dependent on the upgrade, so I rearranged it to put those first. I'm also filing those commits as a separate PR, so we can get the bug fixes today without having to wait on a traumatic upgrade.

### $small variable typo thing

[Relevant upstream commit](https://github.com/zurb/foundation-sites/commit/aca90d07f63fce559e971eb862887e4a0dc1c2ad)

### Gradation logo thing

Before / After:

 

Obviously the correct solution is to replace that logo image with a PNG with an alpha channel, instead of just baking the gradient in. If someone has easy access to the original source images, please let's just do that instead. 

### Account-links thing

Before / After:

### DIFF
--- a/htdocs/scss/skins/_nav.scss
+++ b/htdocs/scss/skins/_nav.scss
@@ -62,6 +62,8 @@ $nav-small-screen-header-height: 3em;
               $bg:$reveal-modal-bg,
               // Set reveal border style. Default: $reveal-border-style or solid
               $border:true,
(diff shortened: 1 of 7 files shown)


nfagerlund:
(Will need to rebase #2556 once this is merged, but that's a-ok by me.)

Issue #2557, opened 2019-07-12
Tagnav control is wholly inscrutable


nfagerlund:
(worth addressing #1303 at the same time, maybe.) 

That (+) / (-) icon for moving to the next / previous tag is amazingly obscure, and it was years before I understood what it does. That iconography _really looks_ like it's for adding or removing tags! But somehow it's not??

Honestly, it might not belong on the tag list anyway -- it should probably be a drop-down indicator next to the next entry / previous entry controls. That would also help clarify why it only appears on entry pages and not recent/reading pages -- all of those page types have tag lists, but only entry pages have per-entry next/previous controls.

SporkyRat:
The tag nav thing is something you might want to discuss with @woggy about, they wrote the original code.

woggy:
While the original tagnav code was my work (pull request: https://github.com/dreamwidth/dw-free/pull/947 ), it didn't actually use the +/- icon/hover. I just added Unicode arrows in the tags list, which some folks disliked.

The icon thing was added by @afuna ( https://github.com/dreamwidth/dw-free/pull/1206 , plus a couple adjustments/bug fixes since then).

Unfortunately I'm not likely to be much help in brainstorming, as my personal opinion is that the Unicode arrows were just fine (if slightly cluttered on entries with an abundance of tags.)

nfagerlund:
Oh interesting, thank you both for the history! That definitely clears some things up. I can see why people had issues with the unicode arrows thing, but at least it made obvious sense. 

After reading those threads (especially #1126), I really feel like a modal modifier to the next/prev links would be a better move. Well.... someday.

PR #2566, opened 2019-08-09
[#1303] Fix wandering tagnav dialog

nfagerlund:
The "flipfit" collision recovery option moves the dialog until it fits inside
the current viewport. But in a long journal entry in one of the normal journal
styles (not siteviews), it's meant to be off-screen, down by where the tags
live. (In siteviews this isn't an issue because the tags live ABOVE the entry.)

If you change it to just "flip", it starts off-screen (as is proper), but then
the jQuery UI dialog script calls `.focus()` on part of the dialog, which causes
the browser to scroll down to put it into view. So THAT's obnoxious. But you can
recover by just scrolling the viewport back to where it was. (Why not just
scroll to 0? Because the user might have already scrolled a little bit before
the `load` event fired, or they might be using the back button and the browser
remembers its prior position.)

### DIFF
--- a/htdocs/js/journals/jquery.tag-nav.js
+++ b/htdocs/js/journals/jquery.tag-nav.js
@@ -38,7 +38,7 @@
                             my: "center top",
                             at: "center bottom",
                             of: this,
-                            collision: "flipfit"
+                            collision: "flip"
                         },
                         width: "auto",
                         open: function(e, ui) {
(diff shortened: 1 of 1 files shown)


nfagerlund:
I should note that this save-and-restore scroll position hack is Bad, and this control should instead use a widget that DOESN'T manipulate the active keyboard focus whenever it appears. But I also think the design of the tagnav control needs a full rethink (#2557), which would make the extra work to bulletproof this design basically a waste. So, a minimal hack is the move.

PR #2574, opened 2019-08-28
DRAFT, DON'T MERGE: QR / Talkform redesigns, revamped site skin comment pages, new icon browser

nfagerlund:
This is an omnibus monster PR, and mostly just exists to be a record of the various dependencies in here. The changes in it should be merged as a series of smaller PRs, as described below. 

## Split PRs for Easier Review

Since the mega-version is looking kind of impossible to review, I've split this into eight smaller PRs, each one chained against the last. They're categorized by an intersection of change-type and relative risk. 

1. <del>[Aug19 megamix 1: safe noops](https://github.com/dreamwidth/dw-free/pull/2575)</del> _(done)_
2. <del>[Aug19 megamix 2: middlin noops](https://github.com/dreamwidth/dw-free/pull/2579)</del> _(done)_
3. <del>[Aug19 megamix 3: safe bugfixes](https://github.com/dreamwidth/dw-free/pull/2580)</del> _(done)_
4. <del>[Aug19 megamix 4: safe frontend (next quick-reply design iteration)](https://github.com/dreamwidth/dw-free/pull/2581)</del> _(done)_
5. [Aug19 megamix 5: risky core (`s2foundation` beta)](https://github.com/nfagerlund/dw-free/pull/5) _(ready)_
    - Interlude: [A dw-nonfree PR that should be merged immediately after that one.](https://github.com/dreamwidth/dw-nonfree/pull/128)
6. [Aug19 megamix 6: safe, complex frontend](https://github.com/nfagerlund/dw-free/pull/6)
7. [Aug19 megamix 7: safe frontend bugfixes (talkform stuff)](https://github.com/nfagerlund/dw-free/pull/7)
8. [Aug19 megamix 8: risky noops](https://github.com/nfagerlund/dw-free/pull/8)

7 and 8 are both very low priority compared to the rest of these.

### Demo

http://rr-thrice.hack.dreamwidth.net

My dev server is running all of this. Please kick the tires. Invite codes:

3. Invite code: 5GNA9R3SWM656AAAAAAN
4. Invite code: RJDSZBFADHS8KAAAAAAP
5. Invite code: CJY8TFKMDDEPQAAAAAAQ
6. Invite code: AQGBTBMRE6XRNAAAAAAR

I can post some screenshots later, too.

### Notable Stuff In Here 

- The next quick-reply design iteration.
- **Beta:** Foundation in S2 Pages. Makes site-scheme entry pages mobile-friendly, and enables new features in journal-styled S2 pages. 
[...]

PR #2575, opened 2019-08-28
Cleanups and groundwork for upcoming features (Aug19 megamix 1: safe noops)

nfagerlund:
This is a mergeable version of nfagerlund#1.
This is a subset of #2574. 
Doesn't close any issues.

These commits are:

- Intended as no-ops. They might enable some future visible changes, but should have no effect in their present form. 
- Very low risk. 
- Mostly front-end. Many small CSS tweaks, removing unused libraries from pages, etc.

The only notable one where there might be a design question is 7d40a77, enabling configuration in the modal component. Also, a sanity-check on the docs one.

### DIFF
--- a/cgi-bin/LJ/S2/ReplyPage.pm
+++ b/cgi-bin/LJ/S2/ReplyPage.pm
@@ -59,7 +59,6 @@ sub ReplyPage {
 
     $p->{'head_content'} .= $LJ::COMMON_CODE{'chalresp_js'};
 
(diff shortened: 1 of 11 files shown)


zorkian:
Docs look good, savage but correct.

zorkian:
I did a merge commit, which IIRC you had asked for before

nfagerlund:
Thank you, that helps a lot!

PR #2579, opened 2019-09-28
Cleanups and groundwork for upcoming features (Aug19 megamix 2: middlin noops)

nfagerlund:
This is a subset of #2574.
Doesn't close any issues.

These commits are:

- Intended as no-ops. They remove what appears to be dead code, and refactor an inconsistent function. 
- Backend (middle-end?) code.
- **Medium risk.** Please evaluate to make sure that code is as dead as I thought it was. ☠️

### DIFF
--- a/cgi-bin/LJ/Widget/UserpicSelector.pm
+++ b/cgi-bin/LJ/Widget/UserpicSelector.pm
@@ -21,8 +21,11 @@ use LJ::Talk;
 
 sub need_res {
 
-    # force to not use beta, because this is not used in journal spaces
-    return LJ::Talk::init_iconbrowser_js( 0, 'stc/entry.css' );
+    # Just let need_res work normally for the main stuff
+    LJ::Talk::init_iconbrowser_js();
+
+    # Hand off extra stuff to LJ::Widget's weird need_res shim
+    return ('stc/entry.css');
 }
 
 sub handle_post {
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -215,8 +215,6 @@ sub make_journal {
             )
     );
 
-    # Include any head stc or js head content
-    LJ::Hooks::run_hooks( "need_res_for_journals", $u );
     my $extra_js = LJ::statusvis_message_js($u);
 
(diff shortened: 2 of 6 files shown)


rahaeli:
No idea about the other two commits, but you can generally remove any call to a hook that doesn't have a corresponding file in dw-nonfree's cgi-bin/DW/Hooks or dw-free's cgi-bin/DW/Hooks -- hooks are the way we used to link to LJ's closed source bits from base LJ code and we didn't bother reimplementing most of them because they were (for the most part) marketing or advertising bullshit and/or referencing distinctions or account structure bits that we didn't keep.

cgi-bin/LJ/Widget/UserpicSelector.pm
```diff
@@ -21,8 +21,10 @@ use LJ::Talk;
 
 sub need_res {
 
-    # force to not use beta, because this is not used in journal spaces
-    return LJ::Talk::init_iconbrowser_js( 0, 'stc/entry.css' );
+    # Just let need_res work normally for this, instead of using LJ::Widget's
+    # berserk reimplementation of it.
+    LJ::Talk::init_iconbrowser_js('stc/entry.css');
```
nfagerlund: Actually, looking further at this, I think the whole "take extra includes as arguments" thing is ALSO just an artifact of the weird widget thing. Can probably remove that too, have init_iconbrowser take no args, and just return that extra stc/entry.css include from this method for the normal widget need_res system to deal with.

nfagerlund:
- It looks like the `stylesheet_override` option was only ever used in [_one_ spot](https://github.com/apparentlymart/livejournal/blob/8c24ffae22479331728a543ca87b7d6d5d64be8b/cgi-bin/LJ/Widget/Browse.pm) in the old LJ code. Decent chance there was some closed-source stuff they were doing with it, but definitely nothing in dw-free or dw-nonfree ever uses this. 
- Why does widget have its own need_res shim, anyway? Two reasons:
    - If you pass it a filename without a path, it finds it in a special directory namespaced to the widget. None of the remaining widgets use this feature. 
    - If you pass that stylesheet_override option, it'll strip out all the CSS files before actually calling need_res. None of the remaining widgets use this feature.

PR #2581, opened 2019-10-03
Quick-reply design updates: identity and button tidying (Aug19 megamix 4: safe frontend qr)

### LINKED ISSUE #1484: entered text lost when switching from Reading Page QuickReply to full comment form
If you click on a 'reply' link on your reading page and trigger a QuickerReply form, but decide you want more options and hit the 'more options' button after entering text, the text you've already entered is lost. We should preserve anything you've already entered and pre-fill it on the More Options form instead.

### LINKED ISSUE #2521: QR: Bring back username indicator and move "more options" up top

### LINKED ISSUE #2522: QR: Link preview to icons page if no browse; move random into drop-down and spawn re-roll button

nfagerlund:
This is a subset of #2574.
Fixes #2521
Fixes #2522
Fixes #1484 

These commits are:

- Front-end only. 
- Low-risk in terms of breaking the site.
- Medium-risk in terms of bothering the userbase.
    - But, this iteration has been extensively discussed in Discord, GitHub issues, and an ad-hoc alpha on dw-dev.
- Not dependent on anything that isn't already merged.

This design update could use functional and aesthetic review from a site owner (hey @rahaeli). Again, we've talked all these changes through before, but let's dot/cross the final eyes/tees.

-----

Meet the new quick-reply:

To play along at home, log into http://www.rr-thrice.hack.dreamwidth.net/ and mess with the reply form there — you can use your DW openid if you don't have an invite code. Active test users with threads to reply to are "betagal" and "sheworks4themoney". 

Spotter's guide: 

- There's only one "browse" button now, and it's on the icon preview. It's clearly labeled when the form first loads; the label fades after an icon is selected, but reappears on hover.
- If you can't use the icon browser, the icon preview links to your icons page (just like most icons do).
- The random icon feature has **moved to the icons menu.** Easily accessible if you want it, tidily out of the way if you don't. After the first use, it reveals a dedicated "random" button (for re-rolls).
- Username indicator is back.
- "More options" is by the identity controls, and appears as a link.

-----

Other stuff:

- This PR has a fix for "more options ripped my flesh;" it's in a separate commit.
- I made sure you can resize the textarea! 😭

### DIFF
--- a/htdocs/js/jquery.quickreply.js
+++ b/htdocs/js/jquery.quickreply.js
@@ -49,6 +49,8 @@ function update(data,widget) {
     if ( ! customsubject || cur_subject == "" )
         subject.val(data.subject);
 
+    $("#prop_picture_keyword").change();
+
     // If we previously munged the layout of #qrformdiv, reset it.
     $('#qrformdiv').removeAttr('style').removeClass('width-adjusted');
 
@@ -225,18 +227,24 @@ jQuery(function($) {
         $("#qrform").attr("action", Site.siteroot + "/talkpost_do" );
     });
     $("#submitmoreopts").on("click", function(e) {
+        e.preventDefault();
+        e.stopPropagation();
+
+        var qrform = $("#qrform");
         var replyto = Number($("#dtid").val());
         var pid = Number($("#parenttalkid").val());
         var basepath = $("#basepath").val();
 
         if(replyto > 0 && pid > 0) {
-            $("#qrform").attr("action", basepath + "replyto=" + replyto );
+            qrform.attr("action", basepath + "replyto=" + replyto );
         } else {
-            $("#qrform").attr("action", basepath + "mode=reply" );
+            qrform.attr("action", basepath + "mode=reply" );
         }
 
-        $("#qrform").data("stayOnPage", false);
+        qrform.data("stayOnPage", false);
+        qrform.submit();
     });
+
 });
 
 
--- a/htdocs/js/jquery.quickreply.js
+++ b/htdocs/js/jquery.quickreply.js
@@ -234,6 +234,7 @@ jQuery(function($) {
         var replyto = Number($("#dtid").val());
         var pid = Number($("#parenttalkid").val());
         var basepath = $("#basepath").val();
+        var isSameDomain = (new URL(basepath)).hostname === document.location.hostname;
 
         if(replyto > 0 && pid > 0) {
             qrform.attr("action", basepath + "replyto=" + replyto );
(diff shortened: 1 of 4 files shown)


rahaeli:
I scrutinized the screenshot and didn't find the more options until you told me where it was, I think it's too hidden up there. (My eye skims straight over that block unless I'm setting an icon, which I rarely do.) Add a duplicate button?

Aside from that, this looks great.

nfagerlund:
@rahaeli Oof, I'm _really_ trying to avoid duplicated controls here... how about if it stayed in the upper spot but looked like a button instead of a link? 

image

-----

There's two reasons I'm trying to make more-options-on-top happen:

- The bottom row gets way easier to use if it only has "post" and "preview."
- The actual options you get _more_ of are mostly identity-related, so it seems good to expose them near the spot that shows who you currently are (i.e. the username and icon preview). Here's the full list of what the talkform adds:
    - Switch user or go anon (identity)
    - Make it an admin post (identity)
    - Use a subjecticon (deprecated)
    - Spellcheck (deprecated)
    - Don't autoformat (will mutate into markup-type chooser and be available in all forms)   
    - Unscreen parent (this is mostly just here in case javascript is disabled, right? Otherwise the ajax unscreen button seems easier.) 

So that's two things that are leaving, one thing that really should be everywhere, one oddball... and then the two main items, which are very relevant to the UI that tells you who you're logged in as (i.e. the username indicator and the icon preview).

zorkian:
Given the discussion, waiting on @rahaeli before merging.

I don't use the button, so am not too plussed about where it goes. I like the rationale of keeping related things together tho, and agree w/the "identity up top" stuff.

rahaeli:
Yeah, the "identity on top" argument is a pretty good one. I am primed to look for it now so it's not a perfect test, but I think changing it to button rather than link is noticeable enough -- I saw it a lot easier that way, at least!

nfagerlund:
All right! 🙌🏼 OK then, I updated the PR with a revised version and also pushed it to the dev server. Here's some real screenshots (although, I guess they're indistinguishable from the mockups upthread):

Screenshot_2019-10-07 sheworks4themoney just adding something to reply to
Screenshot_2019-10-07 sheworks4themoney Reading
Screenshot_2019-10-07 betagal Reading

And of course, no design survives first contact with the userbase; might well need to re-evaluate based on how people react to it. We'll find out!

rahaeli:
\o/

PR #2584, opened 2019-10-15
Experimental fix for RTE bustage on Chrome

nfagerlund:
At some point in the past year, Chrome changed something about how measurements
work within a hell-thicket of deeply and multiply nested tables and iframes, and
it broke the old RTE. This commit replaces some of the compounded height: 100%s
with some simple vertical flex, and tries to avoid touching any of the rest of
this rickety markup.


nfagerlund:
I tested this on IE11... and it turns out we don’t even allow rte on IE11 right now. So, this is probably fine to just merge. Please poke at it on my hack if you want to check!

PR #2586, opened 2019-10-20
Enable Foundation in S2 pages (Aug19 megamix 5: risky core)

### LINKED ISSUE #2570: S2 journal-styled pages call res_includes super early
Need some more investigation of this, but the upshot of it is that, when building a journal-styled S2 page (distinct from the special `siteviews` style), there's a certain point (when?) beyond which `need_res` is useless, because res_includes was already called. This affects:

- Template components (like new-new icon browser) that try to load JS or CSS with need_res.
- Captchas that try to load libraries with need_res (like textcaptcha when in the `'foundation'` resource group -- that behavior was written with the expectation that it would only ever occur on site skin pages, because no one had considered bringing foundation stuff into journal styles). 
- Probably other stuff I'm not aware of yet.

nfagerlund:
This is a subset of #2574.
Fixes #2570 (as best as it looks like we can, without cutting a core3)
Functional basis for #2427
Functional basis for #2411

These commits are:

- Front-end code, but it's Perl and S2 front-end, not CSS/JS/markup.
- **Sorta high risk.**
    - Yeah, this doesn't mess with the database or core logic, but it does futz with the weirder corners of a code path that basically EVERY journal page MUST pass through, and it uses some subsystems that seemed to be poorly understood when I was asking questions about them.
    - The functional changes are all behind a new beta flag. So the most critical thing is to make sure nothing new and bad can possibly happen when the beta is OFF. I feel pretty confident about that. 

Bits I want advice on:

- What would you like on that beta feature text?
- Does the implementation of the beta switch match what you want? I followed what the `updatepage` beta does, which is different from what the instructions on the wiki say.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -933,6 +933,9 @@ sub create_qr_div {
 
             current_icon_kw => $userpic_kw,
             current_icon    => LJ::Userpic->new_from_keyword( $remote, $userpic_kw ),
+            foundation_beta => LJ::BetaFeatures->user_in_beta( $remote => "s2foundation" )
+                && $LJ::ACTIVE_RES_GROUP
+                && $LJ::ACTIVE_RES_GROUP eq "foundation",
 
             remote => {
                 ljuser => $remote->ljuser_display,
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -4569,7 +4569,7 @@ sub statusvis_message_js {
     $statusvis_full = "memorial" if $u->is_memorial;
     $statusvis_full = "readonly" if $u->is_readonly;
 
-    LJ::need_res("js/statusvis_message.js");
+    LJ::need_res( { group => 'all' }, "js/statusvis_message.js" );
     return
           "<script>Site.StatusvisMessage=\""
         . LJ::Lang::ml("statusvis_message.$statusvis_full")
--- a/views/beta.tt.text
+++ b/views/beta.tt.text
@@ -19,6 +19,22 @@
 
 .betafeature.updatepage.title=New Create Entries Page
 
+.betafeature.s2foundation.cantadd=Sorry, your account is not eligible to test this feature.
+
+.betafeature.s2foundation.off<<
+<p>Activate several updated components on journal pages, including a new version of the icon browser for paid users and a mobile-friendly cleanup of the site-styled comment pages. The underlying components are already well-tested, but introducing them into a new environment might turn up some bugs.</p>
+
+<p>TODO: post a link for reporting bugs.</p>
+.
+
+.betafeature.s2foundation.on<<
+<p>You are currently testing several updated components on journal pages, including a new version of the icon browser for paid users and a mobile-friendly cleanup of the site-styled comment pages. The underlying components are already well-tested, but introducing them into a new environment might turn up some bugs.</p>
+
+<p>TODO: post a link for reporting bugs.</p>
+.
+
+.betafeature.s2foundation.title=Updated journal page components
+
 .nofeatures=There are no features currently available for beta testing.
 
 .staytuned.newscomm=Stay tuned to [[news]] for upcoming testing opportunities. Thank you.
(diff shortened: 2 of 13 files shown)


nfagerlund:
If it's relevant to reviewers: this PR specifically is why I wrote [that big comment over in Web.pm.](https://github.com/dreamwidth/dw-free/blob/master/cgi-bin/LJ/Web.pm#L2803)

kareila:
In my experience, the most reliable (and tedious) way to update strings on live is to change the codename of the string to something new, by adding '2' at the end or whatever. If it's important to keep the codename the same, you'll need to coordinate the rollout with someone who is able to edit the live site strings (which I think is just @rahaeli at this point).

nfagerlund:
Hmm, the unfortunate thing about beta features is that those string IDs are firmly tied to the ID of the feature itself. `.betafeature.${handler.key}.title` and the like; there's no indirection. So, it's probably best not to change the beta feature ID everywhere just to update the text, if it's an update that we KNOW we'll have to do immediately. 

How about this, instead:

- Make a dw-beta post to coordinate feedback, but it basically just says "Hey this is coming soon; stay tuned, and we'll make another post here when it's live." (Or I guess we could just backdate it too, right? That'll keep it off people's reading pages?)
- Link to that post in the beta text. 
- When we code push, make another post in -beta with comments disabled, and link to the main discussion post. 

? 

Also, does the current text match what y'all want on the beta page, other than the missing link?

> ### Updated journal page components
> 
> Activate several updated components on journal pages, including a new version of the icon browser for paid users and a mobile-friendly cleanup of the site-styled comment pages. The underlying components are already well-tested, but introducing them into a new environment might turn up some bugs.

azurelunatic:
IIRC, community entries cannot be suppressed from the reading page.

cgi-bin/LJ/Web.pm
```diff
@@ -4569,7 +4572,7 @@ sub statusvis_message_js {
     $statusvis_full = "memorial" if $u->is_memorial;
     $statusvis_full = "readonly" if $u->is_readonly;
 
-    LJ::need_res("js/statusvis_message.js");
+    LJ::need_res( { group => 'all' }, "js/statusvis_message.js" );
```
nfagerlund: Wait, I just realized I should call this out for special attention — right now I believe this memorial/readonly status message thing is actually broken on prod, because it gets added to the `default` (read: "legacy") resource group and there's no way for an S2 page to actually be in that group; they're locked to `jquery`. Soooo this would fix it, but it only just now occurred to me to wonder if this is another of those The Ring "you weren't supposed to _help_ her" situations.

kareila: AFAIK no one had realized it was broken. Memorial journals are rare (we get one or two requests for this per year) and we don't use readonly for any situations that I'm aware of.

nfagerlund: Wait.... actually, it's still broken even if I reenable this. That JS file calls `LiveJournal.register_hook()` instead of `$(document).ready()`, and the `livejournal.js` file that declares that function is never loaded in the jquery or foundation resource groups. 

Easy enough to un-break it for real, now that I understand what's happening. Do we wanna?

zorkian: I'm ambivalent about fixing this. As long as the profile can show memorial status, which should be the case, I'm OK if a journal can't know? Readonly we don't use.

That said it _attempts_ to do it, so I'd prefer we either remove the attempting or we fix it. Better to be clear one way or another than just broken.

views/beta.tt.text
```diff
@@ -19,6 +19,22 @@
 
 .betafeature.updatepage.title=New Create Entries Page
 
+.betafeature.s2foundation.cantadd=Sorry, your account is not eligible to test this feature.
+
+.betafeature.s2foundation.off<<
+<p>Activate several updated components on journal pages, including a new version of the icon browser for paid users and a mobile-friendly cleanup of the site-styled comment pages. The underlying components are already well-tested, but introducing them into a new environment might turn up some bugs.</p>
+
+<p>TODO: post a link for reporting bugs.</p>
```
zorkian: For now, the placeholder can just be "report bugs in dw_beta" and if people make new posts, or comment on an existing one, that's probably fine. I don't imagine it'll be high traffic?

zorkian:
I'm going to merge this but _won't_ deploy until we update those strings. Figured it'd be nicer to get the big thing landed so we have less outstanding code?

Issue #2587, opened 2019-10-21
The Squash, Part Two (expand content images w/ JS)


nfagerlund:
In a bunch of styles, I added some CSS to shrink user-provided images (while preserving aspect ratio), so they don't overflow their container (or the viewport) too badly. 

I think that's been mostly a win, but it's not a universal win; sometimes you want to see the whole image and don't mind if it blows out the page, and sometimes the image shouldn't be shrunk in the first place. (The latter is mostly just vertical comic strips like Questionable Content, which run afoul of the default behavior of "a portrait-orientation image should fit on one screen" -- if it's split into panels, it's actually a bunch of images anyway, so that doesn't make sense.)

We've been sharing some CSS hacks to work around that, but I think we can do better by adding a little JS-powered "expand/shrink" widget next to every image in user-provided content. That's the sort of fix that's impossible for users to provide in their own journals, but trivial to do sitewide. 

### Design problems to solve:

- Definitely need to exclude userhead icons, poll bar graphs, and a variety of other things that are really more like _UI elements_ than _figures._ This can also include things like stitched-together frame images in posts where someone went hogwild with decorative HTML. (At least post-content/comment-content areas are well-marked, so it's easy to not mess with the rest of the page.) 
- Ideally, only images that _actually shrank_ should get the expand/shrink button. But how to detect that? 
    - Most users don't provide height/width attributes when posting images, so info about the actual image dimensions isn't available at the time of DOMContentLoaded. 
    - When the dimensions of the window change, that can shrink new images or unshrink ones that were shrunk. 

So, a naïve implementation is pretty easy, but I want to think about this a bit more to see if we can get something that feels simple but avoids wacky side-effects.

-----

Being selective about what we shrink is also an interesting idea (e.g. detecting really long skinny pics), but IDK if it's feasible yet and it wouldn't solve the entire issue anyway, so never mind that for now.

nfagerlund:
A good start is to compare HTMLImageElement.naturalWidth / naturalHeight with the standard width/height properties. (Can't do that until the `load` event, of course.) That leaves the question of how and where to handle window size changes.

nfagerlund:
If [resize observer](https://drafts.csswg.org/resize-observer/) was a thing yet, I’d have it made in the shade, but no.

nfagerlund:
Hmm, also remember that new images can be added via cut tag or expand comment.

nfagerlund:
[...]

PR #2595, opened 2019-11-27
[#2587] Add click-to-zoom for images in entry/comment content

nfagerlund:
Previously, we added CSS that limits image sizes, to avoid reading page blowouts
when people post large images. This works as expected most of the time, but:

- It sometimes shrinks images it shouldn't, like vertical comic strips.
- Sometimes you just want to see the full size and don't mind a temporary blowout.

This commit handles those cases by adding a click-to-zoom behavior for images in
entry/comment contents. On desktop, a special cursor indicates when you can use
it. The new behavior only targets images that aren't inside links (since those
already have a click behavior), and has a special exemption for poll bar graphs.

This also moves the shrink CSS into a global fragment that gets pulled in with
need_res, so we don't need to maintain it in every layout.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -208,6 +208,9 @@ sub make_journal {
             stc/jquery/jquery.ui.dialog.css
             stc/jquery/jquery.ui.theme.smoothness.css
 
+            js/jquery.imageshrink.js
+            stc/css/components/imageshrink.css
+
             js/jquery.poll.js
             js/journals/jquery.tag-nav.js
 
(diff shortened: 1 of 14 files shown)


nfagerlund:
There's a live demo of this code at http://sheworks4themoney.rr-thrice.hack.dreamwidth.net/2802.html.

Can anyone think of other special cases that should be exempt from displaying zoom cursors on hover?

zorkian:
Well it works for me ... but I saw there was some ongoing discussion/you were asking for feedback, so I'm not merging yet?

nfagerlund:
I think this is probably good to go? The only wart I'm really aware of right now is that it also displays a zoom cursor on images that haven't actually been shrunk, and in that case click-to-zoom won't actually do anything. I don't yet see a great way to get around that, though.

PR #2597, opened 2019-11-28
Aug19 megamix 6: safe, complex frontend

### LINKED ISSUE #2411: Icon browser / `iconselector` pop-up is useless on mobile

### LINKED ISSUE #2427: Make S2 site scheme pages responsive/Foundation-ized

### LINKED ISSUE #2480: Redesign the talkform
Issue #2437 is all over but the screaming, and that means it's time to decide what to do about the talkform. I think I need signoff from @zorkian and @rahaeli before moving on any of this, but I also _want_ feedback from basically everyone.

I have four questions, some of them larger than others.

- ~Can I kill spellcheck?~ Yes, #2481 
- ~Can I kill or drastically mutate subjecticons?~ Almost definitely, #2483 with a hard dependency on #2482 
- How should the identity controls work? (TBD)
- ~Editing affordances?~ Yes — support Markdown, let users disable Markdown (and remember their choice), and make editor buttons emit MD or HTML tags depending on mode. #2573

## 1. Can I kill spellcheck?

Both the talkform and the less-quick reply have a spellcheck checkbox. **Does Dreamwidth need this in 2019?** Even the worst desktop browsers have had spellcheck in textareas for the last ten years at least, and mobile platforms all have spellcheck in their keyboards. What does preview-spellcheck actually add?

## 2. Can I kill or drastically mutate subjecticons?

**Nota bene: I am not talking about userpics here. Userpics obviously stay.**

When it forked from LJ, DW inherited this weird set of about eighteen incredibly shitty icons that look like this: the skull subjecticon Super awesome PHP-BBS-circa-y2k stylings on these. They are called "subjecticons," not to be confused with Decepticons.

- You can choose one per comment (_not_ top-level post), and it'll appear above or beside the subject header (presentation is delegated to journal CSS).
- They're hardcoded. There's no abstraction for working with a DW site's collection of subjections.
- They're ONLY available from the full-size talkform; impossible to use them in the quickreplies.
- They have a dedicated button in the full talkform.
- Every one of them (with the sole exception of hitchhiker boi) has a one-to-one replacement in the standard set of emoji circa 2015. 🙂😉😊😮🙁😠✅⭐️✉️👀☠️🤢☢️😎💡👎👍
- Raise your hand if you actually use these on the regular.
- Raise your hand if you _knew these existed._

**Zeroth question: Please raise all affected peoples' hands for them.**

Subjecticons are stored in the database as a distinct property of each comment, I think as an ID like `sm09`. Would someone with access to the prod databases be willing to do some queries and calculate what percentage of the last three years worth of comments include a subjecticon?

**First question: Can these please just _go?_**

I know deleting features is socially hazardous, but please answer from the dark silence of your soul instead of the reflex flipout of your user support persona. (The reflex flipout of my own user support persona commiserates with you, but still.)

I think no one's been motivated to actually make a decision about this feature because it's been relegated to the Old Talkform all decade, and everyone knows the Old Talkform is shitty and not to be used by normal people. But now that the talkform is maintainable, it's time to end that way of thinking and bring the comment forms together! _If subjecticons should stay in the talkform, they should also go into both quickreplies._ Do you want that? If not, I think it's unreasonable to want them to stay in the talkform.

**Second question: _If not,_ can we just replace them with Big First Subject Emoji?**

(I submitted this as a `dw_suggestions` post several weeks back, but `dw_suggestions` might be sleeping.)

When I raised the possibility of killing subjecticons before, someone said they were too beloved by _someone,_ possibly a subset of RPers. If that's really & truly the case, would those people be satisfied with an alternate interface for showing a little picture above the comment subject? Specifically:

If the first character of the subject is an emoji, AND the second character is NOT an emoji (to preserve the integrity of important sequences like 😝😜😛🤪😛😋😛😝😜😋😝):

- The first character will be removed from the subject and displayed outside the subject in a special element.
- By convention, this special element will be styled by S2 layouts as a big-ass emoji, font-size 32px or thereabouts.
- That big-ass emoji will be displayed using:
    - Preferably, the system's default emoji font.
    - If that's untenable for some external reason I'm not aware of, a standard freely available emoji set like Twemoji.
- The removal and big-ass re-display of the first emoji will be solely part of the display logic:
    - In the event that a user edits their comment, the emoji will still be present in the edit form's subject field as the first character.
    - In the event that some user replies to that comment, the default reply subject will be `Re: <big-ass emoji> <remainder of subject>`.
    - In the database, the subject will be persisted as originally written.

When answering this question, please combine various constraints — for example, while there might be theoretical technical situations where "emoji don't work," _do those situations apply in real life to a meaningful subset of the specific users who currently use subjecticons heavily enough that we can't just remove them?_

## 3. How should the identity controls work?

Extracting the talkform taught me many things. `</glados-voice>` Foremost among them was that it used to be _really hard_ to express what you wanted the "from" controls to do.

Now it's less so, so: what should they do and how should they look?

I have two semi-concrete proposals:

### Drop-down with optional appendix

IMG_9477

Compress the "from" controls into a single drop-down that starts with the default option for the current conditions selected. If the currently selected option needs extra info (logged-out DW user, logged-out OpenID user), show the sub-form below; if not, not.

### Radio buttons with "other" and "current" compressed

IMG_2725

Much like today, use radio buttons to expose the full range of "from" options. Unlike today, rearrange which things get conflated:

- The third field, "currently logged-in user," accommodates either a site or openID user. You can only be logged in as one at a time, so this just makes more sense.
    - The third field is absent if no one's logged in. Two fields only.
- The second field, "other account" — its label transforms into a drop-down once selected, and exposes the relevant child form fields below. It defaults to "site user" with username/password/login fields, but you can switch it to "openid" and get the URL/login fields instead.

Both of these options leave room for future expansion of login options, not that I know of anything specific that's actually planned. Facebook and Twitter are sure looking like dead-ends these days. Maybe we'll want "log in with Apple," lol.

**No-JS fallback:** (because it's good for your soul, even if it's not generally realistic) The dropdown acts like normal, but _all_ (both) appendix forms display below it. Either we use backend form validation to reject submissions that have wrong fields filled out, or (better, IMO) we just ignore fields that aren't relevant. Can include no-JS instructions that get display: none'd otherwise to help resolve the confusion.

## 4. Editing affordances?

I know someone's working on a Markdown-based editor interface for the beta new-post page, but I don't know how close it is to being usable, and I don't know whether they've considered the commenting forms. **What's the status on the "RTE," and can I add some stupid markup buttons to comments in the meantime?**

Ideally I want comments to get Markdown powers too, and I want that to be controlled with a UI checkbox instead of a !Secret !Word. And then I want the comment forms to get UI buttons for basic Markdown constructs like `<strong>` / `<em>` / `<ul>` / `<ol>` / `<a>` / ``, which should all be shared with the new-post page.

But if that's more than two months or so in the future: lack of Markdown makes comments bad, and I'd like to ameliorate that with some buttons and keyboard shortcuts for the most basic markup options. In particular:

- "b" button/cmd-B: `<strong>`
- "i" button/cmd-I: `<em>`
- "quote" button: `<q>/<blockquote>` with selection
- "link" button: `<a href="URL">TEXT</a>`
- "@" button: `<user name="SELECTION">`
- "`<>`" button: `<raw-code>`

Most of these should transfer pretty well to a post-Markdown regime, yeah? Is there any reason not to add them in today's HTML world?

These buttons should appear in the talkform AND both quickreplies.

### LINKED ISSUE #2542: Talkform: Tune up the no-js flow
The talkform could stand to be a bit easier to get through if JavaScript is disabled. 

- I think you have to manually blank out the username in the “other user” fields if you’re logged in and commenting anon, bc the js won’t  do it on submit. Why _did_ that always default to the current user, anyway? Was it because the only way to show the form was to disappear the logged-in user option, so that was the escape hatch for letting you go through with your comment? IMO it doesn’t serve a useful purpose today, but maybe I’m missing something. 
- Might be another rough edge or two.

nfagerlund:
This is a subset of #2574.
Fixes #2411
Fixes #2427
Fixes #2480
Fixes #2542 

These commits are:

- Front-end only.
- Very safe. 
    - They're introducing or rewriting a lot of complex CSS and JS, but those changes only affect users who have opted into a beta. (Mostly the new `s2foundation` beta, but also the `updatepage` beta because of the icon browser.) 
    - There's also a commit that fixes some annoying talkform behavior when JS is disabled. (Those problems are currently live on prod, but are rare enough that I don't think we've heard much about them.)

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1627,8 +1627,7 @@ sub talkform {
             editreason  => $comment ? $comment->edit_reason : '',
             oidurl      => $form->{oidurl},
             oiddo_login => $form->{oiddo_login},
-            user        => $form->{userpost}
-                || $form->{cookieuser},
+            user        => $form->{userpost},
             do_login    => $form->{do_login},
             body        => $form->{body},
             subject     => $basesubject,
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1838,7 +1838,6 @@ sub init_iconbrowser_js {
         'js/foundation/foundation/foundation.js',
         'js/foundation/foundation/foundation.reveal.js',
         'js/components/jquery.icon-browser.js',
-        'stc/css/components/block-grid.css',
         'stc/css/components/icon-browser.css',
     );
 
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1712,7 +1712,6 @@ sub talkform {
         },
 
         help_icon               => sub { LJ::help_icon_html(@_) },
-        ejs                     => sub { return LJ::ejs(@_) },
         print_subjecticon_by_id => sub { return LJ::Talk::print_subjecticon_by_id(@_) },
     };
 
@@ -1907,6 +1906,7 @@ sub init_s2journal_js {
     LJ::need_res(
         { group => "all" }, qw(
             js/jquery.replyforms.js
+            stc/css/components/quick-reply.css
             )
     );
 
@@ -1918,7 +1918,6 @@ sub init_s2journal_js {
             js/jquery/jquery.ui.widget.js
             js/jquery.quickreply.js
             js/jquery.threadexpander.js
-            stc/css/components/quick-reply.css
             )
     ) unless $opts{noqr};
 
--- a/htdocs/js/components/jquery.icon-browser.js
+++ b/htdocs/js/components/jquery.icon-browser.js
@@ -2,10 +2,12 @@
 
 function IconBrowser($el, options) {
     var iconBrowser = this;
+    var modalSelector = "#" + options.modalId;
+    var scrollPositionDogear;
 
     $.extend(iconBrowser, {
         element: $el,
-        modal: $("#" + options.modalId),
+        modal: $(modalSelector),
         modalId: options.modalId
     });
 
@@ -14,12 +16,34 @@ function IconBrowser($el, options) {
     new Options(this.modal, options.preferences);
 
     $(document)
-        .on('open.fndtn.reveal', "#" + options.modalId, function(e) {
+        .on('open.fndtn.reveal', modalSelector, function(e) {
             // hackety hack -- being triggered on both 'open' and 'open.fndtn.reveal'; just want one
             if (e.namespace === "") return;
 
+            // If the page scrolled sideways, don't put the modal way out in left field.
+            iconBrowser.modal.css('left', window.scrollX);
+
             iconBrowser.loadIcons();
             iconBrowser.registerListeners();
+        })
+        // Save and restore the scroll position when opening and closing the
+        // modal. This is crucial on mobile if you have dozens of icons, because
+        // otherwise it'll ditch you miles out into the comment thread, as you
+        // wonder where you left your reply form and whether you have enough
+        // water to survive the walk back to the gas station.
+        .on('opened.fndtn.reveal', modalSelector, function(e) {
+            // hackety hack -- being triggered on both 'opened' and 'opened.fndtn.reveal'; just want one
+            if (e.namespace === "") return;
+
+            scrollPositionDogear = $(window).scrollTop();
+        })
+        .on('closed.fndtn.reveal', modalSelector, function(e) {
+            // hackety hack -- being triggered on both 'closed' and 'closed.fndtn.reveal'; just want one
+            if (e.namespace === "") return;
+
+            if ( Math.abs( $(window).scrollTop() - scrollPositionDogear ) > 500 ) {
+                $(window).scrollTop(scrollPositionDogear);
+            }
         });
 }
 
@@ -69,16 +93,16 @@ IconBrowser.prototype = {
                             icon.height,
                             width: icon.width,
                             "class": "th" } )
-                        .wrap("<div class='icon-image'></div>").parent();
+                        .wrap("<a>").parent()
+                        .wrap("<div class='icon-browser-icon-image'></div>").parent();
                     var $keywords = "";
                     if ( icon.keywords ) {
                         $keywords = $("<div class='keywords'></div>");
-                        var last = icon.keywords.length - 1;
 
                         $.each(icon.keywords, function(i, kw) {
                             iconBrowser.kwToIcon[kw] = idstring;
                             $keywords
-                                .append( $("<a href='#' class='keyword radius' data-kw='" + kw + "'></a>").text(kw) )
+                                .append( $("<a class='keyword radius' data-kw='" + kw + "'></a>").text(kw) )
                                 .append(document.createTextNode(" "));
 
                         });
@@ -143,32 +167,43 @@ IconBrowser.prototype = {
         }
     },
     selectByClick: function(e) {
-        // this may be on either the icon or the keyword
-        var container = $(e.target).closest("li");
-        var keyword = $(e.target).closest("a.keyword");
-
-        this.doSelect(container, keyword.length > 0 ? keyword.text() : null, true);
-
         e.stopPropagation();
         e.preventDefault();
+
+        // If this is the second click, treat it as confirmation:
+        if ( $(e.target).hasClass("active") ) {
+            this.updateOwner.call(this, e);
+        } else {
+            // this may be on either the icon or the keyword
+            var container = $(e.target).closest("li");
+            var keyword = $(e.target).closest("a.keyword");
+
+            this.doSelect(container, keyword.length > 0 ? keyword.text() : null, true);
+
+            // If they chose a keyword, treat it as confirmation:
+            if (keyword.length > 0) {
+                this.updateOwner.call(this, e);
+            }
+        }
     },
     selectByDoubleClick: function(e) {
         this.selectByClick.call(this, e);
         this.updateOwner.call(this, e);
     },
     selectByKeywordMenuClick: function(e) {
+        e.stopPropagation();
+        e.preventDefault();
+
         var keyword = $(e.target).text();
         var id = this.kwToIcon[keyword];
         if ( id ) {
             this.doSelect($("#" + id), keyword, false);
+            // If they chose a keyword, treat it as confirmation:
+            this.updateOwner.call(this, e);
         }
-
-        e.stopPropagation();
-        e.preventDefault();
     },
     selectByKeywordMenuDoubleClick: function(e) {
         this.selectByKeywordMenuClick(e);
-        this.updateOwner.call(this, e);
     },
     initializeKeyword: function() {
         var keyword = this.element.val();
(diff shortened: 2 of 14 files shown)


zorkian:
"Safe" I mean how safe is anything that talks about IE11 and contains comments like getting lost in the desert and dying of dehydration and having to spend the night in your camel's abdomen...

But, still, let's roll. Merged.

PR #2599, opened 2019-11-29
Aug19 megamix 8: risky noops

nfagerlund:
Wow, we made it! 

This is a subset of #2574. Part 7 got subsumed into part 6.

This PR refactors some duplicated and awkward code. It also fixes one bug, where the beta entry page flips out and disables a bunch of stuff if you have no default icon. Other than that, it shouldn't cause any visible behavior changes.

**Risk: Middlin'.** This messes with some higher-level methods that fetch userpics from the database, and it's a twisty area of code with a bunch of duplications and oddities. There's a possibility I misunderstood something in here, but I think I got it.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -772,44 +772,41 @@ sub icon_keyword_menu {
 
     my @icons = grep { !( $_->inactive || $_->expunged ) } LJ::Userpic->load_user_userpics($user);
 
-    if (@icons) {
+    return () unless (@icons);
 
-        # Get a sorted array of { keyword => "...", userpic => userpic_object } hashrefs:
-        @icons = LJ::Userpic->separate_keywords( \@icons );
+    # Get a sorted array of { keyword => "...", userpic => userpic_object } hashrefs:
+    @icons = LJ::Userpic->separate_keywords( \@icons );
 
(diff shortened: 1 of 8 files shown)


cgi-bin/LJ/Web.pm
```diff
@@ -752,53 +752,64 @@ sub check_referer {
...
-            my $defaultpicurl = $res{defaultpicurl}
-                || ( $LJ::IMGPREFIX . $LJ::Img::img{nouserpic_sitescheme}->{src} );
-            @pics = (
-                {
-                    value => "",
-                    text  => LJ::Lang::ml('/talkpost.bml.opt.defpic'),
-                    data  => { url => $defaultpicurl }
+    my @icons = grep { !( $_->inactive || $_->expunged ) } LJ::Userpic->load_user_userpics($user);
+
+    if (@icons) {
```
zorkian: nit: Slightly cleaner (less indenting) if this is,

```perl
my @icons = ...;
return () unless @icons;

@icons = LJ::Userpic->...
return @icons;
```

nfagerlund: Yeah, ok! 

If I'm doing that, maybe I should also just return that `(default-icon, map-expression)` list directly, instead of re-assigning `@icons` right before returning it.

zorkian:
Nice thank you! Looks good to me.

PR #2600, opened 2019-12-03
TaskQueue: to replace TheSchwartz

zorkian:
I accidentally went on a tear today and wrote a replacement for TheSchwartz. Well, that's a bold claim ... this is a first draft of a potential replacement.

In essence, the workflow is to define a DW::Task subclass and then you use DW::TaskQueue to send/receive those objects. You can create a worker pump pretty easily (see the bin/worker jobs) and the rest should be handled automatically.

This doesn't have quite the same semantics as TheSchwartz, in particular we don't really support the super long grabbed_fors, the deduplication, and some other niceties. For _now_, my plan is just to see if I can replace the super high-impact ESN workload which doesn't require those niceties.

(Except possibly in the case that it takes us a long time to execute one of the stages, i.e., >30 seconds... but we can chase the message visibility timeout if we need to in order to give us more time.)

I uh, I don't know who to tag for eyeballs, so I'm just going to open this PR and see.


rshatch:
I'll have a deeper look later, but on the timeout/dedupe stuff - FIFO queues can handle the dedupe stuff. You'll have to supply a message group ID for each message, which you will want to be different for each message (messages with the same group ID are processed one at a time) - uuid4 is sufficient. For visibility timeout, it's probably better to err on the side of a longer timeout, because if a job takes longer than the timeout, its reciept handle will have expired, and it can end up getting processed over and over forever.

PR #2603, opened 2019-12-05
WIP - /customize conversion

rshatch:


nfagerlund:
I'm going to try and just update one comment with issues I notice! 

- On the journal style picker, the pagination links navigate to the raw RPC json. 
- Looks like the "Apply Layout" buttons will reliably break things if you click them on one of the options pages, rather than the style picker page. After doing that, you'll get `Invalid user. at /dreamhack/home/8111-rr-thrice/dw/cgi-bin/DW/Controller/Customize/Customize.pm line 413.` errors when trying to interact with them. 
- Lots of error logging. Notable instances: 
    - `Reference found where even-sized list expected at /dreamhack/home/8111-rr-thrice/dw/cgi-bin/DW/Controller/Customize/Customize.pm line 439.`
    - Empty error messages, where it lists an " at such and such" annotation but without saying what the beef was.

Issue #2601, opened 2019-12-04
Broken Markdown processing in html cleaner


nfagerlund:
PR #2540 broke Markdown pretty thoroughly. It hasn't been deployed to prod.

Here's what I've learned so far: 

- ~[This bit right here](https://github.com/dreamwidth/dw-free/pull/2540/files#diff-a84240ca7cbff9cf6b58a0eee16ba5b4R1231-R1238) isn't tenable -- dropping into HTML is part of the Markdown language As She Is Played, so disallowing it breaks all existing user expectations about how it works and a lot of existing text.~

    Correction!!! I misunderstood how this breakage was working. Intended behavior is fine. As described in the PR text:

    > While using Markdown, you can also "drop to HTML" by opening any tag. This will suspend Markdown processing and @-user mentions until you close the tag. This enables you to drop in a textarea or whatever and not have to stress about it.

    That's part of the md spec, and is totally fine. Sorry for the false alarm! I interpreted the code as "any HTML locks you out of Markdown mode," due to the next bug:
- MD processing can't recognize "empty" (singleton) HTML tags, unless they're self-closed (XHTML style). So throwing in an `<img>` tag will break the rest of the post, unless you write it like `<img />`. We should at least make it so it can recognize the HTML5 forms of the most common empty tags. (img, br, etc.)
- Same for DW special tags, which are mostly the same issue but more inconsistent. For example, `<user>` tags can be self-closed, but poll placeholders can't; you need to add a full dummy closing tag like `<poll-132></poll-132>`. Cut tags are an even specialer case, because they render differently on last-N pages and entry pages -- using a cut tag around some markdown will make a post that works on its entry page but breaks on the flist. (Actually, pretty sure cut tags need to be a special exception to "no markdown-in-html-in-markdown".)
- There's also bugs when an HTML tag occurs inside a Markdown token. So far I've observed this with images inside links (`[<img src="..." />](url)`), but it might also break with tags inside blockquotes or something; still need to test. 
- Markdown image tags are broken. That syntax looks like this: 

    ```markdown
    ![alt text](http://image-url.example)
    ```

    This might be a problem with `HTML::TokeParser`'s configuration, but I can't tell yet.
- Actually, MD seems to be broken entirely on last-N pages. Switch to canary and check out [this entry page](https://roadrunnertwice.dreamwidth.org/577566.html) vs. how it appears [on the tag page](https://roadrunnertwice.dreamwidth.org/tag/didread). 
[...]

PR #2604, opened 2019-12-05
Markdown/HTMLCleaner fixes

nfagerlund:
This fixes most of the problems from #2601.

- Switches MD processing back to document-at-once mode. Markdown might look simple, but it's such a complex language that we can't expect an HTML tokenizer to reliably give us a valid chunk of Markdown. 
- Moves the `!markdown` check into `clean_event`, so it can hit everything that needs it (but not hit comments). 
- Preserves the desired @-mentions behavior as codified in the tests (with one exception, where the codified behavior didn't quite make sense — see `qq{<a href="$url">link from \@system</a>}`). 
- Handles a bit of a sideline question about how user tags within links should work. (Link-in-link is illegal.) 
- Adds some more contexts where @-mentions aren't processed. (code spans and pre blocks.)
- Makes reading/recent and cuttag-expander go through the same formatting route as the entry page does -- instead of (manually get raw, manually call clean_event), it's (call centralized thing that gets raw and cleans it). ...I think this is harmless deduplication, but you should double-check. 
- Threads editor prop through in a bunch of places that were missing it.
- Catches up some oddball uses of markdown (Sendmail). 

AFAICT, that should handle all of the new crop of Markdown bugs! EXCEPT for the following:

### Not addressed by this PR:

Crossposter's hosed. DW/External/XPostProtocol/LJXMLRPC.pm -- Right now it'll just explode, because it calls a function that got removed. 

That module has a bunch of code copy-pasted from CleanHTML, and it Seems Bad. It _looks_ like the right move would be somewhere in the neighborhood of: 

- Modify `clean()` to take an option that means "turn lj tags into their old-style LJ form and otherwise leave them undigested." Looks like these tags are all nicely isolated in their own conditionals, and you could handle the `@user` case by modifying `user_link_html`. 
- Thread that option through `clean_event`.
- Delete all that code from crossposter and just have it call `clean_event` with the special "transform and retain LJ tags" option.

BUT! IDK enough about the constraints the crossposter is operating under, so someone else should probably do it. 

I guess the quick dirty fix would be to just call Text::Markdown directly and forget about `@mentions`, passing them through to LJ undigested. That'd be a regression, but maybe not a critical one, and it might buy time to work on a less dirty fix if you're in a hurry to ship.

### DIFF
--- a/cgi-bin/DW/Controller/RPC/CutExpander.pm
+++ b/cgi-bin/DW/Controller/RPC/CutExpander.pm
@@ -84,10 +84,6 @@ sub load_cuttext {
     my $journalid = $entry_obj->journalid;
     my $journal   = LJ::load_userid($journalid);
 
-    #load and prepare text of entry
-    my $text = LJ::CleanHTML::quote_html( $entry_obj->event_raw, $get->{nohtml} );
-    LJ::item_toutf8( $journal, \$subject, \$text ) if $entry_obj->props->{unknown8bit};
-
     my $suspend_msg    = $entry_obj && $entry_obj->should_show_suspend_msg_to($remote) ? 1 : 0;
     my $cleanhtml_opts = {
         cuturl              => $entry_obj->url,
@@ -99,7 +95,8 @@ sub load_cuttext {
         cut_retrieve        => $cutid,
     };
 
-    LJ::CleanHTML::clean_event( \$text, $cleanhtml_opts );
+    #load and prepare text of entry
+    my $text = LJ::CleanHTML::quote_html( $entry_obj->event_html($cleanhtml_opts), $get->{nohtml} );
 
     LJ::expand_embedded( $journal, $jitemid, $remote, \$text );
 
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -628,7 +628,14 @@ TOKEN:
                     : exists $attr->{comm} ? $attr->{comm}
                     :                        undef;
 
-                $newdata .= user_link_html( $user, $attr->{site}, $opts );
+                if ( $opencount{'a'} ) {
+                    $opts->{delink_user_links} = 1;
+                    $newdata .= user_link_html( $user, $attr->{site}, $opts );
+                    $opts->{delink_user_links} = 0;
+                }
+                else {
+                    $newdata .= user_link_html( $user, $attr->{site}, $opts );
+                }
             }
             elsif ( $tag eq "lj-raw" ) {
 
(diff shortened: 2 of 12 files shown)


nfagerlund:
Oh hmm, also need to doublecheck behavior with auto-links! Those kind of violate some expectations, because they're adding HTML while we're inside a text node.

nfagerlund:
OK, I think this fixes all of the outstanding issues? or at least all of the ones I'm aware of. Got cut tags and reading/recent working, and also got RSS and Atom feeds working again. 

BUT, I'm concerned about... hold on, I'll just add an inline comment for it.

cgi-bin/DW/Controller/RPC/CutExpander.pm
```diff
@@ -99,7 +95,10 @@ sub load_cuttext {
         cut_retrieve        => $cutid,
     };
 
-    LJ::CleanHTML::clean_event( \$text, $cleanhtml_opts );
+    #load and prepare text of entry
+    my $text = LJ::CleanHTML::quote_html( $entry_obj->event_html($cleanhtml_opts), $get->{nohtml} );
+
+    # LJ::item_toutf8( $journal, \$subject, \$text ) if $entry_obj->props->{unknown8bit};
```
nfagerlund: I'm concerned about this guy, because I really have no idea what's up with it. Is there anything in our db that might be in weird 8-bit encodings? Is this safe to re-enable if we're not using "event_raw"? 

The reason I suspected it's not necessary is because LJ/S2/EntryPage.pm doesn't do any of this.

nfagerlund:
Oh wait, the crossposter is almost definitely hosed. It's calling a method that @zorkian removed previously. 

Wow, that's a lot of duplicated code from HTML cleaner in there... is there a way to unify that?  :|

nfagerlund:
I think I actually caught everything but the crossposter, so I'll go ahead and mark this ready.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -617,7 +628,14 @@ TOKEN:
                     : exists $attr->{comm} ? $attr->{comm}
                     :                        undef;
 
-                $newdata .= user_link_html( $user, $attr->{site}, $opts );
+                if ( $opencount{'a'} ) {
```
zorkian: nit:

Wouldn't it have been semantically the same and cleaner to just do,

```perl
$opts->{delink_user_links} = $opencount{'a'} ? 1 : 0;
$newdata .= user_link_html( $user, $attr->{site}, $opts );
```

Since, in the first path, you're unconditionally setting it -- but not in the second path -- seems like we should just always set it appropriately (and make sure it gets a default initizliation state in opts at the beginning of the sub?).

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1228,36 +1253,44 @@ TOKEN:
...
+                # a few.)
+                if (   $local_content
+                    && !$disable_user_conversion
+                    && !$opencount{'code'}
+                    && !$opencount{'pre'}
+                    && !$opencount{'textarea'}
+                    && !$opencount{'lj-raw'} )
+                {
+                    # Special no-link user tag if we're already inside another link
+                    if ( $opencount{'a'} ) {
```
zorkian: same nit

cgi-bin/LJ/Protocol.pm
```diff
@@ -2024,6 +2027,7 @@ sub editevent {
     }
 
     ## handle meta-data (properties)
+    # FIXME: Hey... I think this just throws the changed props away???? -NF
```
zorkian: Not all props are "user editable" through the protocol, is that what you are referring to?

zorkian:
Change LGTM, going to get them up on canary so people can poke raound

PR #2609, opened 2019-12-20
Nerf an OP `#content ul li` rule in _page-layout-hacks.scss

nfagerlund:
The tags field on the new entry page currently looks like this: 

But it's _supposed_ to look like this: 

The problem is a default list style that uses an ID selector. I was going to just add yet another hack to work around it (there were several already, see diff), but it's gonna keep biting us, so let's just fix it instead.

-----

The fix: `#content` and `[id="content"]` target the same element, but the former can only be overridden by rules that contain at least one `#id` (or by `!important`). The latter can be overridden by any class, so specific components will always defeat it.

### DIFF
--- a/htdocs/scss/components/autocompletewithunknown.scss
+++ b/htdocs/scss/components/autocompletewithunknown.scss
@@ -30,7 +30,6 @@ $half-form-spacing: $form-spacing / 2;
 
     .token.new {
         border-style: dashed;
-        margin-left: 0 !important;
     }
 
     .token .token-remove {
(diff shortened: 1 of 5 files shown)


nfagerlund:
Oh, wait... now that I think of it, I bet selecting the *id attribute* (`[id="content"]` instead of `#content`) would work identically to this and be conceptually cleaner. I’ll try that when I can get some computer time.

nfagerlund:
Also just spotted a typo in one of the hack removals.

nfagerlund:
Yay, I was right. Fixed/updated.

zorkian:
Merry Christmas PR!

PR #2616, opened 2020-01-24
Fix crossposter, maybe

### LINKED ISSUE #2606: Crossposter markdown processing is broken
DW/External/XPostProtocol/LJXMLRPC.pm -- Right now it'll just explode, because it calls a function that got removed in #2540. Luckily this apparently runs in a separate worker process, so @zorkian said he plans to just pin that crosspost worker on an old version of the code for the time being. But it'll need a fix someday! 

### Speculation copied from #2601 / #2604: 

That module has a bunch of code copy-pasted from CleanHTML, and it Seems Bad. It _looks_ like the right move would be somewhere in the neighborhood of: 

- Modify `clean()` to take an option that means "turn special DW tags into their old-style LJ form and otherwise leave them undigested." Looks like these tags are all nicely isolated in their own conditionals, and you could handle `@user` conversion by modifying `user_link_html`. 
- Thread that option through `clean_event`.
- Delete all that code from crossposter and just have it call `clean_event` with the special "transform and retain LJ tags" option.

BUT! IDK enough about the constraints the crossposter is operating under, so someone else should probably do it.

nfagerlund:
Fixes #2606 

We discussed this approach briefly in the discord. Roughly: currently the crossposter has to know things about all of DW's formatting dialects and do Something to whip them into a format that will display as expected on LJ. This PR tries to simplify that a bit:

- Crossposts will now always get sent as `opt_preformatted`, and we'll process them ahead of time to render all of our autoformatting correctly, no matter what the original formatting dialect is. So if you try to edit a crosspost manually on LJ, it's not going to much resemble your original source — we'll have already turned markdown into HTML, or replaced line breaks with `<br>` tags, or what have you. 
- But we're keeping the existing behavior where we preserve cut tags, and the one where we translate `<user name="whoevs" site="DESTINATION.SITE">` into `<lj user="whoevs">` if DESTINATION.SITE matches the crosspost's target. (Works with `@mentions` now!)

More in the commit messages. 

## 🚨 THIS PR HAS NOT BEEN TESTED. 🚨

Because I don't have any good way to field-test the crossposter. I know it doesn't blow up normal site display stuff, but beyond that I'm at the mercy of my kind reviewers here. So please stare hard. 😬

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -579,8 +579,10 @@ TOKEN:
                 if ( $opts->{preserve_lj_tags_for} ) {
                     $opencount{'lj-cut'}++;
                     $newdata .= qq{<lj-cut};
-                    $newdata .= qq{ text="$attr->{'text'}"}
-                        if $attr->{'text'};
+                    if ( $attr->{'text'} ) {
+                        my $etext = $link_text->();
+                        $newdata .= qq{ text="$etext"};
+                    }
                     $newdata .= '>';
                     next TOKEN;
                 }
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -292,6 +292,12 @@ sub clean {
             || $_[0];
     };
 
+    my $usertag_opts = {
+        textonly             => $opts->{textonly} ? 1 : 0,
+        no_ljuser_class      => $opts->{to_external_site} ? 1 : 0,
+        no_link              => 0,
+    };
+
     # if we're retrieving a cut tag, then we want to eat everything
     # until we hit the first cut tag.
     my @cuttag_stack = ();
@@ -300,6 +306,8 @@ sub clean {
 TOKEN:
     while ( my $token = $p->get_token ) {
         my $type = $token->[0];
+
+        $usertag_opts->{no_link} = $opencount{'a'} ? 1 : 0;
 
         if ( $type eq "S" )    # start tag
         {
@@ -625,14 +633,7 @@ TOKEN:
                     : exists $attr->{comm} ? $attr->{comm}
                     :                        undef;
 
-                if ( $opencount{'a'} ) {
-                    $opts->{delink_user_links} = 1;
-                    $newdata .= user_link_html( $user, $attr->{site}, $opts );
-                    $opts->{delink_user_links} = 0;
-                }
-                else {
-                    $newdata .= user_link_html( $user, $attr->{site}, $opts );
-                }
+                $newdata .= user_link_html( $user, $attr->{site}, $usertag_opts );
             }
             elsif ( $tag eq "lj-raw" ) {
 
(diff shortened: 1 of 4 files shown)


cgi-bin/LJ/CleanHTML.pm
```diff
@@ -571,33 +573,47 @@ TOKEN:
...
-                        . $journal . "_"
-                        . $ditemid . "_"
-                        . $cutcount
-                        . "\" aria-live=\"assertive\">";
-                    $newdata .= "</div>";
-                    $newdata .= "</div>" if $tag eq "div";
-
-                    unless ( $opts->{'cutpreview'} ) {
-                        push @eatuntil, $tag;
+                    if ( $opts->{preserve_lj_tags} ) {
```
nfagerlund: Wait. I think I put this in the wrong branch of the enclosing "if". Now that I'm re-reading it, I think `$cut` means "this is displaying on last-N, so remove the cut content." That should never be enabled for crossposter.

nfagerlund: (In the process of fixing this, I rewrote things enough that I could just remove the second-pass parser in the crossposter itself.)

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1900,6 +1886,10 @@ sub convert_user_mentions {
         !mge;
 }
 
+# Keys available in opts hashref:
```
zorkian, line 1915: Kind of makes you want for types or something.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -570,7 +583,15 @@ TOKEN:
                     }
                     return $text;
                 };
-                if ($cut) {
+                if ( $opts->{preserve_lj_tags_for} ) {
+                    $opencount{'lj-cut'}++;
+                    $newdata .= qq{<lj-cut};
+                    $newdata .= qq{ text="$attr->{'text'}"}
```
zorkian, line 582: Does the text attribute need to be HTML stripped or anything? I'm guess maybe, because later uses of `$newdata` are explicitly using `$etext`?

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -532,13 +532,6 @@ TOKEN:
                 $ljcut_div = 0;
             }
 
-            # no cut URL, record the anchor, but then fall through
-            if ( 0 && $ljcut_div && !$cut ) {
```
zorkian, line 526: Over 14 years, I wonder just how much CPU we spent on the Perl interpreter reading this code and deciding to throw it away. And I wonder if, at any point, it bitflipped and ran it...

cgi-bin/LJ/CleanHTML.pm
nfagerlund: Ack, you're right!!! Good catch. I'll add a commit.

cgi-bin/LJ/CleanHTML.pm
nfagerlund: 😑 srsly

nfagerlund:
Okay — made sure to escape the cut text, re-using the same escaping function the other branches use. 👍🏼

nfagerlund:
Oops, that exploded because I forgot how to perl. One sec.

zorkian:
Ty sir!

PR #2619, opened 2020-03-21
Remove MD5 challenge/response login

rshatch:


zorkian:
Because we force SSL now, so this whole "md5 in the browser" is silly?

rshatch:
Yes, and this method is incompatible with hashed passwords.

zorkian:
LGTM, merged!

PR #2621, opened 2020-04-20
Feature: bcrypted passwords

zorkian:
This implements using properly bcrypted passwords with unique salts, per
best practices. This change has a couple of sharp edges:

1) ATOM posting and digest-auth will break. Unclear how impactful this
is or what we should consider doing (if anything) about it.

2) Renames used to verify that the accounts being merged shared a
password. We think that having validated email accounts that match is
good enough, though.

3) Probably some bugs, I've only tested that I can still log in/out,
haven't gone wild testing every flow.

The implementation leverages our dversion system so that users can be on
the 'old style' passwords and we can upgrade them when we are confident
and ready to roll.

### DIFF
--- a/bin/renameuser.pl
+++ b/bin/renameuser.pl
@@ -66,10 +66,6 @@ unless ( $args{force} ) {
         print "   " . $acct[1]->email_raw . "\n";
         exit 1;
     }
-    unless ( $acct[0]->password eq $acct[1]->password ) {
-        print "Passwords don't match.\n";
-        exit 1;
-    }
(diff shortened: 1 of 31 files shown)


zorkian:
Hey @momijizukamori I know you were looking at this, I got bored tonight and took a stab at it. I'm sure you've thought through all these edge cases a lot more than I did tonight, perhaps we can combine forces on this? I'd like to push it through this week if we can!

zorkian:
Ok @momijizukamori this merges our work, I think? Want to take a look?

PR #2622, opened 2020-04-21
Add local TaskQueue

zorkian:
This makes TaskQueue behave like BlobStore and use abstract backends.
The primary one is DW::TaskQueue::SQS but there is now also a new one
DW::TaskQueue::LocalDisk that is invoked if you're a development server
and haven't set up SQS.

This should allow developers to use the task system without having to
think about it too hard, but you'll still have to run workers like
production if you want to use them. TODO: make it synchronous?


zorkian:
cc @nfagerlund

PR #2624, opened 2020-04-21
Initial pass at TOTP (2fa)

zorkian:
This implements support for configuring 2fa using TOTP codes. At the
moment, this is not actually used in our authentication flows, but it at
least is the plumbing to do so.

TODO:
* recover account with recovery codes
* admin tools for manipulating 2fa
* support tools for showing 2fa status
* actual authentication integration for login
* think about all the other flows we use passwords?
* ???
* profit!

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -2217,9 +2217,22 @@ EOC
 
 register_tablecreate( "password2", <<'EOC');
 CREATE TABLE password2 (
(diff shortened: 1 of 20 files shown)


zorkian:
Let's get some screenshots.

# Account management page links to it now

image

# Index (when disabled)

image

# Setup

image

# Failed setup

image

# Scan QR code with Authy

image

# Successful setup

image

# Account status now

image

# Enabled management page

image

# Showing recovery codes

image

# Disabling flow

image

# Disabling error state

image

# Disabled OK

image

PR #2626, opened 2020-04-24
Add password 'peppering'

zorkian:
After discussion with a security engineering friend of mine, this is
take two at the secure password storage approach. Now, we are applying a
'pepper' on the web server, before we write to the database.

The advantage of this is that a database exfiltration does not result in
useful data (when it comes to passwords). In order to get useful data,
you need to get data off of the webserver memory/filesystem AND from the
database, which raises the bar.

The chosen peppering solution is symmetric encryption, which enables us
to do pepper rotation if we ever believe that our pepper key was
compromised somehow.

This also re-introduces @momijizukamori's schema/revision/version
column, so we can change how passwords are stored in the future.
Additionally this pulls most of the functionality out into
DW::Auth::Password which will let us isolate things better and ensure
that it's easy to audit credential management code.

### DIFF
--- a/bin/upgrading/d10-passwords.pl
+++ b/bin/upgrading/d10-passwords.pl
@@ -54,6 +54,12 @@ while (1) {
             or die "Failed to get password on $u->{user}($uid)!\n";
         $u->set_password( $password, force_bcrypt => 1 );
         $u->update_self( { dversion => 10 } );
+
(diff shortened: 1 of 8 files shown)


zorkian:
This also gets rid of caching passwords in memcache and the user object. Seems like we shouldn't need to hit them that much that we need to cache them...?

PR #2641, opened 2020-05-01
improvements to bin/checkconfig.pl

kareila:
This takes an existing stub of an idea -- making sure that etc/config-local.pl exists -- and expands it to include config.pl and config-private.pl, as well as the configs in t/ for testing. If the files exist, they are also evaluated to make sure they compile properly (no missing punctuation).

Having this would have saved me approximately two hours of pain yesterday.


zorkian:
Sorry for your lost two hours D: but ty for improving this!

Issue #2633, opened 2020-04-27
s2foundation beta: little stuff


nfagerlund:
Right now I think we've fixed everything, but I'm going to hold off on closing this until the beta goes opt-out and we get a new wave of people using it.

<details><summary>Completed</summary>

- ~Duplicated "thread" links in the comment footers! The first one is `.footer .commentpermalink`, and the old site skin CSS has a rule to set it as display: none. The second one is `.footer .thread`, and is apparently the one to keep. In conclusion... what?~
- ~In "Blue for Drifting" theme (possibly all Driftings), new talkform gets pushed down past the end of the sidebar content. The `clear: both` I put on on `#qrformdiv` to fix quickreply issues in a different style seems to be reacting to the float-based sidebar. I hate float-based layouts.~
- ~In lynx skin, the turquoise blue of the action box background makes it hard to distinguish the blue prev/next arrows.~ #2644 
- ~Actually, I could use some policy input about lynx. The old lynx comments had basically NO css, so adding normal CSS makes major changes.~ #2644 
- ~The new icon browser still wastes too much space in the upper area, esp. on mobile. I really want to tighten these controls up a bit more.~ (#2643) 
- ~Code blocks (like `<pre><code>...</code></pre>` as opposed to plain `<code>` spans) look whacked.~
- ~"action links" box on entries can sometimes impinge on large mood icons. (e.g. yhlee's mood theme.) Add some margin maybe.~
- ~The username indicator above quick-reply looks way too small on site skin pages.~
- ~Although the font size went up a bit for comments on tropo, it went DOWN for comments on gradation, which has always used 16px. Gotta make the scaling configurable per skin.~
- ~Should comment boxes in site skin `clear: both` to defend against floated images in the previous comment trashing the indentation (which is usually the only indicator of thread hierarchy)? Since these pages shouldn’t do any float layout bullshit in the beta, this should be safe and reasonable.~
- ~In site skin, can't resize reply form textarea wider than its parent.~
    - ~HOLY HELL, how many times do I have to fix this exact same bug??? This time it's Foundation's fault, not mine.~
- ~Tropo purple at mobile width has a weird display turd behind the username menu in the header. (Predates this, but more noticeable now.)~ #2653 
- ~The submit button for the "multiform" at the bottom of the page is missing the foundation button styles for some reason. Must be outside of an area-of-effect container.~

</details>

<details><summary>Not gonna bother with it for now</summary>

[...]

PR #2651, opened 2020-05-08
Gyve Lynx yts rabyes shot

### LINKED ISSUE #2644: Rehabilitate Lynx site skin's foundation version
The Foundationized version of Lynx (aka "style=light") has way too much actual styling in it; although the HTML frame of the page is appropriately chopped, the styling on the remaining inner sections looks exactly like the other foundationized site skins, just using turquoise (?!?!) as an accent color. 

As I understand it, this isn't desired, and Lynx is meant to be a web-brutalist promised land filled with juicy blue-and-purple links. However, until now, no one's really noticed it, because tbh the pages that have been converted to the foundationized skins so far are not pages that get heavily used. There are a few that might be _moderately_ used: 

- Image management stuff
- The beta new entry page

...but even those are likely NOT being used by the type of user who browses their flist from /mobile and sets every page to `=light`. 

The s2foundation beta changes that by imposing foundationized skins on the pages people spend actual time on; the "light" users who've turned it on have immediately noticed the problems.

Anyway, this is looking surprisingly tractable. I'm making some satisfying progress over at https://github.com/nfagerlund/dw-free/tree/may20_lynx_try2. 

- I seem to have gotten entry/reply/icons pages looking about like they used to. 
- All the simple foundationized pages I've tried seem fine. (Up to and including the image management pages.) 
- The beta icon browser looks great, since I already went to some trouble to make it work in journal pages. 
- The beta new entry page is the one remaining mess:
    - The fancy-selects crashed p. hard.
    - Section toggle icons floated away to heaven. (already fixed that.) 
    - Date picker's trashed.
    - Big chunks of the tag completer are transparent. 
    - Something weird happening to checkbox labels in the settings. (Is there some kinda float shit going on in there? Ah, no, it's splitting some inline containers across CSS columns.)
    - Textarea and text input widths go to hell at mobile widths. 
    - But all told, that's actually less busted than I expected...

nfagerlund:
Fixes #2644, plus a pile of minor issues from #2633. Oh, and a thing I never filed a ticket for, where, in the beta create entries page, you can click _exactly_ on the icon of the date/time picker buttons and cause slapstick. (The picker opens and then _immediately_ closes.) 

The Foundation version of the Lynx syte skyn has always been extremely Wrong, but folks let it slide because the really important pages didn't use the modern skins. This PR finally fixes that. It takes advantage of the work I did in the s2foundation beta to enable Foundation features (like the icon browser) on journal pages without forcing them to adopt Foundation styling; that was never possible before, and it turns out that's exactly what Lynx wanted. Please, won't you help preserve thys majestyc wyld kytti.

The beta create entries page is the most intricately bespoke Foundation page, and required the most compromises to keep working:

- Most of its textual elements and controls do use browser default styling, but the structural elements are definitely still heavily CSSed. 
- The fancy-select component is entirely unsuited for Lynx's morally pure web-brutalism, so I opted to frikken explode it. The "output" portion now goes side-by-side with a normal select, so it looks like an extended explanation instead of a replacement.

IDK; it's all beta, we can iterate. I think this is an okay start.

**Oh right:** See also https://github.com/dreamwidth/dw-nonfree/pull/132, which keeps tropo's journal content font size the same after these changes.

## Screenz

<details><summary>Lots of huge pictures</summary>

image

* * * 

image

* * * 

image

* * * 

image

* * * 

image

* * * 

image

</details>


nfagerlund:
## "Before"

<details><summary>Screenshots of the unrehabilitated Foundationized Lynx</summary>

image

image

image

image

</details>

htdocs/scss/skins/lynx.scss
```diff
@@ -1,40 +1,64 @@
...
 
-$body-bg:           $background-color;
-$body-font-color:   $text-color;
-$form-font-color:   rgba(0,0,0,0.75);
-$form-bg-color:     #f7f7f7;
+// Color variables required by entry-styles:
+$soft-accent-color: unset;
+$secondary-color-alternate: #c0c0c0;
+$secondary-color: #e2e2e2;
+$screened-comment-bg: #C29B9B; // old Tropo screened comment color, why not.
```
nfagerlund, line 22: Technically old Lynx had its own screened comment background color, but nobody's ever seen it because the CSS specificity was messed up and it got overridden by the default odd/even backgrounds. But if you go throw an !important on there in the web inspector, you'll see that it's unacceptable anyway.

rahaeli:
You are the best, Nick

htdocs/scss/skins/_journal-typography.scss
```diff
@@ -2,6 +2,8 @@
 * Typography resets for entries and comments:
 */
 
+$journal-content-font-size: 16px !default;
+
 .entry, #comments, .reply-page-wrapper {
   // Intended behavior:
   // - Skins that don't specify otherwise will display entries and comments
```
nfagerlund, line 9: oh... I forgot to update the comments. Sigh.

zorkian:
Going mostly by the screenshots, but this does appear to be what is actually intended as intended by the intentionality of using such a scheme.

That sentence was a hash, but I'm going to go with it.

PR #2656, opened 2020-05-16
require minimum version 0.015 of Params::Classify

kareila:
An older version of this module prevented Apache from starting in some environments. It's not explicitly called in our code, but it's included by Authen::Passphrase::BlowfishCrypt, which DW::Auth::Password uses.


zorkian:
O_o;

Thanks, Perl.

PR #2659, opened 2020-05-21
Investigate bizarre test behavior

nfagerlund:
I ran into some behavior that's blocking my ability to write tests for refactored features. 

Hopefully, I've just misunderstood something basic about how my test environment should be set up, in which case I could use some help from an adult. 

Alternately, maybe something's weird here.

### DIFF
--- a/t/ban-observer-effect.t
+++ b/t/ban-observer-effect.t
@@ -1,0 +1,36 @@
+# t/ban-observer-effect.t
+#
+# A completely empty can containing zero worms. :]
+#
+# Authors:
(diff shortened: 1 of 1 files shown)


nfagerlund:
Mark explained this to me! You can get around it by scorching `LJ::REQ_CACHE_REL`. I don't feel bad for not having known that, because it is horrible. 

https://github.com/dreamwidth/dw-free/blob/6d668cfcf64424705ebd32b68f6522780bc9a0ba/t/vgift-trans.t#L83

PR #2661, opened 2020-05-22
Convert faqsearch and faqbrowse to Foundation/TT

momijizukamori:
Think I managed to test everything other than translation functions, because I don't know how they work (if they even actually work)


cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -154,4 +156,301 @@ sub faqpop_handler {
...
+
+    # FIXME: disallow both faqid and faqcat (or ignore one)
+    my $faqcatarg = $GET->{'faqcat'};
+
+    unless ( $faqidarg || $faqcatarg ) {
+        return $r->redirect("faq");
+    }
+
+    # get language settings
+    my $curlang = $GET->{'lang'} || BML::get_language();
```
nfagerlund, line 192: Still want that BML call? (`<--` doesn't know the ML system very well)

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -154,4 +156,301 @@ sub faqpop_handler {
...
+        my $faqid = $f->faqid;    # Used throughout, including in interpolations
+        $dbh ||= LJ::get_db_writer();
+
+        # log this faq view
+        if ( $remote && LJ::is_enabled('faquses') ) {
+            $dbh->do( "REPLACE INTO faquses (faqid, userid, dateview) " . "VALUES (?, ?, NOW())",
+                undef, $faqid, $remote->{'userid'} );
+        }
+
+        BML::note_mod_time( $f->unixmodtime );
```
nfagerlund, line 271: ?

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -154,4 +156,301 @@ sub faqpop_handler {
...
+                unless $summary =~ m!<[^>]*\Q$qterm\E[^>]*>!;
+
+            $answer =~ s!((?:https?://[^>]+)?\Q$qterm\E)!$term->($1)!ige
+                unless $answer =~ m!<[^>]*\Q$qterm\E[^>]*>!;
+        }
+
+        # run hook to transform the text of this FAQ before it's rendered
+        # FIXME: hook is responsible for clean-up if it changes $display_*
+        # from false to true. Should that be checked/enforced here instead?
+        # FIXME: do we even need that hook? It looks like LJ only ever used
```
nfagerlund, line 329: Seems like an excellent time to just knife this.

views/support/faqbrowse.tt
```diff
@@ -0,0 +1,109 @@
+<script language='JavaScript'>
```
nfagerlund, line 1: 😬 This is extremely loathsome, but I don't think it'll _break,_ per se.

views/support/faqbrowse.tt
```diff
@@ -0,0 +1,109 @@
...
+                <a style='text-decoration: none;' href='faqbrowse?faqid=[% f.faqid %]'>&#x00bb; [%- f.question -%]</a>
+            </h1>
+        [%- END -%]
+        <div style='margin-left: 20px;'>
+            [%- IF f.display_summary -%]
+                <div class='summary'>[% f.summary %]</div>
+                [% UNLESS f.display_answer %]
+
+                    <div class='rm'>
+                        [%- oc = faqcatarg ? '' : "onclick='return showAnswer();'" -%]
```
nfagerlund, line 53: should this be `f.oc`?

views/support/faqbrowse.tt
```diff
@@ -0,0 +1,109 @@
...
+                        <b>( <a [% f.oc %] href='faqbrowse?faqid=[% "$f.faqid$q" %]&view=full'>[% '.more' | ml %]</a> )</b>
```
nfagerlund, line 54: That `"$f.faqid$q"` interpolation-thing looks suspect; would probably be better as multiple prints?

views/support/faqsearch.tt
```diff
@@ -0,0 +1,53 @@
...
+[%- END -%]
+
+[%- IF results -%]
+    [%- IF results.size < 1 -%]
+        <p><i>[% '.error.noresults' | ml %]</i></p>
+    [%- ELSE -%]
+
+    <ul class='spaced'>
+        [%- FOR r IN results -%]
+            <li><a href='/support/faqbrowse?faqid=[% r.id %]&q=[% r.ueq _ r.ul %]'>[% r.dq %]</a></li>
```
nfagerlund, line 49: What's that underscore operator in `r.ueq _ r.ul` ?

momijizukamori: That's Template Toolkit's concatenate operator, for whatever reason.

nfagerlund: 💀

views/support/faqbrowse.tt
rshatch: Honestly I'm not sure we actually use the expand-in-place stuff, so I could probably just rip it out 🤔

views/support/faqbrowse.tt
rshatch: Nah, it's setting a new variable named oc, not pulling one from the faq object. That said this is only used for the expand-in-place JS which we don't appear to use.

cgi-bin/DW/Controller/Support/Faq.pm
```diff
@@ -154,4 +156,301 @@ sub faqpop_handler {
...
+        $user     = $remote->user;
+        $user_url = $remote->journal_base;
+    }
+    else {
+        my $u = LJ::load_user($LJ::EXAMPLE_USER_ACCOUNT);
+        $user     = $u ? $u->user         : "<b>[Unknown or undefined example username]</b>";
+        $user_url = $u ? $u->journal_base : "<b>[Unknown or undefined example username]</b>";
+    }
+
+    # get faqid and redirect to faq.bml if none
```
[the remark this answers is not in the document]
rshatch: yes T_T

views/support/faqbrowse.tt
rshatch: So, it works, which means it's a question of whether it'd be more readable as multiple prints. I tend to feel like having a bunch of separate TT prints next to each other feels clunky and doesn't give as clear an idea of what the final result is, but I'm open to other opinions.

nfagerlund: Ok cool! Now that my brain isn't attuned to hunting BMLisms, it kind of looks fine again, so never mind.

views/support/faqbrowse.tt
nfagerlund: AH! I was reading that "=" as a comparison operator, not assignment, and it shifted the context of the whole ternary. (Part of why I kind of dislike ternaries tbh, but there's kind of nothing for it in Perl since if isn't an expression.)

PR #2664, opened 2020-05-23
Add inline session messages

momijizukamori:


schemes/global.tt
```diff
@@ -102,6 +102,23 @@ the same terms as Perl itself.  For a copy of the license, please reference
 [%- END -%]
 [%- END -%]
 
+[%- BLOCK block.msgs -%]
+[%# sections.errors are instances of DW::FormErrors %]
```
zorkian, line 106: Is this comment true in the block.msgs too?

zorkian:
Do you have any screenshots of how this looks, and also, is this being used by another PR? (I haven't gone through all the PRs yet.)

schemes/global.tt
momijizukamori: oops, nope, that's what I get for copying the formerror block wholesale to figure out how to stick it into all templates.

momijizukamori:
[View on Celerity](https://cdn.discordapp.com/attachments/403273218548301866/713850865479450684/unknown.png) - it reuses the alert-pane styling and colors. Most actual instances will mostly like have far fewer messages, and are less likely to have the mix of message types, I just wanted to make sure they all showed up properly.

zorkian:
Gotcha, thank you! That looks like what I expected (except the Celerity part haha).

PR #2665, opened 2020-05-24
The Squish, Part 3

### LINKED ISSUE #2611: Hide zoom cursors for images that can't be zoomed
In https://github.com/dreamwidth/dw-free/pull/2595, I added click-to-zoom behavior for images in entry/comment content, to supplement the existing shrink-to-fit CSS. And as a visual affordance, it changes the cursor (in desktop browsers) to a zoom-in or zoom-out magnifying glass when hovering over an image in content. 

But that also affects images that didn't get shrunk (and thus can't be zoomed). It'd be nice to exclude those! 

Some of my speculation from https://github.com/dreamwidth/dw-free/issues/2587 is still relevant. I THINK we could do something like the following: 

- Write a function to:
    - Select content images, and filter out any `.expanded` or ineligible ones. 
    - Compare each image's `naturalWidth` / `width` and `naturalHeight` / `height` properties; if equal add a `.nozoom` class, and if unequal remove it. 
    - Include a de-bounce behavior so we only call it after resizing stops. Can delay by a full second or longer, probably; this is a real low-priority update.
- Call that function on `load` for the whole document.
- Attach a listener to the window `resize` event to call that function. (Don't care about mobile, so don't need to worry about orientationchange.)
    - ...but suppress it if someone resizes before `load` has actually happened. So actually, I guess use the load handler to attach this resize handler. 
- Also set up some `MutationObservers` with `{subtree: true, childList: true}` to watch for new elements being added to the page; call that function if there are new images. (Cut tags, expanded comments.)
    - <del>...but only call it on whatever mutationobeserver's equivalent of `load` is, to make sure the images are fully downloaded and their sizes are known.</del>
    - There's no good way to tell when an arbitrary set of newly introduced elements has loaded, it looks like, so we should make sure there's also a way to only run this check on a single image, and use the mutationobserver to attach that as the onload handler for every new image.

### LINKED ISSUE #2650: The Squash, Part 3 (Save the RP Museboard Artisans)
This is distinct from The Squash, Part 2.5 (#2611, hide irrelevant zoom cursors) and the not-yet-documented The Squash, Part 2.8 (provide interface for zooming link-wrapped images without leaving the page). 

## Previously

**The Squash, Part 1:** CSS to shrink images in entry/comment content, to prevent them from:

- Overflowing their container width.
- Failing to fit in the browser viewport.

**The Squash, Part 2:** Click-to-zoom, to rescue 4-comi vertical strips and let people opt into overflows for important pics.

## Now

DW RPers (and certain other subcultures) like to use carefully-crafted fragments of markup with inline CSS to create intricate decorative units, for a variety of conversational/navigational/emotive purposes. In the local dialect, these inline-styled markup fragments are called "codes." The Squash has a tendency to vandalize them.

On reflection, all I care about with The Squash is **casual images.** There's always going to be a way around this CSS, if anyone REALLY wants to prank their followers with an flist overflow; all _I_ want to do is make it so anyone can huck an `` up on there without stressing out about possibly being rude to your peeps. 

I _don't_ want to vandalize anyone's codes, and I especially don't want to keep fielding complaints and requests from the people whose codes I've vandalized. So, maybe we can find a semi-reliable way to exempt intricately styled markup from The Squash, while still hitting 95% of casual images. 

## Ideas

I had two initial ideas: 

- Look for `height=` and `width=` attributes on the img tags, and treat them as a (partial?) indicator that the author knows what they're doing. (This is problematic because a lot of things that give you pasteable markup for casual images will automate those attributes.) 
- Look for deeply nested structures or some other marker of complexity. 

## Complications

- A lot of casual images are pasted from "share" fields on image hosts, and they might have surprisingly complicated markup. Need to investigate prevalence of that, and also to what extent DW actually tolerates their weird shit. (e.g. Flickr uses JS to spawn an iframe, given its druthers.)
    - iframe contents are unaffected by the embedding document's CSS anyway, so shrug. 
    - html cleaner seems to disallow random embed JS, so what you paste is what you get. In flickr's case, that's `a img`. 👍🏼 
    - Instagram are such creeps. Their image embeds don't even appear on sites that don't allow their spyware JS to run. So, don't have to worry about them. (Although that test did turn up another context where @ mentions aren't appropriate to render.)
- Various feeds wrap their casual images in markup that might look like a code to an insufficiently rigorous check, because they expect to be displaying in a totally different original context. (Shit like `div div img`, `figure a img`...) Casual images in feeds are kind of public flist enemy number 1.
- Combinatorial CSS selector explosion.

nfagerlund:
Fixes #2650
Fixes #2611
 
Shoutout to Alyndra, @rshatch, and AlexSeanchai for helping test and refine this.

We currently constrain the size of images in entry and comment content. But the
REASON we do this is so that people can casually post medium-sized images
and not stress about annoying their friends and readers; the beginning and end
of it is to make it easier to post on DW.

There's another set of user behaviors that's basically the OPPOSITE of casually
posting a medium image, where people express themselves by doing weird intricate
HTML shit no modern site would allow. It's part of DW's fandom-centric culture,
so we'd rather not interfere with it or make it harder.

This commit uses some extra javascript to try and guess which images are part of
a complex hunk of decorative / narrative markup, and changes the CSS to leave
said images alone (under the assumption that the author knows what they're
trying to accomplish and doesn't need any "help").

And while I'm at it, it also removes the "zoom in" cursor from images that are
already actual size and thus don't need to zoom. (That part uses an ultra-modern
API with stringent browser requirements, because 1. it's kind of just a vanity
thing, and 2. it has to update on window resize, but any tangible performance
cost is unacceptable, so if we can't do it for nearly-free then we shouldn't do
it at all. Native ResizeObserver is nearly-free, polyfills ain't.)

### DIFF
--- a/htdocs/js/jquery.imageshrink.js
+++ b/htdocs/js/jquery.imageshrink.js
@@ -1,5 +1,97 @@
-$(document).on('click', '.entry-content img, .comment-content img', function(e){
-    if ( ! $(e.target).is('a img, .poll-response img') ) {
-        $(e.target).toggleClass('expanded');
+// Helper for shrinking images in entry/comment content. The actual Squish is
+// pure CSS; this just handles adding and removing classes to support:
+// - Zooming (on click)
+// - Exempting decorative markup from Squish (should be pure CSS, but can't yet)
+// - Hiding zoom cursors for images that won't zoom
+jQuery(function($) {
+
+    // First: Basic click-to-zoom. (.imageshrink-expanded on/off)
+    $(document).on('click', '.entry-content img, .comment-content img', function(e) {
+        var $that = $(e.target);
+        if ( ! $that.is('a img, .poll-response img, .imageshrink-actualsize') ) {
+            $that.toggleClass('imageshrink-expanded');
+        }
+    });
+
+    // Second: Exempt codes from the squish. (.imageshrink-exempt on/off)
+    // Only shrink casual pics; leave artisanal HTML alone. Can't predict
+    // everything, but 99% of the time that's a <div style="..."> or a table.
+    function protectTheCodes(container) {
+        // Expanded cut tags have style='display: block', but aren't decorations.
+        // Feeds (.journal-type-Y) always make a mess, so no mercy.
+        var exemptImages = container.querySelectorAll(
+            '.journal-type-P .entry-content div[style]:not(.cuttag-open) img, ' +
+            '.journal-type-P .entry-content table img, ' +
+            '.journal-type-C .entry-content div[style]:not(.cuttag-open) img, ' +
+            '.journal-type-C .entry-content table img, ' +
+            '.comment-content div[style] img, ' +
+            '.comment-content table img'
+        );
+        for (var i = 0; i < exemptImages.length; i++) {
+            exemptImages[i].classList.add('imageshrink-exempt');
+        }
     }
+
+    // Check the whole document to start with.
+    protectTheCodes(document);
+
+    // Third: Don't show zoom cursors for non-zooming images (.imageshrink-actualsize on/off)
+
+    // Dummied out observeImages function, to simplify browser support in (Fourth).
+    var observeImages = function(img) {return;};
+
+    if (typeof ResizeObserver === 'function') {
+        var zoomCursorCleaner = new ResizeObserver(function(resizeList, observer) {
+            resizeList.forEach(function(entry) {
+                var img = entry.target;
+                if (img.tagName !== 'IMG') { return; }
+
+                // Zoomed images don't count.
+                if ( ! img.classList.contains('imageshrink-expanded')
+                    && img.width === img.naturalWidth
+                    && img.height === img.naturalHeight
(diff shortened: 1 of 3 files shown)


nfagerlund:
Hmm, while debugging something else, I found out this doesn't play nicely with Foundation site skins yet. Will investigate.

htdocs/js/jquery.imageshrink.js
```diff
@@ -1,5 +1,97 @@
...
+        }
     }
+
+    // Check the whole document to start with.
+    protectTheCodes(document);
+
+    // Third: Don't show zoom cursors for non-zooming images (.imageshrink-actualsize on/off)
+
+    // Dummied out observeImages function, to simplify browser support in (Fourth).
+    var observeImages = function(img) {return;};
```
nfagerlund, line 41:
```suggestion
    var observeImages = function(container) {return;};
```

I think I changed what that arg was during a refactor and missed renaming it in the stub.

PR #2666, opened 2020-05-24
Prevent infinite loop in htmlcleaner

nfagerlund:
N.B.: @alierak discovered this and did some kind of hotfix for it, so he should get the commit credit if he wants it. I'm just tossing this up here to make sure it doesn't get forgotten. 

If the user's raw text has more closing `</cut>`s than it had opening `<cut>`s
(resulting in a negative `opencount{'lj-cut'}`), don't keep adding closing
`</cut>`s until the end of time.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1352,8 +1352,8 @@ TOKEN:
     }
 
     # If crossposting, explicitly close cuts to keep the crosspost footer visible.
-    if ($preserve_lj_tags_for) {
-        while ( $opencount{'lj-cut'} ) {
+    if ( $preserve_lj_tags_for && $opencount{'lj-cut'} ) {
+        while ( $opencount{'lj-cut'} > 0 ) {
             $newdata .= "</lj-cut>";
             $opencount{'lj-cut'}--;
         }


alierak:
Here's what I did:

```diff --git a/cgi-bin/LJ/CleanHTML.pm b/cgi-bin/LJ/CleanHTML.pm
index 127e969..b745fb9 100644
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1105,7 +1105,7 @@ TOKEN:
                 # state variable which is stashed across tokens.
             }
             elsif ( $tag eq "lj-cut" ) {
-                if ( $opts->{preserve_lj_tags_for} ) {
+                if ( $opts->{preserve_lj_tags_for} && $opencount{'lj-cut'} ) {
                     $opencount{'lj-cut'}--;
                     $newdata .= "</lj-cut>";
                 }
```

This prevents the count from going negative in the first place, and also avoids copying an unmatched `</lj-cut>` into the output. I also realized that if we don't stop the count from going negative, it would allow something like `</cut><cut>` without the cleaner closing it at the end. 

But there's probably no harm in doing this plus protecting the while loop.

kareila:
This is now #2669.

PR #2668, opened 2020-05-26
convert /admin/invite pages to TT

kareila:
Part of the ongoing effort to remove BML pages from the site.


zorkian:
We need a BML killing theme song. Another one bites the dust is probably too easy.

rahaeli:
We Are The Champions?

kareila:
IDK about BML conversion but this is the official code machete song: https://www.youtube.com/watch?v=-VaVKE82wMs

rahaeli:
I think that works for BML conversion too!

PR #2669, opened 2020-05-26
Prevent infinite loop in htmlcleaner, belt + suspenders version

kareila:
One pull request combining both @nfagerlund's fix and @alierak's fix, because why not both.

(I'm merely the git-wielding monkey in this merge scenario.)


zorkian:
Props for debugging this, everybody. Saw the scrollback and was super happy to have a team like y'all :)

PR #2670, opened 2020-05-28
fix tidy failures

kareila:
Looks like the recent experimental email worker changes were untidied.


zorkian:
My sins were many, sorry about that :'(

PR #2671, opened 2020-05-28
convert admin/pay pages to TT

kareila:
Impossible to test these thoroughly in a dev environment, but I think I covered the obvious stuff.

Some reorganization here:

- pay/index is split into two views, index.tt and viewuser.tt

- the striptimefrom path in pay/view is now its own page, pay/striptime

The pay/import page was a forgotten relic of bygone days and we shall never
speak of it again.


zorkian:
I tried to fix merge conflicts, but now it says it still can't be rebased. It LGTM, so if you want to merge it after poking the conflict dragon, go for it? Else LMK and I can try to do so.

kareila:
I redid the "remove /admin/pay from legacy index" commit and then tidied up, so this should be good now! (That's why it was a separate commit, so I could just drop it in the rebase.)

kareila:
Merging this so we can hopefully get the master branch tidy again.

Issue #2350, opened 2018-07-09
Prevent personal account from being converted to a community if it still administers other communities?


kareila:
There doesn't appear to be any sort of sanity checking regarding community administration when converting a personal journal to a community.  We already prevent conversion if the journal isn't empty; we should also make sure that particular list of edges is empty as well.  Otherwise it's easy to end up with a community whose only admin is another community.

Once this change is made, FAQ 178 should be updated as well.

kareila:
Reopening for update of FAQ 178 once this goes live.

kareila:
Ugh, it looks like there's a second, completely independent path for this conversion, in the convert_handler defined in DW::Controller::Community. That's going to need a similar fix and maybe some code refactoring to have a single path.

kareila:
FAQ is updated!

PR #2674, opened 2020-05-29
Fix change_journal_type console command

kareila:
This command had been temporarily disabled because when changing a community journal to a personal journal, the old behavior was to copy both the email and password of the named community admin over to the converted journal. But the previously used method for accessing that password was recently removed in favor of better password practices, so we needed a different solution for that aspect of the change.

The modifications here are copied in part from the `reset_password` console command, which uses LJ::rand_chars to generate a random string and assigns that as the value of the new password, then communicates it directly to the user via email with the expectation that it will immediately be changed.

The actions taken by the change_journal_type console command are executed by the `reset_email` user method, which sends a message asking the user to confirm the new email address for the journal. For the case where we are also changing the password, the message sent now asks the user to both confirm the email and change the password. In the case of converting to a community, the password is silently reset to the empty string.

While I was in here, I also took care of #2350 and cleaned up the test suite a bit.

### DIFF
--- a/cgi-bin/LJ/User/Message.pm
+++ b/cgi-bin/LJ/User/Message.pm
@@ -183,6 +183,13 @@ sub reset_email {
 
     $update_opts ||= { status => 'T' };
     $update_opts->{email} = $newemail;
+
+    # this is no longer done using update_self
+    my $changepass = delete $update_opts->{password};
+    if ( defined $changepass ) {
+        $u->set_password($changepass);
+    }
+
     $u->update_self($update_opts)
         or return $errsub->( LJ::Lang::ml( "email.emailreset.error", { user => $u->user } ) );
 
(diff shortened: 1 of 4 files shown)


cgi-bin/LJ/User/Message.pm
```diff
@@ -183,6 +183,13 @@ sub reset_email {
 
     $update_opts ||= { status => 'T' };
     $update_opts->{email} = $newemail;
+
+    # this is no longer done using update_self
+    my $changepass = delete $update_opts->{password};
+    if ( defined $changepass ) {
+        $u->set_password($changepass);
+    }
+
```
kareila, line 192: I just realized that I should clarify - this isn't strictly necessary since `update_self` does eventually do the correct thing, but since I had to check the value of the password option anyway, I thought the logic would be clearer if I explicitly call `set_password` here.

zorkian: Yeah I like the usage of set_password here, good call!

cgi-bin/LJ/User/Message.pm
```diff
@@ -198,13 +205,26 @@ sub reset_email {
...
-                        auth     => $auth
-                    }
+                    $changepass
+                    ? (
+                        "email.emailreset.body_withpasswd",
+                        {
+                            user     => $u->user,
+                            newpass  => $changepass,
+                            sitename => $LJ::SITENAME,
+                            siteroot => "$LJ::SITEROOT/",
```
zorkian, line 215: oh tfw

t/console-changejournaltype.t
```diff
@@ -66,11 +73,16 @@ foreach my $to (qw(person community)) {
...
-    $run->("change_journal_type $commname shared $owner"),
-    "error: Type argument must be 'person' or 'community'."
-);
+    if ( $comm->is_community ) {
 
+        # have to check the database directly because convenience methods
+        # know that communities aren't supposed to have passwords
+
+        my $dbh  = LJ::get_db_writer() or die "Couldn't get db master";
+        my $pass = $dbh->selectrow_array( q{SELECT password FROM password WHERE userid = ?},
```
zorkian, line 83: Shouldn't this be `password2` now?

cgi-bin/LJ/User/Message.pm
kareila: yeah, the other way is probably better but oh well

t/console-changejournaltype.t
kareila: Here's what I couldn't figure out: you can't read the password2 table to see if the password is blank, but you also can't use the check_password method here because it immediately bails if you're trying to check the password of a community. Hence the test message saying OK if "community password is blank or not stored in password table"

zorkian: Ahh alright. ;;

PR #2677, opened 2020-06-01
Remove unused admin pages

kareila:
Discussion with Denise & Mark determined that the following admin pages were no longer worth keeping around:

- /admin/memcache and /admin/memcache_view (still BML)
- /admin/clusterstatus and /admin/mysql_status (in DW/Controller/Admin/StatusCheckLegacy.pm)

This touches some of the same files as my /admin/pay conversion (#2671) but I believe they can both be merged independently.


cgi-bin/DW/Controller/Admin.pm
```diff
@@ -45,22 +45,6 @@ DW::Controller::Admin->_register_admin_pages_legacy(
...
-        [
-            'siteadmin:memcacheview',
-            'siteadmin:*',
-            sub {
-                return ( $LJ::IS_DEV_SERVER, LJ::Lang::ml("/admin/index.tt.devserver") );
-            }
-        ]
-    ],
     [
         'pay/',
```
kareila, line 49: I realized when I saw the failed build status that tidyall wants to reformat the surrounding lines after removing this block, but that will definitely interfere with merging the admin/pay changes...

zorkian: It's probably OK to leave broken a few days while we merge... :s

PR #2680, opened 2020-06-03
Markup type switching, in comment forms and beta create entries

### LINKED ISSUE #2573: Enable markdown in comment forms
Today, DW has three markup dialects for entries and comments:

- HTML with auto-linebreaks (default)
- HTML without auto-linebreaks
- Markdown

(I'm eliding some complexity here, because there are multiple interacting properties that produce all this. But that's the observable behavior.) 

You can switch to raw HTML for comments, but only if you're replying from the full talkform. And you can't use Markdown in the comment forms at all. But the machinery to deal with Markdown comments is already here! Reply-by-email comments actually get saved and displayed as Markdown.

So, the comment forms should let you choose your markup dialect. 

I already know how to do this, but:

- ~This should wait until after #2540, because that might change the underpinnings of the `editor` prop.~
- Should also wait for #664, which might influence the desired design.

Once this is up and running, should make another ticket for editor helper buttons, to simplify basic markup like bold, italic, lists, links, images, etc. That's another one that will be heavily influenced by #664.

### LINKED ISSUE #2608: Nail down formatting dialects and `editor` prop
DW uses several different dialects of formatting. Currently they're implemented as an array of semi-independent feature switches and heuristics. But @zorkian centralized and normalized some stuff in the HTML cleaner, and now we have a nice handle for rationalizing some of this stuff and guarding archived content against future formatting shifts. 

I propose we centralize all content formatting questions in the `editor` prop on entries and comments. (That prop previously existed for comments, and is currently used to mark emailed replies as markdown.) For existing content with an undefined `editor`, we can autodetect into the appropriate older dialect before handing off to the cleaner function; for all new content, we’ll save an explicit editor value. 

Note that this is about the formatting languages used in the *content,* not about the different kinds of output formatting (e.g. normal display, crossposting, truncate, etc.). 

Here are the dialects I'm aware of, and their behavior.

- `html0` — text w/ HTML tags for formatting, written between 2009 and mid-2019 OR imported from LJ at any time.
    - Auto newlines (replace `\n` with `<br>`)
    - Auto links (wrap bare text URLs with a clickable link to that URL)
    - Allows special DW tags (cut, user, etc.) plus legacy LJ tags (lj-cut, lj, etc.)
    - No `@mentions`.
- `html1` — text w/ HTML tags for formatting, written from 2019 onward.
    - Auto newlines
    - Auto links (wrap bare text URLs with a clickable link to that URL)
    - Allows special DW tags (cut, user, etc.) plus legacy LJ tags (lj-cut, lj, etc.)
    - Yes `@mentions`.
        - excluded contexts: code, pre, textarea, raw-code, twitter's magic blockquotes.
    - When to bump: Adding new syntax on the order of `@mentions`, adding new kinds of auto-formatting that would misinterpret older content, removing legacy lj tags, changing how auto line breaks/paragraphs work.
- `html_raw0` — real HTML, possibly generated by something but still intended for posting on DW. Also imported from LJ with the preformatted flag set.
    - No auto newlines, gotta use p tags. No auto links either.
    - Allows special DW tags (and LJ forms).
    - No `@mentions`.
    - When to bump: Might not ever need to, but if we decided to restrict allowed tags or something, then maybe. Also if we wanted to enable `@mentions` here. 
- `html_extra_raw` — Real HTML, not originally intended for posting on DW. (Syndicated content, for the most part.)
    - No auto newlines or auto links.
    - No special DW/LJ tags.
    - No `@mentions`.
    - When to bump: Probably never, because the definition of “what is html” mostly isn’t our problem and html5 is a living standard.
- `markdown0` — Markdown implemented with `Text::Markdown` module.
    - Yes `@mentions`.
        - excluded contexts: code, pre, textarea, raw-code, twitter's magic blockquotes.
    - Gets processed by Markdown, then treated as basically `html_raw` except for allowing `@mentions`. (So yes DW/LJ tags, no auto-links, and paragraph breaks were already handled by the MD step.)
    - When to bump:
        - If we ever change markdown processors or fork our existing one (to get better performance, new syntax, etc.). (If anyone is thinking "c'mon, how incompatible could two markdown processors be," please just trust me on this one. I can go into details later.)
        - If we decide to add new syntax to our own Markdown dialect, something on the order of `@mentions`.
- `rich0` (?) — Generated by the old RTE.
    - Actually, I‘m pretty positive this is just `html_raw0`. So, probably doesn't need a separate format.
    - Looks like RTE correctly sets the preformatted flag, so autodetection of old content should catch that it's `html_raw0`.
- `rich1` — Need to discuss w/ @wohali!
    - My tentative understanding is that it should _act_ like `html_raw0`, but needs a separate editor ID so that we can warn people whenever they're switching to or from RTE that there might be some lossy formatting conversion.
    - Do we need to handle special tags differently? Unsure yet, depends on editor’s behavior.
    - `@mentions` hopefully just handled interactively by the editor and saved as normal user tags.

nfagerlund:
Fixes #2573 
Fixes #2608 

### Notable 🚨 s: 
 
- Two new userprops.
- Imposes new semantics on an existing entry prop and comment prop.
- Imposes new names for things that never used to have a name before. (DW's default HTML-with-auto-linebreaks format is now "casual HTML".)
- Changes design of comment forms again. 
- Requires a post-deploy string update on prod for a help/faq link. 
    - (Requires said faq page to be written, which I can help with.)
- Affects implementation of new RTE.

-----

### The dirt

Mostly copied from one of the earlier commits in here: 

This PR introduces "named markup formats" as a central interface for dealing with user-entered text.

Each unit of user-provided body text (entry and comment bodies) can include, as metadata, the name of the markup format it uses. There's already a space reserved for this: both entries and comments include an `editor` prop, originally intended for exactly this.

We'll use these named formats for two purposes:

- **Picking the correct formatting options for the HTML cleaner.** DW stores source text raw, and then formats it on display; the saved "editor" value will be a clear signal of exactly what the user expected to have happen when they wrote the post, so we can freely change our markup behaviors in the present day (adding `@mentions`, changing Markdown processors, etc.) without worrying about wreaking havoc on archived content. Formats are versioned, and any significant change to markup behavior is now expected to result in a new format name.

- **Picking the correct format-dependent states and behaviors for the create entries form and the comment forms.** In the initial implementation, this only affects the selection state of the markup format drop-down. In the future, it will affect the output of formatting helper buttons for plain-text formats, and will control whether to display the plain-text or rich text editor. In the event of a future migration from one rich text editor to another, named formats will allow us to treat the old one and the new one as separate entities; we won't need to ensure the new editor can accommodate text created by the old one, and can even keep both editors operational for some time during the transition.

The initial set of formats is a codification of the formats DW **already uses.** It's just that they didn't have names before, and were selected by emergent interactions between the `opt_preformatted` prop, the imported/syndicated status, Secret Glyphs (`!markdown`), and the editor prop (only supported in comment-by-email).

### Likely future work after this PR

- Simple markup helper buttons (bold, link, etc.) for plaintext editing modes. 
- Integrate formats into the new entries API (which isn't merged yet). 
- User profile bios get format selector. (Needs an additional prop somewhere, probably a singleton userprop?)
- Inbox messaging gets format selector? ......maybe, theoretically, but I'm not touching that until Momiji finishes converting the inbox off BML. 

-----

### Screenz

<details><summary>Here's some pictures.</summary>

QR on journal page: 

Screenshot_2020-06-03 sheworks4themoney Recent Entries

Screenshot_2020-06-03 sheworks4themoney Recent Entries(1)

New talkform on siteskin: 

Screenshot_2020-06-03 sheworks4themoney (no subject) (Reply)

Screenshot_2020-06-03 sheworks4themoney (no subject) (Reply)(1)

Old talkform, why not. 

Screenshot_2020-06-03 sheworks4themoney (no subject) (Reply)(2)

Beta create entries: 

Screenshot_2020-06-03 Create Entries

Screenshot_2020-06-03 Create Entries(1)

</details>

### DIFF
--- a/cgi-bin/DW/Controller/Editor.pm
+++ b/cgi-bin/DW/Controller/Editor.pm
@@ -15,6 +15,7 @@
 package DW::Controller::Editor;
 
 use strict;
+use v5.10;
 use LJ::JSON;
 use DW::Controller;
 use DW::Routing;
@@ -60,10 +61,12 @@ sub default_editor_handler {
     my ( $ok, $rv ) = controller( form_auth => 1 );
     return $rv unless $ok;
 
-    my $r      = $rv->{r};
-    my $POST   = $r->post_args;
+    my $r = $rv->{r};
+    return error_ml('/default_editor.tt.error.nopost') unless $r->did_post;
     my $remote = $rv->{remote};
     return error_ml('/default_editor.tt.error.nouser') unless $remote;
+
+    my $POST = $r->post_args;
 
     my $type       = $POST->{type};
     my $new_editor = $POST->{new_editor};
--- a/cgi-bin/DW/Controller/Editor.pm
+++ b/cgi-bin/DW/Controller/Editor.pm
@@ -1,0 +1,119 @@
+# DW::Controller::Editor
+#
+# Single-purpose routes for setting the editor userprops, which determine the
+# default formatting type when writing new entries/comments/etc in the web UI.
+#
+# Authors:
+#   Nick Fagerlund <nick.fagerlund@gmail.com>
+#
+# Copyright (c) 2020 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+package DW::Controller::Editor;
+
+use strict;
+use LJ::JSON;
+use DW::Controller;
+use DW::Routing;
+use DW::Template;
+use DW::Formats;
+use Carp;
+
+DW::Routing->register_string( "/default_editor", \&default_editor_handler, app => 1 );
+
+DW::Routing->register_rpc(
+    "default_editor", \&default_editor_rpc_handler,
+    format  => 'json',
+    methods => { POST => 1 },
+);
+
+# Returns (1, "format_id") on success, (0, "error") on error.
+sub default_editor_helper {
+    my ( $remote, $type, $new_editor ) = @_;
+    my $userprop;
+
+    my $err = sub { return ( 0, $_[0] ); };
+
+    return $err->('nouser') unless $remote;
+
+    $new_editor = DW::Formats::validate($new_editor);
+
+    if ( $type eq 'comment' ) {
+        $userprop = 'comment_editor';
+    }
+    elsif ( $type eq 'entry' ) {
+        $userprop = 'entry_editor2';
+    }
+    else {
+        return $err->('unknowntype');
+    }
+
+    $remote->set_prop( $userprop, $new_editor );
+    return ( 1, $new_editor );
+}
+
+# zero-javascript version
+sub default_editor_handler {
+    my ( $ok, $rv ) = controller( form_auth => 1 );
+    return $rv unless $ok;
+
+    my $r      = $rv->{r};
+    my $POST   = $r->post_args;
+    my $remote = $rv->{remote};
+    return error_ml('/default_editor.tt.error.nouser') unless $remote;
+
+    my $type       = $POST->{type};
+    my $new_editor = $POST->{new_editor};
+    my $exit_text  = $POST->{exit_text};
+    my $exit_url   = $POST->{exit_url};
+
+    my ( $set_ok, $set_result ) = default_editor_helper( $remote, $type, $new_editor );
+
+    unless ($set_ok) {
+        return error_ml( "/default_editor.tt.error.$set_result", { type => $type } );
+    }
+
+    return DW::Template->render_template(
+        'default_editor.tt',
+        {
+            title      => ".title.$type",
+            type       => $type,
+            new_format => DW::Formats::display_name($set_result),
+            exit_text  => $exit_text,
+            exit_url   => $exit_url,
+        }
+    );
+}
+
+sub default_editor_rpc_handler {
+    my $r    = DW::Request->get;
+    my $POST = $r->post_args;
+
+    # make sure we have a user of some sort
+    my $remote = LJ::get_remote();
+    return DW::RPC->err('Unable to load user for call.') unless $remote;
+
+    my $type       = $POST->{type};
+    my $new_editor = $POST->{new_editor};
+
+    my ( $set_ok, $set_result ) = default_editor_helper( $remote, $type, $new_editor );
+
+    return DW::RPC->err( LJ::Lang::ml( "/default_editor.tt.error.$set_result", { type => $type } ) )
+        unless $set_ok;
+
+    return DW::RPC->out(
+        success    => 1,
+        new_editor => $set_result,
+        message    => LJ::ehtml(
+            LJ::Lang::ml(
+                '/default_editor.tt.success',
+                { type => $type, new_format => DW::Formats::display_name($set_result) }
+            )
+        )
+    );
+}
+
+1;
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -114,13 +114,13 @@ sub clean {
     my $remove_sizes            = $opts->{remove_sizes} || 0;
     my $remove_abs_sizes        = $opts->{remove_abs_sizes} || 0;
     my $remove_fonts            = $opts->{remove_fonts} || 0;
-    my $local_content           = $opts->{local_content} || 0;
-    my $formatting              = $opts->{formatting} // 'html';
-    my $auto_links              = !( $extractlinks || $opts->{noautolinks} );
+    my $at_mentions = $opts->{at_mentions} || 0;                   # @person.place -> user tag
+    my $formatting  = $opts->{formatting} // 'html';               # html, or do we need to convert?
+    my $auto_links  = !( $extractlinks || $opts->{noautolinks} );
 
-    $auto_links    = 0 if $formatting ne 'html';
-    $cut           = 0 if $nodwtags;
-    $local_content = 0 if $nodwtags;
+    $auto_links  = 0 if $formatting ne 'html';
+    $cut         = 0 if $nodwtags;
+    $at_mentions = 0 if $nodwtags;
 
     my $blocked_links =
         ( exists $opts->{'blocked_links'} ) ? $opts->{'blocked_links'} : \@LJ::BLOCKED_LINKS;
@@ -409,7 +409,11 @@ TOKEN:
                 $attr->{value} = "sameDomain" if $attr->{value} ne 'never';
             }
 
-            if ( $tag eq "span" && lc $attr->{class} eq "ljuser" && !$noexpand_embedded && !$nodwtags ) {
+            if (   $tag eq "span"
+                && lc $attr->{class} eq "ljuser"
+                && !$noexpand_embedded
+                && !$nodwtags )
+            {
                 $eating_ljuser_span = 1;
                 $ljuser_text_node   = "";
             }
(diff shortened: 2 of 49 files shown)


nfagerlund:
### Questions I think someone will probably have at some point

#### What about the RTE? 

The work in this PR should lay usable groundwork for a new RTE and not sabotage anything for it, while still making the site more usable without it. 

The old RTE just resulted in `html_raw0` (via opt_preformatted), so entries created with it remain fine.

From talking with @wohali a long time back, it sounded like her new RTE will output markup that expects to be displayed exactly like `html_raw0`. I believe it should have its own dedicated format name (`rich1`) that happens to produce the same cleaner arguments, because that will let the create entries page restore the right state when editing an existing entry that was created with it. (As well as set the user's preferred editor correctly when saving new entries.) 

We also discussed at some point adding some significant friction (basically an "ARE U SURE" prompt) for switching onto or off of the RTE _for entries with existing text_ (either WIP or on an edit), because it WILL reformat your HTML and possibly change its behavior, and will NOT attempt perfect 1-1 round-trips with casual HTML. I still support that plan. 

#### How about that comment form layout? 

I want to replace the current "quote" button with formatting buttons like you see here on GitHub -- bold, italic, blockquote, link. They'll output either HTML or Markdown depending on the editor's current state. 

That should improve the balance of the layout, and is few enough buttons that I should be able to keep it from overflowing on mobile. 

Why bother adding a "broke text editor" like that? Well, basically because typing angle brackets and straight-quotes on your telephone completely sucks. Markdown's better with its asterisks and underscores, but still not fantastic. Just adding the four main types of formatting will make things......... so much nicer. 

Anyway, WIP.

htdocs/scss/components/quick-reply.scss
```diff
@@ -183,9 +183,12 @@
...
         }
-        #comment-text-quote {
-            margin-left: 5px;
-        }
+    }
+
+    .qr-markup {
+        display: flex;
+        justify-content: space-between;
+        align-items: flex-end;
```
nfagerlund, line 191:
```suggestion
        align-items: flex-end;
        
        & > div {
            display: inline-block;
        }
```

probably ought to throw this in there for non-flexy browsers, so they'll still display side-by-side.

htdocs/scss/pages/entry/new.scss
```diff
@@ -200,6 +200,16 @@
...
 }
 
+.markup-container {
+    display: flex;
+    align-items: center;
+
+    img { // help icon
+        margin-bottom: 1rem;
+        margin-left: 1rem;
+    }
```
nfagerlund, line 210:
```suggestion
    }
    
    & > * {
        display: inline-block;
    }
```

nfagerlund:
Note to self: I finally got my persistent spurious test failures problem dealt with (welllllll almost, still no idea wtf's going on with usermoves.t), and this PR definitely causes failures in post.t. They seem to be test-only problems; it's all junk like:

```
#   Failed test 'decoded entry form for entry with custom security'
#   at t/post.t line 435.
#     Structures begin differing at:
#          $got->{props}{editor} = ''
#     $expected->{props}{editor} = Does not exist
```

But, gotta update the tests anyway.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -114,13 +114,13 @@ sub clean {
     my $remove_sizes            = $opts->{remove_sizes} || 0;
```
zorkian, line 114: Interesting change. I had to ponder it for a bit, but I am on board.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1845,6 +1900,54 @@ sub quote_html {
...
+        at_mentions => 1,
+        noautolinks => 1,
+    },
+);
+$markup_formats{markdown} = $markup_formats{markdown0};
+$markup_formats{html}     = $markup_formats{html_casual1};
+
+# Return a group of arguments to pass to clean() based on the requested format
+sub formatting_args {
+    my $format = shift;
```
zorkian, line 1943: nit: use `$_[0]` for performance reasons (shift does an array modification). Ok, this is a mega-mega-ultra-nit, but I figured I'd mention since it's just a good keyboard habit to have.

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1845,6 +1900,54 @@ sub quote_html {
...
+    my $args   = $markup_formats{$format};
+    unless ( ref $args ) {
+        $args = $markup_formats{html_casual1};
+    }
+    return %$args;
```
zorkian, line 1948: Will anything anywhere blow up if we get an invalid format somehow? I guess, is there any chance someone can forge an editor format and get default/unplanned behavior and then exploit that?

cgi-bin/LJ/Talk.pm
```diff
@@ -2885,18 +2886,20 @@ sub prepare_and_validate_comment {
...
+        u           => $commenter,
+        parent      => $parent,
+        entry       => $entry,
+        subject     => $subject,
+        body        => $body,
+        unknown8bit => 0,
+        subjecticon => $subjecticon,
+
+        # TODO need a more organized way to carry approved props forward.
+        editor          => DW::Formats::validate( $content->{'prop_editor'} ),
```
zorkian, line 2898: Ah this answers my question earlier, okay!

zorkian:
Code LGTM. Not merging yet, pending discussion on dw_dev, and the auto-updating-default-preference thing. I think that's good feedback.

cgi-bin/LJ/CleanHTML.pm
nfagerlund: ah, cool! I never thought that through all the way. ("the more you know" star flies across the background) 

```suggestion
    my $format = $_[0];
```

cgi-bin/LJ/CleanHTML.pm
nfagerlund: Right, so there's two levels where we might care about that:

- Sending true garbage in your form in order to put garbage into the editor prop. Like you saw below, we're normalizing unknown to empty before inserting, so hopefully that's fine. 
- Setting a known format that you shouldn't be able to pick from the normal form, in order to make a post that uses a non-selectable format. I'm actually not doing anything to prevent that, and since it's a known format, validate would allow it through into the persisted editor prop. But the failure states of this seemed totally benign to me -- after all, if it's a known format, we're already displaying _some_ user-provided content in that format on purpose. So people could hack their form to keep posting in casual0 (maybe even use a browser extension to do it persistently), but it wouldn't have any worse effect on the site than someone editing an old post that's already allowed to stay in casual0. (Unless we kept an obsolete format that had _actual security holes,_ which, let's not.)

cgi-bin/DW/Controller/Editor.pm
```diff
@@ -0,0 +1,119 @@
...
+# Authors:
+#   Nick Fagerlund <nick.fagerlund@gmail.com>
+#
+# Copyright (c) 2020 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+package DW::Controller::Editor;
```
zorkian, line 15: nit: I know we're all over the map on this, but I think best practice now is to have a preamble like:

```perl
use v5.10;
use strict;
```

cgi-bin/DW/Controller/Editor.pm
```diff
@@ -0,0 +1,119 @@
...
+    return ( 1, $new_editor );
+}
+
+# zero-javascript version
+sub default_editor_handler {
+    my ( $ok, $rv ) = controller( form_auth => 1 );
+    return $rv unless $ok;
+
+    my $r      = $rv->{r};
+    my $POST   = $r->post_args;
```
zorkian, line 64: You will need to check `$r->did_post` to ensure that the caller actually did `POST` and return something valid if they didn't. That doesn't come for free in the standard handler. (It does in the RPC handler, when you specify that only `POST` methods are valid...)

cgi-bin/DW/Controller/Editor.pm
```diff
@@ -0,0 +1,119 @@
...
+        }
+    );
+}
+
+sub default_editor_rpc_handler {
+    my $r    = DW::Request->get;
+    my $POST = $r->post_args;
+
+    # make sure we have a user of some sort
+    my $remote = LJ::get_remote();
```
zorkian, line 96: I know this is the pattern of the other RPC endpoints, but it's weird that they don't use the `controller(...)` flow. Just noting for myself...

nfagerlund:
@zorkian Ah ha, thank you for catching the did_post thing! I totally forgot about that. (Have never added a controller from scratch, I think.)

zorkian:
HERE. WE. GO!

PR #2687, opened 2020-06-04
Tweak `@mention` scanning regexps to fix end-of-sentence bug

nfagerlund:
Prior to this patch, `@mentions` went Bad if they occurred before a
sentence-ending period -- mentions that used a shortcut site name (like `.ao3`)
just exploded, and other mentions would work but swallowed the sentence-ending
period.

This tweaks the regexps a bit:

- Cleanup: `\w` implies both `\d` and `_`, so they're not needed.
- Cleanup: `-` has special meaning depending on where it occurs within the
character class square brackets, so it's better to always escape it.
- Cleanup: hostnames can have hyphens.
- Bug fix: Final character of the matched @mention must not be a period.

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1902,7 +1902,7 @@ sub convert_user_mentions {
 
     # First pass is just to look for an edge case where an unescaped
     # username that needs to be converted is the first item in the string.
-    $$ref =~ s!^(\@([\w\d_-]+)(?:\.([\w\d\.]+))?)(?=$|\W)!$usertag->($1, $2, $3)!mge;
+    $$ref =~ s!^(\@([\w\-]+)(?:\.([\w\-\.]*[\w\-]))?)(?=$|\W)!$usertag->($1, $2, $3)!mge;
 
     # Second pass is to look for all other occurrences of unescaped usernames.
     # If we find an escaped username, remove the escape sequence and continue.
@@ -1910,7 +1910,7 @@ sub convert_user_mentions {
     # sequences here, to avoid parsing edge cases like '\\@foo' incorrectly
     # (note that's two user-supplied backslashes).  That's why the (\\.) case is
     # actually (\\.) and not (\\\@).
-    $$ref =~ s!(\\.)|(?<=[^\w/])(\@([\w\d_-]+)(?:\.([\w\d\.]+))?)(?=$|\W)!
+    $$ref =~ s!(\\.)|(?<=[^\w/])(\@([\w\-]+)(?:\.([\w\-\.]*[\w\-]))?)(?=$|\W)!
         defined($1) ? ( $1 eq '\@' ? '@' : $1 ) : $usertag->($2, $3, $4)
         !mge;
 }
--- a/t/cleaner-markdown.t
+++ b/t/cleaner-markdown.t
@@ -17,7 +17,7 @@
 use strict;
 use warnings;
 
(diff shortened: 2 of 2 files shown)


nfagerlund:
Oh wait, I think I'm missing a case here. Lemme check.

nfagerlund:
Hmm, now that I think about it, should the final character disallow hyphens too? I think that's an illegal way to end a hostname, and someone could conceivably want to hyphenate two usernames together...

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1902,15 +1902,15 @@ sub convert_user_mentions {
 
     # First pass is just to look for an edge case where an unescaped
     # username that needs to be converted is the first item in the string.
-    $$ref =~ s!^(\@([\w\d_-]+)(?:\.([\w\d\.]+))?)(?=$|\W)!$usertag->($1, $2, $3)!mge;
+    $$ref =~ s!^(\@([\w\-]+)(?:\.([\w\-\.]*[\w\-]))?)(?=$|\W)!$usertag->($1, $2, $3)!mge;
```
nfagerlund, line 1905:
```suggestion
    $$ref =~ s!^(\@([\w\-]+)(?:\.([\w\-\.]*\w))?)(?=$|\W)!$usertag->($1, $2, $3)!mge;
```

cgi-bin/LJ/CleanHTML.pm
```diff
@@ -1902,15 +1902,15 @@ sub convert_user_mentions {
...
 
     # Second pass is to look for all other occurrences of unescaped usernames.
     # If we find an escaped username, remove the escape sequence and continue.
     # We also have to look for (and explicitly ignore) Markdown-supported escape
     # sequences here, to avoid parsing edge cases like '\\@foo' incorrectly
     # (note that's two user-supplied backslashes).  That's why the (\\.) case is
     # actually (\\.) and not (\\\@).
-    $$ref =~ s!(\\.)|(?<=[^\w/])(\@([\w\d_-]+)(?:\.([\w\d\.]+))?)(?=$|\W)!
+    $$ref =~ s!(\\.)|(?<=[^\w/])(\@([\w\-]+)(?:\.([\w\-\.]*[\w\-]))?)(?=$|\W)!
```
nfagerlund, line 1913:
```suggestion
    $$ref =~ s!(\\.)|(?<=[^\w/])(\@([\w\-]+)(?:\.([\w\-\.]*\w))?)(?=$|\W)!
```

nfagerlund:
On _further_ further thought, no, I don't think we need to fuss with that. For one thing, I don't know whether it's possible for USERNAMES to end in hyphens, and I don't want to create a situation where you can hyphenate two foreign usernames together but not site usernames. For another, I seriously don't think anyone cares; anyone who sees a glitch due to a trailing hyphen is going to just find some other way to express whatever they meant to express, whereas the end of sentence thing is legitimately frustrating.

nfagerlund:
Adding some tests too, while I'm at it. (Makes a merge conflict with the formats PR, but I want feedback on the implementation of that one anyway.)

nfagerlund:
Hmm, sure! Only reason not to would be if we don’t have any hyphenated sites in the list. One sec.

nfagerlund:
HILARIOUSLY, we don't support ANY external sites with hyphens in their hostnames, cf. the list in DW::External::Site. Which means you won't be able to properly get a external user object to call ljuser_display on. 

Maybe I can fake up a site from within the test...

PR #2689, opened 2020-06-05
Fix failing comment-create.t

nfagerlund:
Thank you to @kareila for catching this! My bad. 

- First problem: Legit!! I forgot to add an early exit to LJ::Comment::create
when it's called without a ditemid. Previously, this would blow up slightly
later during LJ::Talk::Post::init and then create would return a generic error,
but because LJ::Entry::new explodes without it, we need to handle it earlier.
Could result in uninformative errors in both LJ::Comment::create and the Talk
controller.

- Second problem: Annoying. The test is looking for specific error codes, which
means if you give it input that's illegal in MULTIPLE ways, you're really
testing the ORDER in which it checks things, which I struggle to believe anyone
actually cares about. Well, oh well.


zorkian:
I don't think order of error cases _generally_ matters, either, FWIW.

PR #2693, opened 2020-06-09
hush warnings in /support/faqbrowse

kareila:
Some of these were due to the viewed FAQ having no category defined, but some were just debug warnings that escaped into the wild, and I had to go back to the original BML file to see what the value of $backfaqcat was supposed to be.

I replaced BML::get_language with the mostly-equivalent LJ::Lang::get_effective_lang, but I didn't see an LJ::Lang equivalent for BML::get_language_default.

### DIFF
--- a/cgi-bin/DW/Controller/Support/Faq.pm
+++ b/cgi-bin/DW/Controller/Support/Faq.pm
@@ -189,10 +189,8 @@ sub faqbrowse_handler {
     }
 
     # get language settings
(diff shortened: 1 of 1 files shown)


zorkian:
Hm @kareila this failed tests, but I'm not sure why. It's complaining about something in the REST module? but that doesn't seem to have been touched at all here.

kareila:
That’s the thing where we had to pin the version of the JSON validator. It
hasn’t rerun tests since we pinned that module.

PR #2699, opened 2020-06-17
TT conversion of shop pages, part 1

kareila:
index.bml was just a redirect to /shop so I removed the BML page and added the redirect to redirect.dat.

gco_notify.bml and pp_notify.bml were callback methods so I did a straightforward translation of the code but didn't have a way to test them.

creditcard.bml was intended to direct customers to enter their credit card info on Paypal's site. I made sure the template (named legacy/pp_button.tt) displayed correctly using a fake cart.

entercc.bml and creditcard_wait.bml were the pages we used for credit card sales before we moved to Stripe. Again, I was able to test the template display using a fake cart, and tested the various error checks for the user input to make sure I hadn't introduced any bugs there.

I think these are all of the "obsolete" shop pages. I'll try to start working on the ones we actually still use in the next batch.


zorkian:
Going to merge this so we have it in history, but maybe we should consider actually removing these and just letting anybody who wants to do a Paypal integration go dig them up from history?

Basically, I am thinking that if we're not actively using it, there's always some chance of some issue with the code still sitting around... I can't _see_ any problems, but not sure that won't always be true.

PR #2700, opened 2020-06-18
convert /lostinfo

kareila:
I lumped this in the controller with other related pages (accountstatus, changepassword) and reorganized the text strings into three different files, because I felt the need to tame the chaos.

I made a few minor adjustments to the text strings. There's no current DW::Request equivalent of BML::get_remote_host, so I dropped that and just printed the IP address.

I had to use the trick again of replacing `$()` with `document.getElementById()` because JS is cranky.

I wasn't able to test this with a working CAPTCHA config, but I don't expect that will make much of a difference.


views/settings/lostinfo.tt
```diff
@@ -0,0 +1,74 @@
...
+                   onkeyup = 'enable_user();', onchange = 'enable_user();' ) %]
+
+  <br />
+
+  [% form.submit( name = 'lostuser', id = 'lostuser',
+                  value = dw.ml( '.btn.proceed' ) ) %]
+
+  <script type="text/javascript">
+  function enable_pass () {
+      document.getElementById("lostpass").disabled =
```
nfagerlund, line 62: 👍🏼 

IDK if this came out clearly last time around, but this IS a one-to-one translation of the old JS. The old sixapart libraries implemented their own `$()` function that, even though it _looks_ like the jQuery thingy, really just does `getElementById` under the hood. Presumably this is part of why old 6alib pages don't play nice w/ jq.

I suppose we could also update these to jQuery calls, but then we'd also need to tweak the subsequent method calls (since jQuery doesn't return bare DOM elements like `getElementById` does) and would also need to split the JS out into separate files because mid-page jQuery is a no-no on Foundation pages. So this is the move unless you need to update the JS for other reasons.

PR #2701, opened 2020-06-19
convert /site/index

kareila:
This was so close to being a static page.

I'm a bit annoyed that the `<user name=...>` markup doesn't work on bare pages, but its BML equivalent does.


kareila:
Now with bonus Foundation / flexible column styling from the unstoppable @nfagerlund!

PR #2702, opened 2020-06-22
Fix misaligned site logo

nfagerlund:
I commented out Foundation's default image CSS because it was inappropriate for
a world where we display entries and comments in the foundationized site skins.
But the site logo was relying on that default `vertical-align: middle` rule, and
floated up a centimeter or so once it was gone. Didn't break anything, but sure
looked gross.


nfagerlund:
looked like this: 

image

PR #2706, opened 2020-06-24
Machete common code  🔪

kareila:
I was wandering around some dusty places and found a bunch of stuff that looked like it didn't need to be there any more. But I haven't verified these changes any further than just checking for compile errors, so it may need a closer look.

### DIFF
--- a/cgi-bin/LJ/S2/ReplyPage.pm
+++ b/cgi-bin/LJ/S2/ReplyPage.pm
@@ -57,7 +57,9 @@ sub ReplyPage {
     # canonical link to the entry or comment thread
     $p->{head_content} .= LJ::canonical_link( $entry->url, $replytoid );
 
-    $p->{'head_content'} .= $LJ::COMMON_CODE{'chalresp_js'};   # Hey, I think this is destroyed? -NF
+    $p->{'head_content'} .= qq{
+<script type="text/javascript" src="$LJ::JSPREFIX/md5.js"></script>
+};
 
     LJ::need_res( LJ::S2::tracking_popup_js() );
 
--- a/cgi-bin/bml/scheme/global.look
+++ b/cgi-bin/bml/scheme/global.look
@@ -13,15 +13,11 @@
 
 _parent=>../../lj-bml-blocks.pl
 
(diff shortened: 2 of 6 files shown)


cgi-bin/LJ/S2/ReplyPage.pm
```diff
@@ -57,7 +57,9 @@ sub ReplyPage {
     # canonical link to the entry or comment thread
     $p->{head_content} .= LJ::canonical_link( $entry->url, $replytoid );
 
-    $p->{'head_content'} .= $LJ::COMMON_CODE{'chalresp_js'};   # Hey, I think this is destroyed? -NF
```
nfagerlund, line 60: Lol thank you for dropping the other shoe. 

But I'm like 98% sure MD5 can be removed entirely from ReplyPage; there's no case in which it will get used. @momijizukamori, want to chime in?

htdocs/update.bml
```diff
@@ -663,7 +662,11 @@ head<=
     //]]>
     </script>^;
 
-    $ret .= (! $LJ::REQ_HEAD_HAS{'chalresp_js'}++) ? $LJ::COMMON_CODE{'chalresp_js'} : "";
+    my $chalresp_js = qq{
+<script type="text/javascript" src="$LJ::JSPREFIX/md5.js"></script>
+};
+
+    $ret .= (! $LJ::REQ_HEAD_HAS{'chalresp_js'}++) ? $chalresp_js : "";
```
nfagerlund, line 669: This one, otoh, MIGHT be used for crossposting. I don't understand that flow well enough to be sure. 

Sidenote, this "check for zeroness while mutating as a side effect, also you're inside a ternary" thing is... sure something.

cgi-bin/LJ/S2/ReplyPage.pm
kareila: Thanks for taking a look! My understanding of JS is shaky at best so I didn't want to make any assumptions, but I had a sense that might be the case. Of course right now it's doing exactly what it was before, only more obviously.

cgi-bin/bml/scheme/global.look
```diff
@@ -13,165 +13,25 @@
```
zorkian, line 13: Oh wow global.look it's been a _hot minute_ since I've peered behind these particular curtains.

PR #2709, opened 2020-06-25
test html cleaner with minimal external dependencies

martindemello:
Proof of concept mocking out a few functions and testing LJ::CleanHTML without pulling in all of ljlib.pl. It's a mishmash of stubs and mocks right now, but if people like the general idea I would love to get this cleaned up and merged, and then see what other modules could be similarly tested.


cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
+package LJ;
+
+use strict;
+no warnings 'uninitialized';
+
+
+BEGIN {
+    use Test::MockObject;
```
zorkian, line 8: You'll need to add any new dependencies to the `doc/dependencies-cpanm` file. That will auto-install them in the various environments. This will fix the first failure in the build.

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+
+    my $mock = Test::MockObject->new();
+
+    sub fake_lang_ml {
+	my ( $code, $vars ) = @_;
```
zorkian, line 13: Looks like this file never got tidied, this is the second failure in the build! You can do this from your local installation by doing something like:

`perl -I$LJHOME/extlib/lib/perl5 $LJHOME/extlib/bin/tidyall -a`

(Tidy is basically a Perl formatter we use, to ensure that code stays readable!)

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+	my $aopts = $vars->{'aopts'};
+	if ( $code == "cleanhtml.error.markup.extra" ) {
```
zorkian, line 15: Unfortunately, Perl uses different comparators for string vs numerical comparison. You'll want to use `eq` here.

See https://perldoc.perl.org/perlop.html#Equality-Operators for more information.

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
+package LJ;
+
+use strict;
```
zorkian, line 3: Above this please throw a,

`use v5.10;`

This won't have much practical impact on this particular file, but it clearly communicates to the reader what Perl semantics to expect in this file, in essence. We use this in all new files we're creating for the past few years.

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+    my $mock = Test::MockObject->new();
+
+    sub fake_lang_ml {
+	my ( $code, $vars ) = @_;
+	my $aopts = $vars->{'aopts'};
+	if ( $code == "cleanhtml.error.markup.extra" ) {
+	    return "[<strong>Error:</strong> Irreparable invalid markup ('&lt;" . $aopts . "&gt;') in entry. Owner must fix manually. Raw contents below.]";
+	}
+	return "[[[" . $code . "]]]". $aopts;
+	if ( $code == "cleanhtml.error.markup" ) {
```
zorkian, line 19: Same `eq` comment here.

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+BEGIN {
+    use Test::MockObject;
+
+    my $mock = Test::MockObject->new();
+
+    sub fake_lang_ml {
+	my ( $code, $vars ) = @_;
+	my $aopts = $vars->{'aopts'};
+	if ( $code == "cleanhtml.error.markup.extra" ) {
+	    return "[<strong>Error:</strong> Irreparable invalid markup ('&lt;" . $aopts . "&gt;') in entry. Owner must fix manually. Raw contents below.]";
```
zorkian, line 16: nit: You don't need to break out of the string here, if you're using double-quotes, Perl will do variable interpolation for you so you can just drop the `$aopts` into the string.

```perl
return "[<strong>Error:</strong> Irreparable invalid markup ('&lt;$aopts&gt;') in entry. Owner must fix manually. Raw contents below.]";
```

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+	    ml => \&fake_lang_ml 
+	));
+    $mock->fake_module('DW::Proxy' => (
+	    get_proxy_url => \&fake_get_proxy_url
+	));
+    $mock->fake_module('LJ::EmbedModule' => ());
+
+    # ugly hack to shutup dependent libraries which sometimes want to bring in
+    # ljlib.pl (via require, ick!).  so this lets them know if it's recursive.
+    # we REALLY need to move the rest of this crap to .pm files.
```
zorkian, line 38: Yeah :)

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
...
+
+    use lib ( $LJ::HOME || $ENV{LJHOME} ) . "/extlib/lib/perl5";
+
+    # Please do not change this to "LJ::Directories"
+    require $LJ::HOME . "/cgi-bin/LJ/Directories.pm";
+}
+
+
+sub is_web_context {
+    return $ENV{MOD_PERL} ? 1 : 0;
```
zorkian, line 53: Is this ever possibly true in your anticipated usage of `ljtestlib.pl`?

cgi-bin/ljtestlib.pl
```diff
@@ -0,0 +1,83 @@
+package LJ;
```
zorkian, line 1: This is a neat idea, but what do you think about putting this file in the `t/` directory? I think my concern is having "test libraries" in the same directory as production libraries, and wondering if that can cause problems down the road.

martindemello:
Thanks for the detailed review! This PR definitely needs a lot of work to fit in with the project structure (and, as you note, with basic perl - I admit I haven't used the language since 2001, so I have a lot to catch back up with!), but I wanted to do a quick spike and make sure people were okay with the basic idea before putting in the work to polish it up.

My primary motivation is to isolate those parts of the dreamwidth code that just do text transformations, and let them be tested as pure libraries. I can picture a number of benefits from this, particularly for new contributors - it would let people work on smaller parts of the code without having to have a complete dw setup, it would make the boundaries and connections between various subsystems easier to see, and it would greatly lower the barrier to entry for being able to make a small change and understand what it breaks just by reading the tests. From a code maintenance point of view, if e.g. HTMLCleaner was tested, with good coverage, using a mock User object, we could be reasonably sure that it was not coupled to the user account code in unexpected ways. 

It also makes the tests themselves easier to maintain - e.g. this PR has subtly broken the "span ljuser with junk inside" test, and I have no idea why. I'm hoping that with an explicit mock user object with all relevant properties right there in the code it will be easier to trace down and reason about just the bits that `HTMLCleaner` is doing, as opposed to the bits that e.g. `ljuser()` and `ljuser_display()` are doing, which right now all go into one big expected HTML string.

Also in my experience having tests that are lightweight, fast to run and easy to debug encourages people to write lots more of them :) That said, I definitely take your point about building up a detailed suite of mocks that goes out of sync with the code; I'm hoping to keep the mocking to a minimum (and indeed to move bits of it out of `ljtestlib.pl` and into individual tests, so that we could do things like set up a mock user with the properties that a single test needs right before calling the test, rather than have to jump back and forth from the mock definition to the code that uses it).

cgi-bin/ljtestlib.pl
martindemello: no, you're right, i should just return `0`

cgi-bin/ljtestlib.pl
martindemello: agreed; I'll do that.

cgi-bin/ljtestlib.pl
martindemello: i'm not even sure how this works to be honest :) i copied it wholesale from ljlib.

cgi-bin/ljtestlib.pl
martindemello: done

martindemello: done

cgi-bin/ljtestlib.pl
martindemello: done

cgi-bin/ljtestlib.pl
martindemello: done

cgi-bin/ljtestlib.pl
martindemello: done

cgi-bin/ljtestlib.pl
martindemello: done

cgi-bin/ljtestlib.pl
martindemello: done

kareila:
Encouraging decoupling is definitely a good code philosophy.

I was curious and ran `prove t/cleaner*` on this branch and my current branch - there's already a significant time savings (13 seconds vs. 23 seconds).

@martindemello: Please add Test::MockObject to `doc/dependencies-cpanm` so we know that module needs to be installed for everything to work.

Also, have you submitted a CLA yet? https://wiki.dreamwidth.net/wiki/index.php/CLA

martindemello:
done! i just emailed a signed CLA too

zorkian:
Well, this is awesome! I need to wait for @rahaeli to confirm the CLA before I do a final review & merge, but I think this is a great direction and particularly for these kinds of tests that won't ever/shouldn't ever pull in database/user/etc logic (mostly).

rahaeli:
I got Martin's CLA and have sent him a team invite; merge at will!

t/lib/ljtestlib.pl
```diff
@@ -0,0 +1,92 @@
...
+sub ljuser {
+    my ( $user, $opts ) = @_;
+    return
+          "<span style='white-space: nowrap;'>"
+        . "<a href='http://lj.example/userprofile'><img src='http://lj.example/img.png' alt='[alttext]' width='60' height='80'"
+        . " style='vertical-align: text-bottom; border: 0; padding-right: 1px;' /></a>"
+        . "<a href='http://lj.example/url'>$user</a></span>";
+}
+
+sub temp_user {
```
kareila, line 54: For this particular function, since you're not overriding a LJ::temp_user function in the main code, but rather providing an alternative to LJ::Test::temp_user, I think it would be less confusing if you put it in a different package like LJ::Mock instead of the main LJ package.

That's my only other suggestion for improvement - great work!

martindemello: done. please take a look - i couldn't do it other than by splitting into a new file and then adding the directory to `@INC`, not sure if that's the correct way

kareila:
I was thinking that you could do it just by adding one line above the
test_user function declaration in the original file:

package LJ::Mock;

Try that and see if it’s any easier?

martindemello:
neat, didn't know you could have multiple packages in the same file!

kareila:
@zorkian I think this is ready to merge, just make sure you squash the branch when merging?

zorkian:
Thank you both so much! Awesome :)

PR #2712, opened 2020-06-28
Finish removing spellcheck from non-legacy pages  🔪

### LINKED ISSUE #2481: Remove spellcheck support
- Even the worst desktop browsers have had spellcheck in textareas for the last ten years at least, and mobile platforms all have spellcheck in their keyboards. DW's preview-spellcheck doesn't seem to actually add any value over those. 
- Maintaining a custom dictionary is a fool's errand. 
- We don't support other languages, whereas platform spellchecks do.
- The current spellcheck behavior is bad. (Issues with recognizing contractions, other problems.)
- The current spellcheck code sounds like it might be remarkably bad.
- That valuable UI space can be used for better things.

This came out of #2480, and was the first part to get maintainer consensus in the discussion.

kareila:
This pulls spellcheck elements off of the beta Create Entries page and cleans up one remaining setup check from the rendering of journal/quickreply.tt.

I think spellcheck should stay intact on update.bml and editjournal.bml until we launch those into the sun. If we agree that should be the case, this closes #2481. 

I also discovered that the entry/new.js script was calling an undefined hashpassword function when the user was posting while logged out, so I removed that call as well.

### DIFF
--- a/htdocs/js/pages/entry/new.js
+++ b/htdocs/js/pages/entry/new.js
@@ -188,7 +188,7 @@ var postForm = (function($) {
             } else {
                 $custom_edit_button.hide();
             }
-        });
+        }).triggerHandler("change", true);
 
         // update the list of people who can see the entry
         function updatePostingMembers(e, useCached) {
@@ -314,7 +314,8 @@ var postForm = (function($) {
                 $security.append(opts.join(""));
 
                 // select the minsecurity value and disable the values with lesser security
-                $security.val(rank[oldval] >= rank[data.ret['minsecurity']] ? oldval : data.ret['minsecurity']);
+                var minsecrank = rank[data.ret['minsecurity']] || 0;
+                $security.val(rank[oldval] >= minsecrank ? oldval : data.ret['minsecurity']);
                 if ( data.ret['minsecurity'] == 'access' ) {
                     $security.find("option[value='public']").prop("disabled", true);
                 } else if ( data.ret['minsecurity'] == 'private' ) {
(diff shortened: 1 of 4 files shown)


nfagerlund:
Rad. 

I agree about leaving obsoleted BML stuff untouched. 👍🏼

htdocs/js/pages/entry/new.js
```diff
@@ -195,7 +188,7 @@ var postForm = (function($) {
             } else {
                 $custom_edit_button.hide();
             }
-        }).triggerHandler("change", rememberInitialValue);
+        });
```
kareila, line 191: I think this introduces a bug in the JS. When testing, I noticed that if you are posting while logged out, and you:

1. select a non-public entry security, then:
2. type the name of the journal to post in,
3. the security selection will jump back to public as soon as you tab out of the journal field.

When I tried the same trick on the live site, the security dropdown stayed where it was.

I suspect this line is where I've gone wrong. Does anyone know why it depended on opts.spellcheck before? What should the correct value be here?

kareila:
After spending some time bending my brain from Perl shape into jQuery shape, I decided that the bug I found was unrelated, and this fixes it in my testing: if the account had no minsecurity defined, the menu wouldn't retain the current selection. The aha moment was when I realized the menu didn't flip if I typed in the name of a nonexistent account. Not sure why I couldn't reproduce on the live site, unless it was because I was comparing behavior across different browsers.

As for my original question, I still don't know why this option was linked to spellcheck, but logically, if we're replacing `var rememberInitialValue = !opts.spellcheck;` then that would always be a true value if we've removed spellcheck, so that's where I left it. Not going further down the rabbit hole than that.

zorkian:
Agreed on leaving the old ones alone, but I would like to actually get around to launching those into the sun... if we're not, there is some value in nuking the feature entirely (removes some config, and production/hack setup, I guess?) but since it's already set up I'm not too worried.

PR #2715, opened 2020-06-28
convert /admin/priv

kareila:
I just redid the legacy index changes so this one should be good to go next.

Of note: I edited errors.tt to allow a style argument.


zorkian:
This code is quite ancient, assembling SQL and `$dbh->quote` oh my. Thanks for tackling it!

PR #2718, opened 2020-06-28
[#1841] include priv packages on /admin/priv page

### LINKED ISSUE #1841: suggested improvement for /admin/priv: include packages in dropdown
So now that I'm in the priv-granting business, I found myself looking around at the differences between the /admin/priv page vs. what's available on the console.  The thing that immediately jumped out at me is that the console supports the creation and granting of "packages" of privs, which are not listed in the dropdown menu on the /admin/priv page.  I suggest including them at the top of that menu, prefaced with the # symbol as on the console to avoid any confusion.

(Required reading: the console commands for 'priv' and 'priv_package' and the page source for /admin/priv/index, which is sadly still a BML file.)

kareila:
When granting privileges to a particular user, in addition to the existing dropdown of privileges, I added a second dropdown list of defined priv packages. I decided I liked the look of it better, and it was easier to implement than if I'd smashed the two sets of options into one menu as I had originally proposed. (It was also easier to implement than if I'd tried to do this before encapsulating the logic for granting privileges in the controller.)

This is just for ease of use when granting packages. All other package operations, such as seeing which privileges are included in a package, still require use of the priv_package console command.

Fixes #1841.

### DIFF
--- a/cgi-bin/DW/Controller/Admin/Priv.pm
+++ b/cgi-bin/DW/Controller/Admin/Priv.pm
@@ -125,14 +125,14 @@ sub main_controller {
 
         # these actions are the same on both sides, just different data sources
         my $process_grant = sub {
-            my ( $user, $privid ) = @_;
+            my ( $user, $privid, $arg ) = @_;
 
             my $u = LJ::load_user($user);
             return error_ml("$scope.error.invaliduser") unless $u;
 
             my $privcode = $vars->{priv_by_id}->{$privid}->{privcode};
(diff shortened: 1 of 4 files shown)


kareila:
Oh hey I just remembered I saved a screenshot for this one!

zorkian:
I forgot that priv packages was a thing. Do we use those on DW?

kareila:
Yes we do have one defined for basic support privs. But if there aren’t any
defined the menu won’t show.

PR #2724, opened 2020-06-29
Return to original thread view after commenting/editing

### LINKED ISSUE #2717: possibly wrong redirect after posting from new comment preview when viewing a single thread
Several people have raised this in dw_maint, but I'll summarize it here:

Before the change, the return link argument was of the form `?thread=(######)&posted=1#cmt(######)` which just loaded the relevant thread you were commenting on. 

Now it returns to `?view=(######)&posted=1#cmt(######)` which loads the whole page from the top. 

If there are a lot of comments on the page, this can be a lot slower to load, in addition to losing your place on the page.

nfagerlund:
Fixes #2717

- The quick-reply has always (?) had a hidden `viewing_thread` field that keeps a record of which thread you were viewing when you hit "reply", and we try to redirect you there after posting. 
- The talkform has never had that, so if you used "more options", you'd redirect to the main entry page when done. Same for editing comments or opening a reply link in a new tab (or with no JS). 
- The old previewform, however, used to just copy over every form field it didn't recognize into a hidden input. So if you previewed FROM the quick-reply, you'd get the quick-reply behavior! 

So: the return-to-thread after preview behavior broke because I replaced the previewform with the talkform, which didn't know how to deal with the `viewing_thread` field and doesn't duplicate whatever random garbage you huck at it.

The fourth commit in this PR fixes the reported bug.

The first three commits are cleanups and docs. 

The fifth and sixth commits enable return-to-thread for edit links and reply links, which have never supported that before, but really why shouldn't they.

### DIFF
--- a/cgi-bin/DW/Controller/Talk.pm
+++ b/cgi-bin/DW/Controller/Talk.pm
@@ -109,8 +109,7 @@ DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 # - viewing_thread
 #     The filtered thread view (`?thread=12345`) they were in when they hit the
 #     "reply" link, and which they should be returned to once they finish posting.
-#     Currently only populated by the quick-reply. Consumed by Talk controller to
-#     build the return link.
+#     Consumed by Talk controller to build the return link.
 # - style/format/s2id/fallback
 #     The "viewing style" options (`?style=light`) that were in effect when they
 #     hit the "reply" link, which should be re-instated once they finish
--- a/cgi-bin/DW/Controller/Talk.pm
+++ b/cgi-bin/DW/Controller/Talk.pm
@@ -8,6 +8,154 @@ use Carp;
 
 DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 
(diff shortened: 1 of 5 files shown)


nfagerlund:
(BTW: If you want to fix this quickly and there are any questions about the approach in commits 5 and 6, you can manually merge just the first four! They don't depend on the last two, and they bring us up to parity with the previous behavior.)

rahaeli:
....that's kind of an amazing case of "bug becomes unintentional feature". (I agree that adding the return-to-thread behavior elsewhere is a good idea.)

cgi-bin/DW/Controller/Talk.pm
```diff
@@ -8,6 +8,154 @@ use Carp;
 
 DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 
+# I think this canon isn't written down anywhere else, so HWAET: I sing a record
+# of every damn form field that comes through in reply form POSTs. -NF
```
zorkian, line 12: oh my lord this comment T_T

this is why we can't have nice things. or, alternately: this is why the world is moving towards static typing as a necessity for building software at anything beyond toy-scale.

PR #2704, opened 2020-06-22
convert remaining stats pages

kareila:
This converts stats.bml to TT/Foundation, removes stats/index.bml which was just a redirect, and fixes some bad image links in /admin/stats which I noticed while poking around.

I think everything on the converted page displays okay? I didn't have a lot of data sitting around to test with.

The commit that touches redirect.dat may conflict with my shop pages PR (7068462 ) that touches the same file. Story of my life right now. If one gets merged, I can redo the other one.

PR #2727, opened 2020-06-29
fix tiny whitespace glitch on the converted stats page [#2704]

kareila:
Sorry I missed this, the line didn't display on my dreamhack.


kareila:
Also I just noticed that ljuser links were displaying with the same color regardless of whether they had been visited or not. This should fix that.

PR #2733, opened 2020-07-01
Comment preview: Display toggleable truncated parent above preview

nfagerlund:
- Only takes effect if JS is enabled. No-JS folks get the parent below the reply
form, the way it is today.
- Uses flex `order` to hopefully not put it in the way of screenreaders; they
should still hear it below the reply form, since that's the DOM order.
- Does a pretty little blur at the truncation point, for browsers that support it.

<details><summary>Screenshots</summary>

**Collapsed:**

Screenshot_2020-07-01 Comment Preview(3)

**Expanded:**

Screenshot_2020-07-01 Comment Preview(4)

**Collapsed (without s2foundation):**

Screenshot_2020-07-01 Comment Preview(1)

**Expanded (without s2foundation):**

Screenshot_2020-07-01 Comment Preview(2)

</details>

### DIFF
--- a/htdocs/js/jquery.talkpost.preview.js
+++ b/htdocs/js/jquery.talkpost.preview.js
@@ -1,0 +1,23 @@
+// Tiny script for hiding/revealing the full text of the comment/entry you're
+// replying to.
+jQuery(function($) {
+    "use strict";
+
+    var replyTo = $('#preview-parent-entry');
+    var toggleButtons = $('.js-parent-toggle');
+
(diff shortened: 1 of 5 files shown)


nfagerlund:
Just realized I should probably specifically @ @rahaeli here, because changing the layout of this page is sort of a design-and-culture question.

zorkian:
I think it looks fine, but yeah, let's wait for D.

rahaeli:
Looks good to me!

PR #2737, opened 2020-07-02
[#2735] Include "admin post" labels in new comment preview

### LINKED ISSUE #2735: new comment preview form doesn't include "admin post" metadata when previewing
https://www.dreamwidth.org/support/see_request?id=42154

> Hi, I'm just previewing a comment in my comm, and I checked the "admin post" box, but the little ribbon graphic and "as admin" text doesn't appear in the preview area. It'd be great if it showed up there so I don't keep checking to see if it's selected.

It sounds like this is a request for a new feature, as opposed to something that went missing during the conversion.

nfagerlund:
Fixes #2735 

I missed this the first time around! Admin post labels can appear on both the
WIP comment, and on the reply target.

I also missed the "in (journal)" label the first time around, so let's get that
in there too. Based on the siteviews S2 layout, that appears on both comments
and entries when they're posted somewhere other than the poster's own journal.
(Well, with comments that only appears when they're the target of a ReplyPage.
Entries get it on EntryPages as well.)

Screenshot_2020-07-01 Comment Preview(5)

### DIFF
--- a/cgi-bin/DW/Controller/Talk.pm
+++ b/cgi-bin/DW/Controller/Talk.pm
@@ -871,7 +871,7 @@ sub preview_parent_args {
             $poster      = $parentitem->poster->ljuser_display;
             $poster_name = $parentitem->poster->name_html;
 
-            if ( $parentitem->poster->user eq $entry->journal->user ) {
+            if ( $parentitem->poster->equals( $entry->journal ) ) {
                 $in_journal = '';
             }
         }
@@ -894,7 +894,7 @@ sub preview_parent_args {
         # Replying to entry
 
         my $in_journal =
-            $entry->poster->user eq $entry->journal->user ? '' : $entry->journal->ljuser_display;
+            $entry->poster->equals( $entry->journal ) ? '' : $entry->journal->ljuser_display;
 
         return {
             type        => 'entry',
--- a/cgi-bin/DW/Controller/Talk.pm
+++ b/cgi-bin/DW/Controller/Talk.pm
@@ -835,6 +835,7 @@ sub preview_comment_args {
         body        => $cleanbody,
         subject     => $cleansubject,
         icon        => $icon,
+        admin_post  => $comment->{admin_post},
     };
 
     return $preview;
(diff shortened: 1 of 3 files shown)


cgi-bin/DW/Controller/Talk.pm
```diff
@@ -864,9 +865,15 @@ sub preview_parent_args {
 
         my $poster      = 'Anonymous';
         my $poster_name = '';
+        my $in_journal  = $entry->journal->ljuser_display;
+
         if ( $parentitem->poster ) {
             $poster      = $parentitem->poster->ljuser_display;
             $poster_name = $parentitem->poster->name_html;
+
+            if ( $parentitem->poster->user eq $entry->journal->user ) {
```
kareila, line 874: nit: I think it's preferable to say `$parentitem->poster->equals($entry->journal)` here.

cgi-bin/DW/Controller/Talk.pm
```diff
@@ -875,20 +882,27 @@ sub preview_parent_args {
...
+            time        => $parentitem->{datepost},
             url         => $parentitem->url,
             entry_url   => $entry->url,
         };
     }
     else {
         # Replying to entry
+
+        my $in_journal =
+            $entry->poster->user eq $entry->journal->user ? '' : $entry->journal->ljuser_display;
```
kareila, line 897: Same as above, use the `equals` method.

views/talkpost_do/preview_comment.tt
```diff
@@ -51,6 +51,9 @@ the same terms as Perl itself.  For a copy of the license, please reference
                                 [% END %]
                                 <span class="poster comment-poster">
                                     [% comment.poster %]
+                                    [% IF comment.admin_post %]
+                                        <span>(<span class="admin-post-icon">[% dw.img('admin-post') %]</span> as admin)</span>
```
kareila, line 55: What is the reason for including an additional bare `<span>` here and in the similar block below? Does it serve a function other than syntactic grouping?

cgi-bin/DW/Controller/Talk.pm
nfagerlund: Ah, good call! I had a lingering suspicion that there was a method for that, but was having a hard time finding it for some reason.

views/talkpost_do/preview_comment.tt
nfagerlund: Great question, I have no idea! But for maintainability I want previews to use the exact same markup as real comments do,\* and [layout dot s2 has spoken](https://github.com/dreamwidth/dw-free/blob/194048b/styles/siteviews/layout.s2#L283-L287) ([myeeeeeah](https://www.youtube.com/watch?v=vNdl9NpCMOA)). 

Poking at the web inspector, it seems to be functionally irrelevant. 🤷 

\* (vague ambition to make the siteviews layout call a .tt template per comment instead of just printing lines so we can have a single path all markup goes through \[same for entry preview\], but that sounds hard so I haven't gotten to it yet.)

PR #2742, opened 2020-07-07
Improve robustness of concat_res and cachebuster params

nfagerlund:
- Allow empty `?v=` params without just exploding.
- Provide more useful `?v=` params by looking for files in the same place the real server does.
- Guard against file quirks that can trash things if you concatenate stuff without checking yourself.

### DIFF
--- a/cgi-bin/Apache/LiveJournal.pm
+++ b/cgi-bin/Apache/LiveJournal.pm
@@ -1238,7 +1238,19 @@ sub load_file_for_concat {
     { local $/ = undef; $contents = <FILE>; }
     close FILE;
 
-    return [ $contents, $stat[7], $stat[9], $mime ];
+    # Remove UTF-8 byte-order mark, in case someone set us up the BOM. (It's not
+    # valid after the first position in a file, and can screw up syntax when
+    # concatenating.)
+    $contents =~ s/\A\x{ef}\x{bb}\x{bf}//;
+
+    # Add a newline, juuuust to be safe.
+    $contents .= "\n";
+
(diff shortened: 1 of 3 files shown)


cgi-bin/Apache/LiveJournal.pm
```diff
@@ -221,7 +221,7 @@ sub send_concat_res_response {
 
     # Might contain cache buster "?v=3234234234" at the end;
     # plus possibly other unique args (caught by the .*)
-    $args =~ s/\?v=\d+.*$//;
+    $args =~ s/\?v=\d*.*$//;
```
alierak, line 224: Was this also supposed to allow ?v=non-digit-values? If so, the whole \d part is no longer needed.

nfagerlund: Well, I don't have any real use for specifying non-digit values. But good point -- if we're not actually USING the numbers for anything and are just stripping the params off to throw them away, then why check for them at all?

```suggestion
    $args =~ s/\?v=.*$//;
```

nfagerlund:
I tossed on a couple more commits of stuff that came up while investigating a build-static replacement. I originally wasn't gonna, but then it occurred to me to go check, and SURE ENOUGH, even our current Sass compiler on prod (ruby sass 3.4?) is tacking BOMs onto UTF-8 files when printing compact CSS! I haven't found anything specifically broken due to that yet, but I bet I could manage something eventually.

So anyway, that seems to be a common enough practice that we need to just accept it as a fact of life and adjust accordingly.

PR #2743, opened 2020-07-07
LJ::OpenID::consumer can't handle Hash::MultiValue inputs

kareila:
Oops I broke OpenID logins when I converted the page. Those tricky objects that pretend to be hashes!

### DIFF
--- a/cgi-bin/DW/Controller/OpenID.pm
+++ b/cgi-bin/DW/Controller/OpenID.pm
@@ -351,7 +351,9 @@ sub openid_login_handler {
         return $r->redirect($check_url);
     }
 
-    my $get_args = $r->get_args;
+    # get_args method returns a Hash::MultiValue,
+    # but LJ::OpenID::consumer needs a plain vanilla hash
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Controller/OpenID.pm
```diff
@@ -351,7 +351,9 @@ sub openid_login_handler {
         return $r->redirect($check_url);
     }
 
-    my $get_args = $r->get_args;
+    # get_args method returns a Hash::MultiValue,
+    # but LJ::OpenID::consumer needs a plain vanilla hash
+    my $get_args = { %{ $r->get_args } };
```
nfagerlund, line 356: Oh, I feel like I ran into this in the docs at some point. Do we want...

```suggestion
    my $get_args = $r->get_args->mixed;
```

...instead?

kareila: I had not seen that usage, thanks! I'll update. (Was going to update anyway to drop the warnings commit.)

PR #2745, opened 2020-07-08
Lay some mojibake traps around the inbox and front page

nfagerlund:
(Although, perhaps it's krakozyabry rather than mojibake, since it's been our Cyrillic users running into this.)

There's a persistent problem of mojibake on the front page and inbox _under
certain unknown conditions on prod that no devs have managed to reproduce yet._
There's early evidence that strings from the translation system might have been
causing _some_ of this, but there still seem to be some instances hanging
around.

If we can figure out which choke-points the contaminated strings are passing
through, we can maybe start following them back to their lair.


kareila:
unknown

PR #2748, opened 2020-07-09
convert /extcss/index.bml to DW/Controller/CSSProxy.pm

kareila:
I followed the example of DW::Controller::EmbeddedContent to figure out how to approach this. Once I got past the request setup changes, the rest of the code was basically identical from the original BML.

Has the world moved on from typepad.com shared stylesheets? I left the special treatment in, but it seemed outdated.


rahaeli:
The specialcasing is almost certainly an old 6A thing and can be nuked. The only surviving references to typepad.com other than this is a special case in the feed system amid many others and a link in some test  data.

rahaeli:
Yup, I looked it up: it was added to have a single source of CSS for the (ljcom) TypePad styles that were converted to S2 after the 6A acquisition. We didn't use any of LJ's styles at all, much less the proprietary ones, so it can go.

kareila:
Thanks for checking! I've now removed the special treatment.

PR #2754, opened 2020-07-12
[#2560] run entry excerpts through HTML cleaner instead of escaping everything

### LINKED ISSUE #2560: Manage Entries page shows raw HTML
Discovered this while working on the new RTE, which writes slightly more advanced (and supported) HTML into posts than the old RTE does:

MWSnap 2019-07-22, 00_13_04

We fully escape all of the content here via `LJ::ehtml()`:

https://github.com/dreamwidth/dw-free/blob/master/htdocs/editjournal.bml#L105-L107

Maybe we should instead strip all tags and just show any textual content, unless that's an empty string, in which case, show the escaped HTML instead.

This is lowish priority, but something should be done before RTE 2.0 comes out of beta...suggestions welcome.

kareila:
I took a wild stab at this - the options for the cleaner may need to be tweaked further, but this seems like a reasonable first attempt.

Fixes #2560.

### DIFF
--- a/htdocs/editjournal.bml
+++ b/htdocs/editjournal.bml
@@ -102,9 +102,30 @@ body<=
               }
             $ret .= "<br />\n";
 
-            my $event = LJ::ehtml(LJ::durl($res{"events_${i}_event"}));
-            $event =~ s!\n!<br />!g;
-            $ret .= $event;
+            if ( my $event = LJ::durl( $res{"events_${i}_event"} ) ) {
+
+                # use the cleaner to get an HTML-minimal excerpt
+                LJ::CleanHTML::clean( \$event,
+                    {
+                        textonly  => 1,
+                        addbreaks => 1,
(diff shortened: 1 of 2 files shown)


nfagerlund:
The way all these old BML pages call all these low-level functions constantly makes my skin crawl. 😖 

So, LJ::Entry objects already have an `event_html_summary` method -- would it make more sense to just grab some item IDs, make entry objects (`$u->recent_items`? at least that's what S2 does), and call that? 

The only problem I could think of would be performance concerns, but it seems like that shouldn't be any nastier than just building a last-N journal page, yeah? I checked to see how outrageous this list gets, and while the form will let you request up to 99, it seems to ignore that and actually cap it at 50.

kareila:
I'm happy to investigate that alternative - I'll flag this as having requested changes until I have time to try it out.

zorkian:
Resolved by #3502

PR #2756, opened 2020-07-12
new function LJ::CleanHTML::legacy_markdown

### LINKED ISSUE #2385: investigate possibility of excluding "!markdown" formatting marker from content search for "markdown"
https://www.dreamwidth.org/support/see_request?id=39403

### LINKED ISSUE #2419: make markdown indicator case-insensitive
Reported on Discord:

> My first two attempts at using Markdown failed because I used a capital M without having the reference right in front of my eyes (silly me, I thought it would be easy to remember...) so would it be reasonable to allow the variant !Markdown to do the same thing as !markdown?

Pretty sure this is just a regular expression that needs the case-insensitive modifier added.

kareila:
The first commit defines LJ::CleanHTML::legacy_markdown in order to have the regular expression defined in a single place. The regexp is now case-insensitive, which fixes #2419. I tested this and confirmed that using `!Markdown` with a capital M now works as expected.

The second commit I'm less sure about. It uses LJ::CleanHTML::legacy_markdown when fetching entry text for the sphinx index, so that searching for "markdown" won't hit every entry formatted with markdown using the old syntax. That fixes #2385, in theory. Unfortunately, this only removes the text at the time of indexing, so older results will still turn up unless they get reindexed. Furthermore, the markdown formatting will appear uninterpolated in the results, since we just stripped out the indicator, but that seems okay under the circumstances? I don't know, it's a tradeoff. I wasn't able to test this at all since I don't have Sphinx set up.

(Fun fact: searching for your favorite multicharacter HTML tag will match on every entry that uses that tag, since all the HTML content is indexed as well. It's a more general problem that might benefit from a more general solution.)

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -850,7 +850,7 @@ sub _backend_to_form {
     # except that if we're detecting old-style !markdown, we DO want to also
     # mutate the body text, which makes it hairy.
     unless ($editor) {
-        if ( $event =~ s/^\s*!markdown\s*\r?\n//s ) {
+        if ( LJ::CleanHTML::legacy_markdown( \$event ) ) {    # mutates $event
             $editor = 'markdown0';
         }
         elsif ( $entry->prop('opt_preformatted') ) {
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1646,7 +1646,7 @@ sub clean_event {
 
     # ...and if not, here's how we guess.
     if ( !$formatting ) {
-        if ( $$ref =~ s/^\s*!markdown\s*\r?\n//s ) {
+        if ( legacy_markdown($ref) ) {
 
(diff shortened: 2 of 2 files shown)


nfagerlund:
I'm 👍🏼 with using a function for this, but I'm trying to decide if there's a better name for it. I realize I'm a soft modern weakling, but functions that mutate their arguments AND return a value weird me out unless their names have a 🚨 in em. `remove_legacy_markdown_flag`? `enable_legacy_markdown`?

(Bleh, the Sphinx thing seems super squirrelly. Should we be... stripping ALL markup off content for indexing? Currently I don't think CleanHTML supports that -- clean takes a `textonly` opt, but it only affects a couple of things.)

kareila:
The other option that occurred to me was just making the regexp a globally available scalar to be used with either `m//` or `s///` in the calling context, but that seemed uglier?

kareila:
I pulled the Sphinx-related commit to look at doing a more general cleaning of all formatting content from the event text before indexing.

PR #2759, opened 2020-07-12
Fix comment-eating form_auth expiry bug

nfagerlund:
Controller's automatic form_auth handling happens too early, so it ends up
throwing away your WIP comment if your form expires, and then you're at the
mercy of your back button cache. To protect the text, we need to handle
form_auth manually after we have our POST args.

This was handled properly in the old talkpost_do.bml, but it broke during
the big update because I didn't understand the timing implications. (I guess it
technically broke in the initial conversion, but
de8ce76 was where I removed the old
implementation.)

I'm being a bit more conservative than the old version -- if form_auth fails,
we won't even try and validate your identity, because let's NOT do anything with
side-effects if we think something sketchy is going on. (Letting you start the
OpenID flow from a known-bad state seems especially dodgy.) And without your
identity, there's not much point in trying to validate your comment, so we'll
skip that too. Basically I'm treating form_auth failure as "slightly more fatal"
than all of the other negotiably-fatal errors that we let you recover from.

### DIFF
--- a/cgi-bin/DW/Controller/Talk.pm
+++ b/cgi-bin/DW/Controller/Talk.pm
@@ -92,8 +92,8 @@ DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 #
 # Consistency checks and metadata:
 # - lj_form_auth
-#     Form auth for logged-in users. I don't fully understand how this works yet,
-#     but plenty of others around here do, so go ask them.
+#     Consistency check for CSRF prevention. Similar idea to chrp1, but checks more
+#     stuff. See LJ::check_form_auth in Web.pm for more.
 # - chrp1
 #     A time-expiring server-provided token used to make spam posts inconvenient.
 # - captcha_type (and other captcha fields)
@@ -157,7 +157,11 @@ DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 #     listens for this by name.
 
 sub talkpost_do_handler {
-    my ( $ok, $rv ) = controller( form_auth => 1, anonymous => 1 );
+
+    # This route handles form_auth manually instead of passing form_auth => 1 to
+    # controller() -- we want to fail gently AFTER we have our post args, so we
+    # can regenerate the comment form and not lose your text.
+    my ( $ok, $rv ) = controller( anonymous => 1 );
     return $rv unless $ok;
 
     my $r      = $rv->{r};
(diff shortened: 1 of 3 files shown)


cgi-bin/DW/Controller/Talk.pm
```diff
@@ -157,7 +157,11 @@ DW::Routing->register_string( '/talkpost_do', \&talkpost_do_handler, app => 1 );
 #     listens for this by name.
 
 sub talkpost_do_handler {
-    my ( $ok, $rv ) = controller( form_auth => 1, anonymous => 1 );
+
+    # This route handles form_auth manually instead of passing form_auth => 1 to
+    # controller() -- we want to fail gently AFTER we have our post args, so we
+    # can regenerate the comment form and not lose your text.
+    my ( $ok, $rv ) = controller( anonymous => 1 );
```
kareila, line 164: We need to explicitly say `form_auth => 0` here to signal that we are doing our own auth handling. There was discussion of changing the default from 0 to 1 at some point.

cgi-bin/DW/Controller/Talk.pm
```diff
@@ -176,13 +180,30 @@ sub talkpost_do_handler {
...
+        # OpenID, we already checked earlier.
+        unless ( LJ::check_form_auth( $POST->{lj_form_auth} ) ) {
+
+            # Form_auth failure is extra sketchy, so let's remember this and
+            # avoid calling some other functions later.
+            $form_auth_ok = 0;
+            push @errors, LJ::Lang::ml('/talkpost_do.tt.error.invalidform');
+        }
+    }
+    elsif (( $GET->{'openid.mode'} eq 'id_res' || $GET->{'openid.mode'} eq 'cancel' )
```
kareila, line 201: I had submitted a patch to hush some warnings in this block, but since you just changed it, maybe go ahead and do it in this PR instead? https://github.com/dreamwidth/dw-free/pull/2743/commits/691c90694741cf73571496c9624ee830bba5296b

cgi-bin/DW/Controller/Talk.pm
nfagerlund: Ah, thank you! Didn't know that.

cgi-bin/DW/Controller/Talk.pm
nfagerlund: Good call, thanks for the heads-up. I'll try and rig it so you show up as the commit author.

cgi-bin/DW/Controller/Talk.pm
```diff
@@ -176,16 +179,39 @@ sub talkpost_do_handler {
     # For errors that aren't immediately fatal, collect them as we go and let
     # the user fix them all at once.
     my @errors;
+    my $form_auth_ok = 1;
```
zorkian, line 182: Might be a nit, but I always prefer these things to logically default to false, and then have a very explicit "this is when we turn it on" flow. Otherwise, we have to make sure to turn it off everywhere we need to, and that feels harder to maintain?

nfagerlund: I felt weird about doing it this way too, but it seemed like there was only one branch where it might ever turn off, so it'd end up more verbose. But if we both felt weird about it, might as well make it feel more normal. (Weird feels function as a DOS on the maintainer.)

zorkian: I don't feel *strongly* about it, if it really makes it that much more ugly. LMK.

nfagerlund: Made it smaller, even; go figure. :)

zorkian:
LGTM, ty!

PR #2772, opened 2020-07-16
Tests: Pin (some) @-mention tests to latest format versions

nfagerlund:
I just had one of my occasional visions of the future, is all.

We already have a format alias for "give me the newest Markdown" (to accommodate
email posts), so this commit updates the @-mentions tests to use that by
default.

This way if we switch Markdown processors to one that deliberately eats the
escaping backslash in front of punctuation (like @), we'll get a nice early
warning that we need to tweak something in the rendering path.

(The tests that use the default format by passing editor => undef should already be fine, because the default should remain "newest casual HTML.")

### DIFF
--- a/t/cleaner-markdown.t
+++ b/t/cleaner-markdown.t
@@ -128,7 +128,8 @@ sub check_doesnt_use_markdown {
 }
 
 # local content, converts users when not inside html
-check_doesnt_use_markdown( 'local entry made in default editor (newest casual HTML version)', editor => undef );
+check_doesnt_use_markdown( 'local entry made in default editor (newest casual HTML version)',
+    editor => undef );
 check_uses_markdown( 'local entry made in markdown editor', editor => 'markdown' );
 is( $clean->( '@system', editor => undef ),
     $lju_sys, 'user tag in plain text converted (undef editor)' );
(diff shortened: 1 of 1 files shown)


zorkian:
Nice brainstorm.

Alas tho, tidy fails. D:

kareila:
So with this commit in place, t/cleaner-markdown.t actually begins to fail on my hack, and just from reading the commit I have no idea why it would have this effect, but a lot of tests in the suite are suddenly missing surrounding paragraph tags? Example:
```
#   Failed test 'escaped user tag in plain text not converted, backslash removed'
#   at t/cleaner-markdown.t line 43.
#          got: '@system'
#     expected: '<p>@system</p>'
```

nfagerlund:
1. UGH, I did the dev thing I really really hate, where one just "kicks it from the head" and ships a change _without actually running it,_ in faith that one's mental model is accurate. (I could have sworn I ran that, but obviously no!) 😖 mea culpa
2. Nice nice nice, this totally revealed a bug in formats implementation. Not one you'd usually hit in real conditions, which is why I didn't notice it 'til now. I gave CleanHTML its own separate set of format aliases instead of using Formats::validate() to get canonical names, mostly just bc I implemented it first. Reason it's hard to make it a problem is that it only affects how it interprets the prop value that got stored in the backend, and the prop value gets sent through validate() before it gets stored. But still, incorrect (and also mea culpa).

PR #2774, opened 2020-07-19
give all cookies the secure attribute on HTTPS sites

kareila:
Momiji got a Firefox warning saying cookies with the 'sameSite' attribute set to 'none' or an invalid value would be rejected without the 'secure' attribute. This updates the add_cookie method to add the 'secure' attribute to all our cookies when LJ::PROTOCOL is https. The 'secure' attribute will cause the cookie not to be sent if the connection is not over SSL.

### DIFF
--- a/cgi-bin/DW/Request/Base.pm
+++ b/cgi-bin/DW/Request/Base.pm
@@ -82,6 +82,9 @@ sub add_cookie {
     confess "Must provide value (try delete_cookie if you really mean this)"
         unless exists $args{value};
 
+    # we need to give all cookies the secure attribute on HTTPS sites
+    $args{secure} = 1 if $LJ::PROTOCOL eq "https";
+
     # extraneous parenthesis inside map {} needed to force BLOCK mode map
     my $cookie = CGI::Cookie->new( map { ( "-$_" => $args{$_} ) } keys %args );
     $self->err_header_out_add( 'Set-Cookie' => $cookie );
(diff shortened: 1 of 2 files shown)


nfagerlund:
Hmm... wonder if that browser behavior change will mean dev servers are hosed unless they use HTTPS.

kareila:
Dev servers might be OK if we set Samesite to Lax, but I haven't been able to test that theory yet. Right now we don't set Samesite at all.

nfagerlund:
Well, fwiw I just tried setting SameSite=Lax on my hack, and it seems to be working so far. (The one wrinkle is that SameSite support was only added in CGI@4.29, and the hack server has 4.26 installed globally. But once I installed 4.50 for my user, it worked fine.)

Best I can tell, "lax" matches what we're already expecting to have happen -- it makes it so nothing can use a cookie to POST unless it's from one of our subdomains, but links that cause a GET won't break. 

Client posts to Protocol don't use cookies, right? Or do they?

zorkian:
Seems like pending discussion here, so I'm not merging for today's push.

Maybe we can set lax only if IS_DEV_SERVER and otherwise use secure?

Client protocols don't use cookies, per se, but things like the comment export page is an XML output that _does_. But I don't think that matters here, secure should be fine.

kareila:
Updated to give all cookies SameSite=Lax on dev servers.

I think the comment I left in the code encapsulates the situation as I understand it, but expect further developments. In particular I read that Chrome would start enforcing SameSite=Lax by default at some point, so if we do in fact need SameSite=None, we will need to make that explicit in the future.

PR #2780, opened 2020-07-23
Remove ancient JS function that always returns before doing anything

nfagerlund:
This main body of this function has been dead code since before Mark's 2008
initial import commit. Recently, Firefox started giving "unreachable code after
return statement" warnings for it.

Could just ignore it until the 6alib stuff is all gone for good, but hey,
figured we could all use a good laugh.


nfagerlund:
[beautiful](https://changelog.livejournal.com/4549447.html)

PR #2781, opened 2020-07-23
groundwork for profile conversion

kareila:
This is kind of a monster change so I'm splitting it into two PRs. This one lays the groundwork with new helper templates, string management, new methods, and necessary upgrades to existing methods. I recommend examining each commit separately.


views/error/purged.tt
```diff
@@ -0,0 +1,77 @@
...
+[%- CALL dw.request_status( 410 ); # 410 Gone -%]
+
+[%- sections.title = 'error.purged.name' | ml -%]
+[%- sections.windowtitle = 'error.purged.title' | ml -%]
+[%- CALL dw.active_resource_group( "foundation" ) -%]
+
+<p>[% 'error.purged.text' | ml %]</p>
+
+<!--
+xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```
nfagerlund, line 22: Wow, what?!

I dug -- it looks like this dates back to pauamma's original 404 BML (b1fa950341af0efbabaefcd8c899a674528a44fe). Reckon they'd remember why? I have no idea what this was meant to accomplish 11 years ago, but it has a strong smell of some BML-specific limitation. 

Oh no, wait, I dug further. Apache/LiveJournal.pm: 

```perl
    # Set to 1 if the code should generate junk to help IE
    # display a more meaningful error message.
    my $generate_iejunk = 0;

# ...

    $html .= ( "<!-- xxxxxxxxxxxxxxxxxxxxxxxxxxxx -->\n" x 100 ) if $generate_iejunk;
```

https://github.com/apparentlymart/livejournal/commit/3cb9c74f3f469ab069f82fb8d4edf13ced00d1b5

2003. It's something to do with IE 5 or 6.

I strongly suspect we don't want to continue doing this.

cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -63,6 +63,7 @@ sub action_links {
         "manage_membership", "trust",   "watch",  "post",
         "track",             "message", "search", "buyaccount"
     );
+    return \@ret;
```
nfagerlund, line 66: Ha, I'd totally forgotten that Perl has Ruby's implicit return values.

cgi-bin/DW/Logic/ProfilePage.pm
```diff
@@ -765,7 +768,17 @@ sub external_services {
     my $remote = $self->{remote};
     my @ret;
 
-    return () unless $u->is_personal && ( $u->share_contactinfo($remote) || $self->{viewall} );
+    return [] unless $u->is_personal && ( $u->share_contactinfo($remote) || $self->{viewall} );
+
+    # this has gotten big enough that we should improve database efficiency
```
nfagerlund, line 773: oh!! I always wondered what that method was about -- so I guess it lets you do this as a batch request without having to _remember_ that it's a batch request elsewhere in the code. TIL.

views/error/purged.tt
kareila: Unfortunately, my research indicates that not only could this still be necessary, but that Chrome may have also flirted with joining the party of ignoring the content of error pages if the source is less than 512 bytes.

https://stackoverflow.com/questions/53499913/ie-turn-off-friendly-http-error-messages-via-server
https://stackoverflow.com/questions/11544048/how-do-i-suppress-friendly-error-messages-in-internet-explorer
https://superuser.com/questions/202244/show-http-error-details-in-google-chrome
https://bugs.chromium.org/p/chromium/issues/detail?id=1695

cgi-bin/DW/Logic/ProfilePage.pm
kareila: Yeah, it takes all the listed props and batches them into a single DB lookup, and the subsequent prop calls get the info from the hashref or from memcache.

views/error/purged.tt
kareila: But then, I just realized that these pages are well over 512 bytes because they're being rendered in site scheme. Not sure if that's the case for every error page on the site, but for the ones going through DW::Template, we can definitely pull it.

PR #2783, opened 2020-07-25
remove unused module cgi-bin/LJ/Setting/Bio.pm

kareila:
The translation strings weren't even in the system. We use /manage/profile for this.


zorkian:
lol nice

PR #2786, opened 2020-07-26
[#2785] fix community import error

### LINKED ISSUE #2785: Community import still busted
C/p from comment, this should have been fixed with this push yeah? 

The option to import to/from a community broke in the last code push at the end of June, when the button that let you switch user disappeared.

The switch user button is back now, but when trying to import to a community, I'm still getting the same error message:

https://www.dreamwidth.org/tools/importer?authas=notestoaprawn

Error

You must be logged in to a personal account (not OpenID) in order to import content into it.

I'm logged in to my personal accounts on both DW and LJ when I see this message.

kareila:
In the original BML file, $u was the logged in user and $remote was the authas user, which is backwards from what the controller provides. So this error needs to be checking $remote, not $u.

Fixes #2785.

### DIFF
--- a/cgi-bin/DW/Controller/Importer.pm
+++ b/cgi-bin/DW/Controller/Importer.pm
@@ -54,7 +54,7 @@ sub importer_handler {
     }
 
     return error_ml('/tools/importer/index.tt.error.notperson')
-        unless $u->is_person;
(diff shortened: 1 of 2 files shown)


zorkian:
Oh werps, nice find.

PR #2789, opened 2020-07-26
second try for improvements to bin/checkconfig.pl

kareila:
The change from #2641 had to be reverted because requiring the config files, even within an eval block, tainted the execution state of the script and confused the call to LJ::get_dbh further down.

This version uses 'perl -c' for syntax checking like t/00-compile.t does. It also doesn't look for test config files unless $LJ::IS_DEV_SERVER is set.


zorkian:
That should work!

The other approach I was thinking was just to do it in a fork, and have the child do the requires and exit with a 0/1 as necessary, but that's fiddly in a different way.

PR #2802, opened 2020-07-27
Fix mobile blowouts on Drifting

nfagerlund:
This hack seems to have been an old version of the modern hack I replaced it with.
The drawback of the old one is that it doesn't constrain the widths of the table
'cells', so a long preformatted string of example CSS code or whatever will
blast your flist page off for Mars. Using flex lets it respect the per-entry
overflow: auto it already has for avoiding blowouts.


zorkian:
But I want to go to Mars

nfagerlund:
Not by noclipping through the floor, you don't

PR #2803, opened 2020-07-28
Rework of /customize

momijizukamori:


momijizukamori:
@zorkian when you have a hot minute, I think the Perl in this is as good as it's gonna get from my side of things without any feedback/review.

@nfagerlund I could use some help going over the CSS/JS to clean up any messes I left behind/organize things as sanely as possible.

momijizukamori:
Okay, I think this is as solid as I'm going to get it at this point, and could use some reviews.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    $vars->{u}            = $u;
+    $vars->{remote}       = $remote;
+    $vars->{is_identity}  = 1 if $u->is_identity;
+    $vars->{is_community} = 1 if $u->is_community;
+    $vars->{style}        = LJ::Customize->verify_and_load_style($u);
+    $vars->{authas_html}  = $rv->{authas_html};
+
+    my $queryargs = make_queryargs($GET);
+
+    if ( $u ne $remote ) {
```
zorkian, line 55: These are going to be hashrefs aren't they? Can you compare them like this?

IIRC there's some kind of LJ::same_user function that would work here?

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+DW::Routing->register_rpc( "journaltitles", \&journaltitles_handler, format => 'html' );
+DW::Routing->register_rpc( "layoutchooser", \&layoutchooser_handler, format => 'json' );
+DW::Routing->register_rpc( "themefilter",   \&filter_handler,        format => 'json' );
+
+sub customize_handler {
+    my ( $ok, $rv ) = controller( authas => 1, form_auth => 1 );
+    return $rv unless $ok;
+
+    my $r      = $rv->{r};
+    my $post   = $r->post_args;
```
zorkian, line 40: nit: make this `$POST` since you have `$GET` below? or change them both to `$post_args` and `$get_args`, which I believe we use elsewhere.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+
+    # set up the keywords for basic search
+    my @keywords = LJ::Customize->get_search_keywords_for_js($u);
+    $vars->{autocomplete} = \@keywords;
+
+    # we want to have "All" selected if we're filtering by layout or designer, or if we're searching
+    $vars->{viewing_all} = $queryargs->{layoutid} || $queryargs->{designer} || $queryargs->{search};
+
+    # sort cats by specificed order key, then alphabetical order
+    my %cats = LJ::Customize->get_cats($remote);
```
zorkian, line 70: Is this intentionally `$remote` here? Since we're operating on `$u` otherwise.

cgi-bin/DW/Controller/Customize.pm
zorkian: The next function uses `$POST` as well, so we should be consistent in the file.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    my $theme_data = get_themechooser_data($queryargs);
+    my $theme_html = DW::Template->template_string( 'customize/themechooser.tt',
+        { 'theme_data' => $theme_data, qargs => $queryargs } );
+    my $current_data = get_current_data(0);
+
+    $r->print( to_json( { theme_html => $theme_html, current_data => $current_data } ) );
+    return $r->OK;
+}
+
+sub layoutchooser_handler {
```
zorkian, line 231: Don't you need 

```perl
    my ( $ok, $rv ) = controller( authas => 1, form_auth => 1 );
    return $rv unless $ok;
```

to be safe?

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+
+    # set the new titles
+
+    set_layout($args);
+
+    $r->print( to_json( get_layout_data($args) ) );
+    return $r->OK;
+}
+
+sub journaltitles_handler {
```
zorkian, line 245: Don't you need 

```perl
    my ( $ok, $rv ) = controller( authas => 1, form_auth => 1 );
    return $rv unless $ok;
```

to be safe?

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    # Now we tell it what template to render and pass in our variables
+    return DW::Template->render_template( 'customize/index.tt', $vars );
+
+}
+
+## Functions to handle edits for both RPC and main endpoints
+
+sub set_theme {
+    my %opts = @_;
+    my $u    = LJ::get_effective_remote();
```
zorkian, line 325: I know this is pretty common in BML but since we moved to Controllers, I'd prefer we pass in the `$u` we're operating on instead of inferring it from the environment. Do we do this in other controllers?

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    LJ::Customize->apply_theme( $u, $theme ) or croak("Couldn't apply theme");
+    LJ::Hooks::run_hooks( 'apply_theme', $u );
+}
+
+sub set_layout {
+    my $post = shift;
+
+    my $u = LJ::get_effective_remote();
```
zorkian, line 349: Same as above re: passing in `$u`.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    my $style = LJ::S2::load_style( $u->prop('s2_style') );
+    die "Style not found." unless $style && $style->{userid} == $u->id;
+
+    LJ::Customize->save_s2_props( $u, $style, \%override );
+}
+
+sub set_journaltitles {
+    my $post = shift;
+
+    my $u = LJ::get_effective_remote();
```
zorkian, line 378: Same as above re: passing in `$u`.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    $eff_val = "" unless $eff_val;
+    $u->set_prop( $post->{which_title}, $eff_val );
+}
+
+# Functions to render backend data in a format that is simpler to process in our templates/JS
+
+sub get_themechooser_data {
+    my $queryargs = shift;
+    my @themes;
+    my $u      = LJ::get_effective_remote();
```
zorkian, line 391: Same as above re: passing in `$u`.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    my $remote = LJ::get_remote();
+    my %cats   = LJ::Customize->get_cats($u);
+
+    if ( $u ne $remote ) {
```
zorkian, line 395: Think we need to use same_user or compare userids or something.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    else {
+        $tmp->{preview_url} = LJ::create_url( "/customize/preview_redirect",
+            args => { 'layoutid' => $theme->layoutid } );
+    }
+    return $tmp;
+}
+
+sub get_current_data {
+    my $no_themechooser = shift;
+    my $u               = LJ::get_effective_remote();
```
zorkian, line 537: Same as above re: passing in `$u`.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    push @current_options,
+        { url => "#layout", title => LJ::Lang::ml('widget.currenttheme.options.layout') };
+
+    return { 'current' => $theme_data, 'current_options' => \@current_options };
+}
+
+sub get_layout_data {
+    my $u = LJ::get_effective_remote();
```
zorkian, line 589: Same as above re: passing in `$u`.

cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,654 @@
...
+        layouts           => \@layout_data
+    };
+
+}
+
+sub make_queryargs {
+    my $args = shift;
+    my $ret_args;
+
+    $ret_args->{cat}      = defined $args->{cat}      ? $args->{cat}      : "";
```
zorkian, line 644: nit: all of these can be replaced with the form,

```perl
$ret_args->{cat} //= "";
```

All hail the defined-or operator!

zorkian (changes requested):
Some small changes, a few nits, but generally looking pretty solid!

cgi-bin/DW/Controller/Customize.pm
momijizukamori: Just checked and yeah, there is only one single use of `get_effective_remote` in the new controllers (in Create). We do seem to use `LJ::get_remote();` for RPC endpoints though - I will see if the regular Controller formulation works for that or not.

cgi-bin/DW/Controller/Customize.pm
momijizukamori: Other spots look like they compare $u->user and $remote->user so I updated it to that.

cgi-bin/DW/Controller/Customize.pm
momijizukamori: We aren't super-consistent, but I've converted all of them to $post_args/$get_args in this file at least.

cgi-bin/DW/Controller/Customize.pm
momijizukamori: So because this and the journaltitles_handler are only hit via JS, and don't rewrite large swathes of DOM, the form_auth value goes stale after the first request using it. I feel like these two are fairly low risk for CSRF, though?

cgi-bin/DW/Controller/Customize.pm
momijizukamori: Oh, good catch, checking the old code it looks like should probably be $u (though uh in our case it won't make any difference, all users get the same theme cats)

momijizukamori:
Closing this as it's been superseded by other work!

PR #2808, opened 2020-07-28
fix missing strings from converting /shop/receipt to TT

kareila:
Oh no, I forgot to check for other shop files referencing strings defined in receipt.bml.


kareila:
Oh no, I also unleashed an unintended SQL query change that had unhappy results for older (PP/GCO) order views.

PR #2809, opened 2020-07-29
more misc. cleanup

kareila:
Buncha junk.


zorkian:
Thanks for fixing a Mark todo :)

Issue #2784, opened 2020-07-26
roundup of post-code-push glitch reports on Discord


kareila:
1. Some people are saying the Tropo Purple colors are too bright/saturated in Foundation. Not sure if this is something we need to revise.

2. Show more/less button on comment preview displays even if the comment is short enough to be already fully shown.

3. Foundation default table styling applies to tables in journal entries on site schemed pages.

4. Casual HTML dropdown is bright white in Gradation Vertical.

kareila:
Here's an example of (3) from support: https://www.dreamwidth.org/support/see_request?id=41811

rahaeli:
* the table holding the pagination links on entries that have 50+ comments used to have a background of the skin's background accent color, now it's the default page bg

rahaeli:
* Tropo Red is having at least partially the same problem as in Tropo Purple on comment headers, and I think what's wrong *may be* that the primary accent background color and secondary accent background color have been swapped? The background in comment headers on screened comments in Red is the light-pink secondary accent background color now, not the darker red primary accent bg color, at least.

rahaeli:
* Wide images in inbox notifications are forcing sidescroll/stretch (and the user implies that they weren't before)

rahaeli:
* There's a collection of people who are complaining that their CSS in entries/comments is having new and different padding applying to it (and breaking their CSS designs) and that the padding in comment headers is inadequate now (and I agree it looks a bit wonky), so I'm guessing there's an order-of-load issue going on somewhere maybe? This is probably bigger than just a single bullet point in a catch-all issue, and will need to be broken out into its own issue with some test cases, but I'm trying to make sure I dump all the problems people are reporting somewhere

kareila:
Textarea boxes are width: 100% and should be width: auto - https://dw-maintenance.dreamwidth.org/83294.html?thread=3363678#cmt3363678

kareila:
I think we determined that emailed notifications of entries and/or comments didn't have linebreaks applied correctly? I haven't seen an example of this yet.

kareila:
ilyena_sylph provided this screenshot, which I believe shows an overly narrow reply div on the reply page (as opposed to quick reply). Clicking "reply at the webpage" from an email links to this view. Journal style is Abstractia Abyss.

![unknown](https://user-images.githubusercontent.com/1976742/88492863-83f0b380-cf73-11ea-8be9-2589e7400d27.png)

nfagerlund:
[...]

PR #2810, opened 2020-07-29
Various new site skin fixes

### LINKED ISSUE #2793: Different fix for comment subject height shenanigans
https://github.com/dreamwidth/dw-free/commit/91a90ebab8ac42580376928c2a70002b8d25a030

Nick described this fix and Mark pushed it, but it needs someone to come back and add a media query to only add the extra space for desktop views.

nfagerlund:
(Many of these are mentioned in #2784.)

- The border colors on comment headers were wonky after re-adding the color
alternation. (One of them was half-highlight-color, half-black). Fixed by
exploding "border-bottom" into its constituent parts. (The "combined" border
properties implicitly set border-color to `currentcolor`, which is black-ish.)

- Add padding to .comment-info (comment header) boxes.

- Reduce the space we waste for absent comment headers on mobile.
Fixes #2793

- We shouldn't mess with table styles in Lynx (since they're browser default
anyway), so move that out of the "need this to display comment pages" file and
into the "fix bad Foundation behavior for thick siteskins" file.

- Remove default Foundation styling for textareas in entries (which are used to
share code).

- Fix a rare issue where the height of the comment form textarea got compressed.
(Only affects highly obsolete browsers.)

### DIFF
--- a/htdocs/scss/skins/_entry-styles.scss
+++ b/htdocs/scss/skins/_entry-styles.scss
@@ -205,11 +205,13 @@ ul.entry-management-links {
     }
 
     .comment-info, .userpic {
-      border-bottom: 1px solid;
+      // Using "exploded" border properties so we can set the color later.
+      border-bottom-width: 1px;
+      border-bottom-style: solid;
     }
     .comment-info {
-      border-right: 1px solid;
-      border-color: $soft-accent-color;
+      border-right-width: 1px;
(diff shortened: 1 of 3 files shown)


nfagerlund:
I don't think any of this depends on any user testing the way the wooz stuff does.

zorkian:
All hail potato browsers

PR #2811, opened 2020-07-29
Site skin comments: Attempt to fix the wooz

nfagerlund:
Attempting to fix some vertigo issues reported with the new site skin comments,
working on a hypothesis that it involves some combination of line length and
spacing.

This shrinks the average line length of entries and top-level comments by about
ten or fifteen characters per line, at least with the default font (verdana).
It also adds some vertical spacing around the body of a comment.

### DIFF
--- a/htdocs/scss/skins/_entry-styles.scss
+++ b/htdocs/scss/skins/_entry-styles.scss
@@ -266,6 +266,10 @@ ul.entry-management-links {
   }
 }
 
+.comment-content {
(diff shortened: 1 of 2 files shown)


nfagerlund:
Per @rahaeli, let's try shipping 72em and see what happens. 🤷

nfagerlund:
See also https://github.com/dreamwidth/dw-nonfree/pull/141

PR #2818, opened 2020-08-03
test commafy without ljlib

martindemello:


t/lib/ljtestlib.pl
```diff
@@ -93,4 +93,20 @@ sub temp_user {
     return $u;
 }
 
+# For tests that just need to hardcode a few responses from LJ::Lang::ml this
+# provides an easy way to set up the lookup table.
+# Usage: LJ::Mock::make_fake_lang_ml( {'key' => 'value', ...} )
+sub make_fake_lang_ml {
+    my ($table) = @_;
+    my sub fake_lang_ml {
```
kareila, line 101: We don't use `my sub` syntax anywhere else in the code base - this would be more typically written as a scalar subroutine reference:
```perl
my $fake_lang_ml = sub {
```

t/lib/ljtestlib.pl
```diff
@@ -93,4 +93,20 @@ sub temp_user {
...
+        my ($arg) = @_;
+        return $table->{$arg};
+    }
+    $mock->fake_module(
+        'LJ::Lang' => (
+            ml => \&fake_lang_ml
```
kareila, line 107: Following the above, this would now be:
```perl
 ml => $fake_lang_ml
```

t/lib/ljtestlib.pl
martindemello: done

zorkian:
Nice, thank you!

PR #2820, opened 2020-08-04
Re-enable javascript in core1 with s2foundation features

### LINKED ISSUE #2819: s2foundation breaks javascript on core1 styles
This one's my bad. 

We used to put script tags in the `<head>` element, and LJ/S2.pm controlled all of that. But Foundation expects them at the bottom of the `<body>`, and the body is generally under the control of individual styles, so we can't just handle script tags wholly from perl land anymore. 

In core2, that's easy, because we already had this `Page::print_wrapper_end()` function and we clearly documented that you HAD to use it in place of your closing `</body>` tag. So I just threw the script tags in there. 

In core1, IDK what we need to do. Basically the question is, do we expect core1 styles to override `Page::print()`, or not? 

If not, we can just throw the script tags at the bottom of the body ourselves. 

If yes, everyone will need to update their core1 styles to get JS working again. 😱

OR, maybe we can figure out a way to move script tags back to the head despite being in Foundation land. We might be able to use [the `defer` attribute](https://caniuse.com/#search=defer) on the script tags, which should have a similar effect to putting them at the bottom of the body. (Would also need to replace the inline script tag that just calls `$(document).foundation()` or whatever with an external script that does the same thing, since you can't defer inlines.)

nfagerlund:
Fixes #2819 

I was speculating about ways to fix this without requiring changes to existing core1 styles, but @rshatch pointed out that core1 has been deprecated for the whole last decade and that requiring a one-line change to antique custom styles might not be unreasonable. 

----

To support newer components in journal styles, we needed to change how script
tags enter the document. core2 already had room to do this without disrupting
anything, but naturally core1 does not.

Since most legacy core1 layouts in the wild probably override Page::print(),
they'll need to update their code to call `$this->print_script_tags();` right
before closing the body tag.

### DIFF
--- a/styles/core1.s2
+++ b/styles/core1.s2
@@ -565,6 +565,9 @@ class Page
     function print_stylesheets
     "Prints all defined stylesheets, including default and user-defined ones.";
 
+    function builtin print_script_tags
+    "Prints the JavaScript tags that this page requires. If a layout overrides the Page::print function, it must call this function before closing the body tag.";
+
     function builtin print_trusted(string key)
     "Prints a trusted string by key.";
 }
@@ -2225,7 +2228,9 @@ body, td, th, table, p, div {
     $this->print_body();
     """\n<hr><address>""";
     server_sig();
-    """</address>\n</body></html>""";
+    """</address>\n""";
+    $this->print_script_tags();
+    """</body></html>""";
 }
 
 function Page::print_body() {


nfagerlund:
This compiles just fine, but I have no good way to test it for real, since I don't have any core1 layouts ready to hand.

PR #2824, opened 2020-08-06
new page /admin/recent_accounts

kareila:
Right now when I'm checking recently created accounts for spammers, it requires a lot of time and resources on my part. So I brainstormed with @rahaeli about a way to crowdsource reviewing recently created accounts with trusted volunteers in a way that made it easy to participate on mobile devices for a few minutes here and there and distribute the workload. This is what came out of that conversation.

- Newly created personal accounts have a new userprop, `not_approved`, set to 1 at account creation time.

- Any accounts with `not_approved` set are excluded from /search, /latest, and /random. (New user method `$u->is_approved` faciliates this.)

- After an hour, the new accounts will begin to appear on /admin/recent_accounts for review by users with the existing priv siteadmin:spamreports or the new priv siteadmin:approvenew.

- One of these volunteers will review the journal and flag it as spam or not. If it's not, `not_approved` is cleared and the account is made available in the site features where it was previously excluded. If it is judged to be spam, `not_approved` is incremented from 1 to 2 and it goes into another queue to be suspended.

- Another new priv, suspend:recent, is available for volunteers to suspend only accounts flagged from the /admin/recent_accounts page. Suspending or unsuspending an account will also clear the userprop.

All of this is switched off if `LJ::is_enabled('approvenew')` returns false. I left it turned off in the example config, but it should be active in production unless we explicitly disable it.

Screenshot of journal review form (yes it's an iframe):

Screen_Shot_2020-08-02_at_4 56 58_PM

### DIFF
--- a/cgi-bin/DW/Controller/Admin/Approve.pm
+++ b/cgi-bin/DW/Controller/Admin/Approve.pm
@@ -1,0 +1,137 @@
+#!/usr/bin/perl
+#
+# DW::Controller::Admin::Approve
+#
+# Interface for screening new accounts for spam content.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
+#
+# Copyright (c) 2020 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself. For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Controller::Admin::Approve;
+
+use strict;
+
+use DW::Controller;
+use DW::Routing;
+use DW::Template;
+use DW::FormErrors;
+
+my $privs = [
+    'siteadmin:approvenew',     # just for this page
+    'siteadmin:spamreports',    # include existing antispam team
+    'suspend',                  # include anyone with generic suspend privs
+    'suspend:recent',           # just for this page
+];
+
+DW::Controller::Admin->register_admin_page(
+    '/',
+    path     => 'recent_accounts',
+    ml_scope => '/admin/recent_accounts/review.tt',
+    privs    => $privs
+);
+
+DW::Routing->register_string( '/admin/recent_accounts', \&approve_handler, app => 1 );
+
+sub approve_handler {
+    my ( $ok, $rv ) = controller( form_auth => 1, privcheck => $privs );
+    return $rv unless $ok;
+
+    my $scope = '/admin/recent_accounts/review.tt';
+
+    my $prop = LJ::get_prop( 'user', 'not_approved' );
+    return error_ml( "$scope.error.disabled", { sitenameshort => $LJ::SITENAMESHORT } )
+        unless $prop && LJ::is_enabled('approvenew');
+
+    my $r         = $rv->{r};
+    my $remote    = $rv->{remote};
+    my $form_args = $r->post_args;
+    my $vars      = {};
+
+    $vars->{can_suspend} = 1
+        if $remote->has_priv( 'suspend', '' ) || $remote->has_priv( 'suspend', 'recent' );
+
+    return DW::Template->render_template( 'admin/recent_accounts/review.tt', $vars )
+        unless $r->did_post;
+
+    if ( $form_args->{begin} ) {
+
+        # get a user to review - we don't want to just get the first result,
+        # high chance of collision with multiple active reviewers
+        my @uids;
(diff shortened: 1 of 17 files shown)


azurelunatic:
I like!!

cgi-bin/DW/Controller/Admin/Approve.pm
```diff
@@ -0,0 +1,180 @@
...
+    my $vars      = {};
+
+    $vars->{can_suspend} = 1
+        if $remote->has_priv( 'suspend', '' ) || $remote->has_priv( 'suspend', 'recent' );
+
+    return DW::Template->render_template( 'admin/recent_accounts/review.tt', $vars )
+        unless $r->did_post;
+
+    my $dbr = LJ::get_db_reader();
+    my $sth = $dbr->prepare( "SELECT userid FROM userproplite2 "
```
zorkian, line 64: If you are going to use the clustered property table (`userproplite2`), then you need to use cluster readers. Something like,

```perl
fore my $cid ( @LJ::CLUSTERS ) {
    my $dbcr = LJ::get_cluster_reader( $cid ) or die;
    # look for users
}
```

The other option is to not cluster this property, and use the `userproplite` table, but to do that you need to set `cldversion: 0` and `indexed: 0` in the property settings.

cgi-bin/DW/Controller/Admin/Approve.pm
```diff
@@ -0,0 +1,180 @@
...
+        foreach my $uid (@ids) {
+            my $u = $users->{$uid};
+            next if $u->is_suspended;    # already done
+
+            if ( $form_args->{"user_$uid"} ) {
+                $u->set_suspended( $remote, "from /admin/recent_accounts" );
+                $count++;
+            }
+            else {
+                $u->set_prop( not_approved => 0 );
```
zorkian, line 165: I forgot, does zero delete the property? I would rather not have the database fill up with zeros.

cgi-bin/LJ/Console/Command/Suspend.pm
```diff
@@ -45,6 +45,11 @@ sub execute {
     my $remote = LJ::get_remote();
     my $openid_only =
         $remote->has_priv( "suspend", "openid" ) && !$remote->has_priv( "suspend", "*" );
+    my $recent_only =
+        $remote->has_priv( "suspend", "recent" ) && !$remote->has_priv( "suspend", "*" );
+
+    return $self->error("You can only suspend users using /admin/recent_accounts.")
+        if $recent_only;
```
zorkian, line 53: I think this needs to handle the case where someone can suspend both recent and open ID, but does not have Global suspend.

cgi-bin/DW/Controller/Admin/Approve.pm
kareila: Hmm, for the way I'm using it, I think it makes sense for it not to be clustered, but I thought it should be indexed if I was going to be searching the column for matches as opposed to fetching single user rows? Can you clarify that for me? The issue of clustering didn't come up in my testing because my dreamhack doesn't use clusters. (I thought the multihomed property designated whether it was clustered or not.)

cgi-bin/DW/Controller/Admin/Approve.pm
kareila: Yes, zero deletes. (Line 1157 of cgi-bin/LJ/User/Permissions.pm)

cgi-bin/DW/Controller/Admin/Approve.pm
zorkian: The naming on these properties is definitely wild.

`cldversion` means 'clustered dversion' i.e. if this field is non-zero, the property is clustered and not stored on the global database.

`indexed` has no effect on clustered properties. It applies to global properties (i.e. `cldversion: 0`).

`multihomed` however makes a clustered property _also_ store a copy on the global db. _But_ it also looks crufty and I'm not sure we should ever use it. It used to be used for properties that would be shown on the profile page (such as `icq`) but we stopped setting it and things work, so maybe we delete this feature? heh. 🔪 

So, basically, what you probably want is `cldversion: 0` and `indexed: 1`, that will allow you to do lookups like "where propid = ? and value = 1" and get back everybody who hasn't been approved yet while hitting an index.

FWIW, I think the indexing probably doesn't add much value since the cardinality of values is 2 and the size of the set is O(new users that haven't been reviewed). I would expect that we'd actually spend more time on the index management than we do on filtering out things anyway.

PR #2826, opened 2020-08-09
remove ineffective editjournal feature action:saveunsuspend

kareila:
The intent of this feature was to automatically open a new support request when a suspended entry was edited to remove offending content. This had no effect because `$LJ::UNSUSPENSION_REQUEST_SPCATID` was never set.

Most of the actual changes here are related to removing the `unsuspend_supportid` entryprop, intended to link the suspended entry and the support request. The only user-visible change is the removal of the extra (useless) button on /editjournal.


zorkian:
I don't even remember this feature

rahaeli:
It was just getting coded when we forked to let people edit their suspended stuff and notify ToS automatically.

PR #2827, opened 2020-08-10
get JS test suite working again

kareila:
_DID YOU KNOW_ there is a JS test suite? It's located under /dev/tests/ but only on dev servers, not in production.

I poked at it and found several issues. This set of changes fixes those issues.


nfagerlund:
holy heck

I saw that ages ago, but assumed it was long-abandoned. thank you for kicking it.

PR #2829, opened 2020-08-14
Stop "closing" `<br>` tags at the ends of paragraphs

nfagerlund:
- `<br>` is an [empty element](https://developer.mozilla.org/en-US/docs/Glossary/empty_element), so it has no closing tag. 
- `</br>` is invalid, but browsers appear to interpret it as a `<br>` tag instead of ignoring it.
- `<br>` is the canonical, correct form of `<br>`. However, `<br />` is also accepted, because once upon a time someone was trying to make XHTML happen, so a lot of tools still (wrongly) emit that. (XHTML is _not_ going to happen.) 
- We have two implementations of the thing in the cleaner that closes tags: a per-tag one, and an end-of-document one. The latter handles `<br>` just fine, and the other one craps itself because it is trying to make XHTML happen. (XHTML is _not_ going to happen.)

### DIFF
--- a/cgi-bin/LJ/CleanHTML.pm
+++ b/cgi-bin/LJ/CleanHTML.pm
@@ -1160,6 +1160,7 @@ TOKEN:
                             my $close;
                             while ( ( $close = pop @tagstack ) && $close ne $tag ) {
                                 $opencount{$close}--;
+                                next if $close =~ $slashclose_tags;
                                 $newdata .= "</$close>";
                             }
                         }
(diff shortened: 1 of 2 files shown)


Sophira:
Thank you for fixing this! This has been an issue for the dwqdb community for a while.

nfagerlund:
@sophira for real?! 😱 I just got ambushed by it for the first time (bc I don’t usually do mid-graf brs).

Uh, got any _other_ egregious markup bugs in your pocket that I haven’t heard of yet? 😅

zorkian:
Aww but I _liked_ `<br />` (lol).

PR #2838, opened 2020-08-31
[#2832] Don't drop edit reason when previewing a comment edit

### LINKED ISSUE #2832: "Preview comment" drops reason-for-edit since last code push
Support req: https://www.dreamwidth.org/support/see_request?id=42369

nfagerlund:
Fixes #2832 

This is a relative of #2724 -- old preview would propagate unrecognized fields
as hidden values, so while the preview form wouldn't _show_ you the edit reason,
it _would_ retain it. The new preview actually includes an edit reason field,
but blanks out whatever value you entered before previewing (or resets it to the
saved edit reason, if this is the second+ edit).

Easy enough fix; it's just the talkform wasn't expecting to receive that field
in a partial form submission.

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1491,9 +1491,8 @@ sub talkform {
         default_usertype => 'user',
 
         comment => {
-            editid     => $editid,
-            editreason => $form->{editreason}
-                // ( $comment ? $comment->edit_reason : '' ),
+            editid      => $editid,
+            editreason  => $form->{editreason} // ( $comment ? $comment->edit_reason : '' ),
             oidurl      => $form->{oidurl},
             oiddo_login => $form->{oiddo_login},
             user        => $form->{userpost},
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -1493,7 +1493,7 @@ sub talkform {
         comment => {
             editid     => $editid,
             editreason => $form->{editreason}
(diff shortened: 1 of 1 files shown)


cgi-bin/LJ/Talk.pm
```diff
@@ -1491,8 +1491,9 @@ sub talkform {
         default_usertype => 'user',
 
         comment => {
-            editid      => $editid,
-            editreason  => $comment ? $comment->edit_reason : '',
+            editid     => $editid,
+            editreason => $form->{editreason}
+                || ( $comment ? $comment->edit_reason : '' ),
```
kareila, line 1496: Should we use `//` instead of `||` here, to allow the user to submit the empty string in order to blank out a previously provided edit_reason? Or does that even work?

nfagerlund: Oh! Thank you, that's a good catch. (By my understanding, that would work, and I think this is one of the props where the edit_comment function would turn the empty string into a prop removal.) As written, you'd be able to blank out the edit reason on final submit, but we'd revive the old edit reason if you previewed instead, which is inconsistent and not what we want. 

```suggestion
                // ( $comment ? $comment->edit_reason : '' ),
```

zorkian:
LGTM but fails tidy.

nfagerlund:
😑 looks like perltidy has different wrap rules for `||` and `//`.

kareila:
Clearing the requested-changes label since it got tidied.

PR #2840, opened 2020-09-02
Icon components

momijizukamori:
This PR splits the code for selecting an icon in forms out into reusable components rather than each form reimplementing their own logic for it. 

- Use `[% INCLUDE 'components/icon-select-dropdown.tt' %]` to include the dropdown form, and `[% INCLUDE 'components/icon-select-icon.tt' %]` to include the icon preview block (they're seperate elements because forms use them in different orders). 
- They expect `remote` to be defined somewhere in the template scope, and optionally `current_icon` and `current_icon_kw` if an icon/keyword other than default should be pre-selected.
- `icon-select-icon` includes the JS/CSS for both components - no extra JS/CSS is needed in the parent template
- `icon-select-icon` also includes the icon-browser component, with styling to keep it hidden properly on pages not yet using Foundation.
- if the user has no icons uploaded, the preview picture changes to the generic default DW icon and the click action + text change to taking the user to the manage icons page
- quick reply, entry replypage, and beta create page have all been updated to use the new components (inbox will be coming in it's own conversion PR)
- `icon_keyword_menu`, a sub that creates an array of keywords and values suitable for passing to a select form component, is now a User function, ie, `$user->keyword_icon_menu();`, and has been moved out of Web.pm accordingly.


cgi-bin/LJ/Talk.pm
```diff
@@ -1474,6 +1473,28 @@ sub talkform {
 
     my $screening = LJ::Talk::screening_level( $journalu, $opts->{ditemid} >> 8 ) // '';
 
+    # pre-calculate some abilities and add them to $remote, so we don't have to do it
+    # in the template
+    if ($remote) {
+        $remote->{can_manage_community} =
```
zorkian, line 1479: I'm not comfortable adding fields to an LJ::User here like this. Is there any way we can just pass this as a separate hashref, some 'remote_icon_opts' or something idk?

cgi-bin/LJ/Talk.pm
```diff
@@ -1515,40 +1536,7 @@ sub talkform {
...
-                && $remote
-                && $remote->can_manage($journalu),
-
-            can_use_userpic_select   => $remote->can_use_userpic_select && ( scalar(@icons) > 0 ),
-            iconbrowser_keywordorder => $remote->iconbrowser_keywordorder ? "true" : "false",
-            iconbrowser_metatext     => $remote->iconbrowser_metatext ? "true" : "false",
-            iconbrowser_smallicons   => $remote->iconbrowser_smallicons ? "true" : "false",
-            }
-        : 0,
+        remote  => $remote ? $remote : 0,
```
zorkian, line 1539: Why 0? Shouldn't this be undef, if it's "there is no remote"? Or does TT want a 0 for some reason... and if it's just undef, we should just say `remote => $remote`.

cgi-bin/LJ/User/Icons.pm
```diff
@@ -900,4 +900,62 @@ sub _create_mapid {
...
+# args: user
+# des-user: a user object.
+# returns: An array of hashrefs like:
+#          {value => ..., text => ..., data => { url => ..., description => ... }}
+#          which can be passed directly to an LJ::html_select(). Includes an
+#          item for each keyword (thus duplicating icons with multiple
+#          keywords), and an item for the default userpic. If no userpics or if
+#          user is undefined, returns an empty array.
+# </LJFUNC>
+sub icon_keyword_menu {
```
zorkian, line 917: Why did we move this out of `LJ::Web`? My mental model is that the `LJ::User` classes are the models, and they don't care anything about what the output will be... and separately, we have `LJ::Web` + controllers/TT files which are the views, those actually do transformations.

Maybe it's fine, but I'm curious what the thinking is.

zorkian (changes requested):
One question, one nit, and one actual request for change! Looking good, though :)

cgi-bin/LJ/Talk.pm
momijizukamori: This one is a question for @nfagerlund as that bit of cose is his.

cgi-bin/LJ/Talk.pm
momijizukamori: That might be possible! The original code made a new `$remote` object that consisted of these bits and a handful of properties from the original remote, but not all of them.

cgi-bin/LJ/User/Icons.pm
momijizukamori: Primarily, to simplify the process of using the new icon components. Before, you'd have to either call the function seperately in the controller, or pass in a reference to it as a variable to the template. Now the icon component really just needs the remote, so you don't have to worry about dealing with other dependencies. 

In terms of organization - it generates the icon list for that specific user, and LJ::Web is kind of a chaotic mess with a billion things stuffed in it.

momijizukamori:
@zorkian when you have a free minute, I think I've addressed all the requested changes!

kareila:
Clearing the requested-changes label since Momiji has asked for another round of review.

zorkian:
LGTM, thanks for pulling those opts out!

PR #2660, opened 2020-05-22
talkpost do and preview too

nfagerlund:
MANY thanks to @rshatch for doing the BML conversion at the very start. 🙏🏼 I wasn't scared of the rest of this, but it was out on the far limits of my ability to understand this warty old codebase, and adding on one more thing to learn would have been too damn much. 

Fixes #2646 
Fixes #2551 
Fixes #2572 

This PR dynamites a ton of ancient garbage, including LJ::Talk::Post::init and LJ::Talk::Post::make_preview. A lot of DW's code seems to be working around the fact that no one wanted to touch these two functions ever again, so I bet this is gonna be a blast to review. But their replacements are fairly constrained and testable, as evinced by the fact that they HAVE tests now. 

Speaking of which, what are the cases you really want tested for `prepare_and_validate_comment`? I'm about at my limit for the authentication one, since logins act so screwy in testing, but I'd be down to add several more validation checks if people point out the cases that are most historically thorny. 

LMK if you want a tour through what's going on in here, but hopefully it's broken up into commit units that tell most of the story. The earlier commits with the really wild stuff (after the "scour BML from preview" one) are meant to always leave the site in a working state. 

After some thought, I decided that "check ban to lose the ability to ban" test actually serves as useful documentation, so I brought it in. 

<details><summary>Screens:</summary>

image

image

image
</details>

PR #2843, opened 2020-09-07
[#2841] Properly record anon comment IPs in the temp table for spam

### LINKED ISSUE #2841: reports of anonymous comment spam disappearing into the ether?
https://www.dreamwidth.org/support/see_request?id=42466

@pauamma says that no anonymous spam has been reported since June 27, which corresponds with the date of the code push described in this code tour: https://dw-dev.dreamwidth.org/222909.html

Maybe something went awry with the refactoring of the talkpost code?

nfagerlund:
Fixes #2841 
Caused by #2660 

This is another instance of #2738 -- edit_comment and enter_comment used to
mutate their `$comment` argument by adding a `talkid` key with the jtalkid, and
I missed a few things referenced that value via the mutated hashref rather than
one of their other handles to the jtalkid. In this case, it was preventing
anonymous comment spam reports by not properly writing the IP address of new
anon comments to the tempanonips table.

This time I hunted real hard for anything in the code that might access any
hashref key called `talkid`. As a result, this commit also:

* Fixes one more instance of That Thing (re-screening when editing).
* Just deletes the final remaining instance (an unused hook).
* Yanks a similar "mutate the `$comment` argument" behavior in
`enter_imported_comment` -- nothing ever references that hashref argument again
before it goes out of scope.

All the remaining hashrefs with `talkid` members seem to be either s2 comment
objects or the `parent` element of one of `prepare_and_validate_comment`'s
comment hashrefs.

### DIFF
--- a/cgi-bin/LJ/Talk.pm
+++ b/cgi-bin/LJ/Talk.pm
@@ -2324,7 +2324,7 @@ sub enter_comment {
     LJ::MemCache::incr( [ $journalu->{'userid'}, "talk2ct:$journalu->{'userid'}" ] );
 
     # record IP if anonymous
-    LJ::Talk::record_anon_comment_ip( $journalu, $comment->{talkid}, LJ::get_remote_ip() )
+    LJ::Talk::record_anon_comment_ip( $journalu, $jtalkid, LJ::get_remote_ip() )
         unless $posterid;
 
     # add to poster's talkleft table, or the xfer place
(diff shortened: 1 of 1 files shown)


nfagerlund:
Verified the fix on my hack, although it only affects new anon comments submitted after the fix.

zorkian:
Fight the good fight, sir, and one day when we rewrite this all in Rust, we can banish these gremlins forever!

PS we're not actually going to rewrite this all in Rust.

PR #2846, opened 2020-09-09
The Squish, Part Four

### LINKED ISSUE #2845: The Squish, Part 4
This came out of [a conversation with a cartoonist over in dw-dev](https://dw-dev.dreamwidth.org/224517.html?thread=1765637#cmt1765637).

> Recap: The Squish is so people can casually post normal images without breaking the reading page. Ideally, it's meant to make things easier for posters, not get in their way.
>
> The current squish behavior is to: 
>
> - Shrink images in content so that they fit inside the container width and the viewport height. 
> - Enable click-to-zoom on squished images that aren't inside a hyperlink. Zooming displays the image at its full intrinsic size. 
> - _Don't_ shrink images that are inside decorative markup, like a table or an inline-styled div. 

The Squish still has room for improvement. Here's a couple of things I'm considering: 

### Use `height` and `width` attributes to limit the "zoomed" state

Right now, it's like: 

- **When shrunk:** The width attribute of the img tag act as a maximum on the "squished" state of the image -- it might shrink smaller, but it won't start out any bigger than that. 
- **When zoomed:** The image always uses its full natural pixel size. 

However, some images use oversized natural pixel dimensions so they can look good at higher pixel densities. Those images shouldn't actually be shown at their natural pixel size, because it'll often be too huge to be useful! (See for example potofsoup's comic pages, linked from that thread I mentioned.) 

Well, that's what the height/width attributes are intended for, in modern versions of the HTML spec — they're supposed to indicate the _density-corrected_ intrinsic dimensions of the image (that is, the CSS pixels the image should actually occupy on screen, if the size isn't changed by any style rules).

So, we should treat those dimensions as the maximum size for the _zoomed_ state, not just the the shrunk state. That way, we shouldn't ever blast an image out beyond its useful, intended size. 

Misc considerations: 

- Need to update the "actual size?" check for disabling zoom cursors. 
- Some people still supply percents or other non-integer values to the height/width attributes, which was an HTML4 thing that's now considered illegal. Browsers still generally respect that, but it's bad practice and we won't be able to give reliable "actual size / not zoomable" behavior for it. So we should nudge people to use inline styles instead of height/width attrs, if they need to do something weird. Speaking of which: 

### Respect inline CSS on img elements

Took a while to talk myself into this one, but if an image includes any inline CSS in a `style` attribute, we should probably just exempt it from the squish:

- If someone's hand-writing their own markup, doing anything with inline styles is a strong signal of intent to exit the "fast lane" of casual image use. They clearly want it to Be A Certain Way, and the squish isn't going to save them any time, so we should just get out of the way. This is the same logic behind the table/styled-div exemption. 
- If the image embed copypasta from an image host includes inline styles, basically same argument applies — they're probably doing it to make things fit their destination container, so just let them. (Plus they're probably also using a styled div, so it'd be exempt anyway.)
- Who knows _how_ those inline styles are going to interact with our squshing styles? They're going to make our code behave unreliably, and we should only squish things when we're pretty sure we know what will happen. 

### Make images inside links zoomable? 

I always wanted to make any squished image zoomable, but I haven't managed to come up with a reliable way to handle linked images. I really wanna, though, because the lack of it is really annoying. 

potofsoup pointed out the [SMBC feed](https://smbc-comics-feed.dreamwidth.org/1103365.html) as a prime example of this. Yeah, you can click through to see the full image on the site, but still. 

### Automatically guess tall comics based on aspect ratio?

Or, maybe we could get around the link question another way. 

We constrain max height based on the viewport, because it's completely outrageous to not be able to view a single portrait or landscape image on a single screen by default. 

That's real brutal to tall comics or memes (like SMBC or QC or any 4-comi style thing), because they actually aren't an image — they're a sequence of images that happen to be delivered as a single image file. 

If we exempted tallcomi from _just_ the height maximum, a lot of them wouldn't even need to shrink! (Or would shrink a reasonable, legible amount). 

So, maybe we add a new "semi-shrunk" state that only enforces max width, and we use our JS to put that on images whose natural height is ≥ 2x their natural width.

nfagerlund:
Fixes #2845 

The quest to perfect this image-shrinking behavior continues. This time:

- Distinguish between "normal" and "tall" (height > 2x width) casual images, and
let tall images scroll instead of fitting the viewport. Nicer un-zoomed state
for vertical comic strips.

- Any image with a `style="..."` attribute is exempt from the squish. Same
logic as the decorative code exemption: inline styles means you're spending
effort to make things Just So, and the only point of the squish is to save your
time when you _don't_ want to spend effort. (Sidenote: we can't use the
width/height attributes as this kind of signal, because casual "share" copypasta
often has large w/h values.)

- Use img width/height attributes as the maximum size for the _zoomed-in_ state.
Nicer zoom behavior when using an over-resolution image to look good on retina
screens.

### DIFF
--- a/htdocs/js/jquery.imageshrink.js
+++ b/htdocs/js/jquery.imageshrink.js
@@ -3,6 +3,7 @@
 // - Zooming (on click)
 // - Exempting decorative markup from Squish (should be pure CSS, but can't yet)
 // - Hiding zoom cursors for images that won't zoom
+// - Marking tall images so they get modified squish behavior
 jQuery(function($) {
 
     // First: Basic click-to-zoom. (.imageshrink-expanded on/off)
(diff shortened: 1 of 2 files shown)


nfagerlund:
I bailed on the "zoomable links" thing because it still seems hard (and probably incredibly fiddly, even if I come up with a reliable method), but even without it this seems like a decent batch of improvements.

htdocs/js/jquery.imageshrink.js
```diff
@@ -35,26 +36,57 @@ jQuery(function($) {
...
+                // "Requested" is hard. `object-fit: contain` w/ max-height means
+                // layout width might be wider than visible width. There's no way
+                // to directly measure visible width, and of course that's the one
+                // we care about. Plus maybe they only set one attribute. So we
+                // measure indirectly: if the layout aspect ratio doesn't match the
+                // natural aspect ratio (+/- some slop bc floats), it CAN'T be the
+                // requested size. (And at least one dimension has to match.)
+                // And yes, this ignores the use case of deliberately mutilating the
+                // aspect ratio for comedy; use something other than the height/width
+                // attrs for that.
```
nfagerlund, line 63: Also, never talk to me or my son ever again.

PR #2849, opened 2020-12-03
widen ip columns in database tables

kareila:
This adds tempanonips and loginlog to the list of tables that have had their ip columns widened from 15 characters to 45. The tables for spamreports and userlog had already been widened previously.

I've tested this by running update-db.pl on my hack and making sure the column definitions were updated as expected.

### DIFF
--- a/bin/upgrading/update-db-general.pl
+++ b/bin/upgrading/update-db-general.pl
@@ -1597,7 +1597,7 @@ EOC
 register_tablecreate( "spamreports", <<'EOC');    # global
 CREATE TABLE spamreports (
     reporttime  INT(10) UNSIGNED NOT NULL,
-    ip          VARCHAR(15),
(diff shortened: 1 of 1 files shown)


zorkian:
Always spooky; do you happen to know how violent this alter is? I.e., if we do this in production, will it be a major event?

It looks like this kind of ALTER is not done online until version 5.7, if I'm reading this right: https://dev.mysql.com/worklog/task/?id=6554 We are on 5.6 right now, but it's AWS Aurora's version, so it's possible they support it? Alternately, maybe these tables are small enough that it doesn't matter?

I'll merge it, but will also need to check before we just run these against production.

kareila:
I know that the tempanonips table has been empty since June. The other one, loginlog, might be heavy, since most users have at least one remembered login session. Gotta stop truncating IPv6 addresses, though.

PR #2850, opened 2020-12-07
fixes for mobile pages

### LINKED ISSUE #2833: mobile post page needs updating for new password checking code
I think this was mentioned before as a low priority issue but never got a ticket opened. Now it's come up in support: https://www.dreamwidth.org/support/see_request?id=42375

kareila:
Fixes #2833.


zorkian:
Another password use bites the dust. Nice!

PR #2851, opened 2020-12-07
reformat converted FAQ Search page

kareila:
I think this is an improvement? The old layout really wasn't working for me.

Before:
Screen Shot 2020-12-07 at 4 52 51 PM

After:
Screen Shot 2020-12-07 at 4 53 09 PM

### DIFF
--- a/htdocs/stc/faq.css
+++ b/htdocs/stc/faq.css
@@ -4,5 +4,7 @@
     list-style-position: outside;
 }
 
-.summary {font-weight: bold;}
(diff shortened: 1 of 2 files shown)


momijizukamori:
Works for me - I think I just pulled in the default Foundation forms styling, which is.... questionable.

nfagerlund (approved):

kareila:
Going to go ahead and merge this since Momiji and Nick both signed off on it. Thanks!

PR #2852, opened 2020-12-09
convert /support/see_request

kareila:
@momijizukamori gave me her WIP branch for support page conversion to try to push across the finish line. Here's the conversion for /support/see_request. I'm a little nervous about this one since it's such an important page, so if anyone else wants to take it for a spin, I'd be grateful.


momijizukamori:
Quick runthrough of behavior looks good, though I didn't check any varying permissions edge-cases. Nitpicky upgrade request - `find=cnext` and `find=cprev` should use the new message flashing behavior when a next/prev request can't be found, instead of displaying a single-sentence error page.

(I also still kind of hate how this page looks but it can be gussied up after we convert to TT)

kareila:
I think I would prefer obvious view/functionality changes go through separately for pages that are heavily used like this one, so we can address changes individually and be able to roll them back easily if needed.

momijizukamori:
Extremely valid point!

cgi-bin/LJ/Lang.pm
```diff
@@ -612,7 +612,7 @@ sub get_text {
         ? $from_files->()
         : $from_db->();
 
-    if ($vars) {
+    if ( defined $vars && %$vars ) {
```
zorkian, line 615: nit: This still has some edge cases (defined is true for arrayrefs for example but then will fail when it's coerced into a hash?) so the typically way of doing this is,

```perl
if ( ref $vars eq 'HASH' && %$vars ) {
```

zorkian:
I spent entirely too long trying to figure out why there was no form auth... because it's just `see_request.bml` oh.

zorkian (commented):
One nit but I merged, ty!

PR #2853, opened 2020-12-10
Convert inbox to TT

momijizukamori:
Mostly a straight convert of the BML with HTML/CSS for nicer responsive handling, but a few notes!
- This includes some tweaks to make the jquery autocomplete 1) look nicer inline and 2) be able to take a static array instead of just an endpoint to call back to
- the `no_proto_img` stuff is to make management buttons in the inbox cooperate with our newer JS without having to alter `LJ::img` which is a lot of places and thus likely to have weird knock-on effects
- I had some weird issues getting strings in `index.tt.text` to show up in `index.tt`? Worked around it by using existing strings from en.dat the two spots I needed them.


kareila:
Checks failed - does this need to be retidied?

momijizukamori:
Hmm, thought I ran tidy before I pushed. I'll double-check and see.

views/inbox/compose.tt
```diff
@@ -0,0 +1,76 @@
+[%- CALL dw.active_resource_group("foundation") -%]
+
+[% dw.need_res({ group => "foundation"}, 
+    'stc/css/pages/inbox.css'
```
zorkian, line 4: I'm vaguely surprised this list of files doesn't need trailing commas?

nit: using single quotes for some lines but double quotes for others kills the Mark

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
+package DW::Controller::Inbox;
+use strict;
```
zorkian, line 2: This file is missing the header copyright notices and also the normal preamble (use v5.10 etc).

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+use DW::Routing;
+use DW::Template;
+use DW::FormErrors;
+use LJ::Hooks;
+use Data::Dumper;
+
+DW::Routing->register_string( '/inbox',          \&index_handler,    app => 1 );
+DW::Routing->register_string( '/inbox/compose',  \&compose_handler,  app => 1 );
+DW::Routing->register_string( '/inbox/markspam', \&markspam_handler, app => 1 );
+DW::Routing->register_rpc( "inbox_actions", \&action_handler, format => 'html' );
```
zorkian, line 15: nit: If you aren't interpolating variables, use single-quotes for strings.

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+    my $POST   = $r->post_args;
+    my $GET    = $r->get_args;
+    my $remote = $rv->{remote};
+    my $vars;
+    my $scope = '/inbox/index.tt';
+
+    return error_ml("$scope.error.not_ready") unless $remote->can_use_esn;
+
+    my $inbox = $remote->notification_inbox
+        or return error_ml( "$scope.error.couldnt_retrieve_inbox", { 'user' => $remote->{user} } );
```
zorkian, line 33: nit: Quoting `'user'` here is unnecessary, and we prefer not to stylistically (reduces clutter). You only need to quote keys in hashes when the keys start with numbers or contain certain symbols.

zorkian: I just realized that a lot of this code came from the BML. You can ignore most of the nits... I don't need us to clean up the Perl we're copying from BML...

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+        my $event_time = $item->when_unixtime;
+        $cleaned->{rel_time} = LJ::diff_ago_text($event_time);
+        $cleaned->{abs_time} = LJ::S2::sitescheme_secs_to_iso( $event_time, { tz => "UTC" } );
+
+        my $contents = $item->as_html || '';
+
+        my $expanded = '';
+
+        if ($contents) {
+            BML::ehtml( \$contents );
```
zorkian, line 168: Should this really call BML? Isn't there an equivalent LJ function?

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+
+    # gets the request and args
+    my $r         = $rv->{r};
+    my $args      = $r->json;
+    my $action    = $args->{action};
+    my $ids       = $args->{'ids'};
+    my $view      = $args->{view} || 'all';
+    my $page      = $args->{page} || 1;
+    my $itemid    = $args->{itemid} || 0;
+    my $remote    = LJ::get_remote();
```
zorkian, line 271: Can't this come out of `$rv` instead of calling out to `LJ::get_remote()`?

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+    my $form_auth = $args->{lj_form_auth};
```
zorkian, line 272: This `$form_auth` seems to be unused so now I'm wondering if this needs to be checked in the `controller()` call above? I.e., if this handler is only invoked on POST, we probably need to validate the form auth?

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+    my $msg_limit     = $remote->count_usermessage_length;
+    my $subject_limit = 255;
+    my $force = 0;                     # flag for if user wants to force an empty PM
+    my $scope = '/inbox/compose.tt';
+
+    # Submitted message
+    if ( $r->did_post ) {
+        my $mode = $POST->{'mode'};
+        $errors->add( undef, 'error.invalidform' )
+            unless LJ::check_form_auth( $POST->{lj_form_auth} );
```
zorkian, line 392: This is unnecessary since you passed `form_auth => 1` to the controller call.

cgi-bin/DW/Controller/Inbox.pm
```diff
@@ -0,0 +1,659 @@
...
+        $r->ERROR, "$LJ::SITEROOT/inbox" )
+        if $msg->type eq "out";
+
+    return $r->msg_redirect( "You are not allowed to report messages as spam.",
+        $r->ERROR, "$LJ::SITEROOT/inbox" )
+        if LJ::sysban_check( 'spamreport', $remote->user );
+
+    if ( $r->did_post && $POST->{'confirm'} ) {
+        $errors->add( undef, 'error.invalidform' )
+            unless LJ::check_form_auth( $POST->{lj_form_auth} );
```
zorkian, line 630: Unnecessary check.

zorkian (changes requested):
Some conversion comments that need to be addressed, one major one around form auth, and a few other places with similar comments. I started to nit on formatting but since most of this is copied from the BML, we don't need to address nits on formatting that code.

I didn't really spend a ton of time on the SCSS/JS/HTML, so that's something we'll want to either have other people dig into or just plan to push this when we're around to do fixes live.

zorkian:
Yikes how did I never circle back to this :< sorry.

zorkian (approved):

PR #2855, opened 2021-01-07
Use Math::Random::Secure for auth generation

rshatch:
Our previous settings were proooobably sufficiently random, but it's better to be safe and use a more clear-cut good source of randomness.


zorkian:
And just to confirm, this doesn't depend on using actual-entropy for every random number, right? Just uses it for the seed or something? (Basically, want to make sure it works at scale, given we don't have any good high volume source of entropy.)

rshatch:
Yep, this just uses true entropy as the seed (which we were already doing ourselves the hard way, I think, and then uses a stronger generator function than the default Perl one. It also provides an `irand` function which is apparently faster than `int(rand())` (as well as being tidier), so nice bonus!

zorkian:
Yeehaw, ty!

PR #2856, opened 2021-01-23
Support Ubuntu 20.04 and MySQL 8

zorkian:
This modernizes a few things to support more recent versions of MySQL
and Ubuntu. DO NOT RUN IN PRODUCTION YET.


bin/dump-data.pl
```diff
@@ -144,7 +144,9 @@ sub db_quote {
 print "Dumping proplists.dat\n";
 open( my $plg, ">$ENV{LJHOME}/bin/upgrading/proplists.dat" )       or die;
 open( my $pll, ">$ENV{LJHOME}/bin/upgrading/proplists-local.dat" ) or die;
-foreach my $table ( 'userproplist', 'talkproplist', 'logproplist', 'usermsgproplist' ) {
+foreach my $table ( 'userproplist', 'talkproplist', 'logproplist', 'usermsgproplist',
+    'media_prop_list' )
```
zorkian, line 148: oops forgot this a long time ago ...

bin/dump-data.pl
```diff
@@ -181,7 +183,7 @@ sub db_quote {
 # and dump mood info
 print "Dumping moods.dat\n";
 open( F, ">$ENV{'LJHOME'}/bin/upgrading/moods.dat" ) or die;
-$sth = $dbh->prepare("SELECT moodid, mood, parentmood FROM moods ORDER BY moodid");
+$sth = $dbh->prepare("SELECT moodid, mood, parentmood, weight FROM moods ORDER BY moodid");
```
zorkian, line 186: don't know when this became a thing, but moods.dat contains it so we should dump it.

bin/dump-tablecreates.pl
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/perl
+
+use strict;
```
zorkian, line 3: dumb script for regenerating update-db-general.pl

bin/upgrading/moods.dat
```diff
@@ -130,760 +130,6 @@ MOOD 131 thankful 15 80
 MOOD 132 grateful 15 85
```
zorkian, line 130: This file was re-dumped, which put things in order...

bin/upgrading/proplists.dat
```diff
@@ -62,14 +62,6 @@ userproplist.city:
   multihomed: 0
```
zorkian, line 62: Should be no actual changes to this file, but I re-dumped it so everything is ordered now instead of just manually updated lol.

bin/upgrading/update-db-general.pl
```diff
@@ -18,4190 +18,2655 @@
 
 mark_clustered(@LJ::USER_TABLES);
 
-register_tablecreate( "vgift_ids", <<'EOC');
```
zorkian, line 21: This file has been cleanly dumped from an updated database, using MySQL 8. This basically compacts all of the alters so they're gone, which means that it should be much easier to reason about what's in the db.

cgi-bin/DW/Auth/Challenge.pm
```diff
@@ -67,7 +67,7 @@ sub check {
         }
         else {
             my $dbh = LJ::get_db_writer();
-            my $rv  = $dbh->do( q{SELECT GET_LOCK(?,5)}, undef, $chal );
```
zorkian, line 70: 64 character limit on lock names now. Hacky easy way to do that.

cgi-bin/DW/VirtualGiftTransaction.pm
```diff
@@ -71,7 +71,7 @@ sub save {
 
     my $u       = LJ::want_user( $opts{user} ) or return;
     my $id      = LJ::alloc_user_counter( $u, 'V' ) or return;
-    my $secs    = $opts{time} || localtime;
```
zorkian, line 74: Was trying to insert 'Sat Jan 23...' string into the int field expecting an epoch time.

cgi-bin/LJ/Global/Defaults.pm
```diff
@@ -260,14 +260,7 @@ no strict "vars";
     # 'type' is used as 'type' attribute value in 'codes' table
     # 'save_region_code' specifies what to save in 'state' userprop  -
     # '1' mean save short region code and '0' - save full region name
-    %LJ::COUNTRIES_WITH_REGIONS = (
```
zorkian, line 263: I killed RU here because... we don't do it for Canada, so whatever. (And inserting the plain RU states was causing the strict mode SQL to be angry since it's not latin1.)

cgi-bin/LJ/User/Account.pm
```diff
@@ -75,9 +75,9 @@ sub create {
     my $dbh = LJ::get_db_writer();
 
     $dbh->do(
-        'INSERT INTO user (user, clusterid, dversion, caps, journaltype) '
```
zorkian, line 78: Strict mode fix, name field is required.

cgi-bin/LJ/User/Administration.pm
```diff
@@ -238,18 +238,13 @@ sub banned_userids {
 
 sub ban_note {
     my ( $u, $ban_u, $text ) = @_;
-    my @banned;
 
-    if ( ref $ban_u eq 'ARRAY' ) {
```
zorkian, line 243: Make this function take user objects or ids, because there are callers that do both.

doc/dependencies-system-20.04
```diff
@@ -0,0 +1,21 @@
...
+libpng-dev
+perlmagick
+libapache2-mod-perl2
+libapache2-request-perl
+libdbd-sqlite3-perl
+libdatetime-perl
+libio-aio-perl
+libssl-dev
+apache2
+compass-blueprint-plugin
```
zorkian, line 20: This was `ruby-compass`, it got renamed. Only change from 18.04 list.

t/console-moodthemes.t
```diff
@@ -36,7 +36,7 @@ my $run = sub {
 
 # FIXME: make this neater.
 ok( $run->("moodtheme_list") =~ "Kanji Moods", "Got public theme" );
-ok( $run->("moodtheme_list 1") =~ "18x18 /img/mood/kanji/crazy.gif", "Got a theme" );
```
zorkian, line 39: With the re-dumping of moods.dat, the Kanji Moods is #11 now in a fresh db. Won't matter for old dbs.

t/post.t
```diff
@@ -48,9 +48,6 @@ note("Not logged in - init");
     # user
     ok( !$vars->{remote} );
 
-    # icon
```
zorkian, line 51: Icons were removed by @momijizukamori but the test still expected them, so just removed that.

t/userpic-keyword-select.t
```diff
@@ -31,14 +31,6 @@ my $ICON1 = do { local $/; <$fh> };
 open( my $fh2, 'good.jpg' ) or die $!;
 my $ICON2 = do { local $/; <$fh2> };
 
-note("called with falsy or undefined user object");
```
zorkian, line 34: I don't know why this test, and it fails in Perl 5.30, so I removed it. But LMK if I'm missing something @nfagerlund !

nfagerlund: Oh, that broke when Ruth changed that from a standalone function that takes a user as an argument to a method on a user object -- can't call methods on undef. I think this is a merge conflict now because I noticed that in one of those post-code-push PRs and removed it there.

PR #2859, opened 2021-01-31
Reconcile icon select component with beta entry page

nfagerlund:
CODE TOUR: Fixed a bug where the beta create entries page wasn't properly saving the icon you chose. We recently started handling the icon menu more consistently, and one place that used it still really wanted to do things differently.

#2840 left the entry page using a one-off implementation of the icon select menu, but it didn't quite match what the controller expected. So, let's just sort out the controller so it can handle the component version properly — the component needs to get one extra conditional, and the controller gets slightly more verbose, but this way we only need to maintain one version of that `form.select` call.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -254,6 +254,10 @@ sub new_handler {
     );
     $vars->{formdata}->{editor} = $vars->{editors}->{selected};
 
+    # Set up info for the icon select/preview/browse components
+    $vars->{current_icon_kw} = $vars->{formdata}->{prop_picture_keyword};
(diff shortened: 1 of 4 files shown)


nfagerlund:
Any of y'all can review most of this PR, but the bits we should probably check with @momijizukamori on are

- Removing class="inline" on that select. (It looks like that was detritus from when you were trying to fit it in somewhere? It doesn't seem to have any effect on the comment forms on site skin or journal skins, and it had a bad effect on the entry page.)
- Making the label for the select invisible again. (I'm not sure why you switched it to visible+left, and there wasn't any note about that in the PR.)

momijizukamori:
Both those make a difference for the inbox form (which is where I started when I was working on the components), but if that's the odd one out (which it sounds like it is) I can manually override the styling on that page alone.

PR #2860, opened 2021-01-31
Hack: fix entry page icon chooser (mostly)

nfagerlund:
CODE TOUR: no-impact

(^^ since edits are still bugged, we're going to need to do SOME kind of better fix for this, even if it's not my PR. Whatever that fix is, it can report for the code tour. )

This is the hacky version of https://github.com/dreamwidth/dw-free/pull/2859. Just get posts and edits working again, even though the icon preview button doesn't show what it oughtta show when you start an edit. 

These commits are already part of #2859, so probably github will just swallow any merge conflicts if you merge that one later.


nfagerlund:
Superseded by the non-hack version 🎉

PR #2863, opened 2021-01-31
Fixes for S2 pagination

momijizukamori:
- Change styling to make old version look not completely terrible
- Reorganize page list into nest list for easier styling
- Add visited link colors
- Fix view for < 10 pages

### DIFF
--- a/htdocs/scss/components/foundation-custom/_pagination.scss
+++ b/htdocs/scss/components/foundation-custom/_pagination.scss
@@ -10,3 +10,7 @@
 ul.pagination {
     display: inline-block;
 }
+
+ul.pagination li a {
+    font-weight:bold;
+}
--- a/htdocs/scss/skins/_entry-styles.scss
+++ b/htdocs/scss/skins/_entry-styles.scss
@@ -203,6 +203,7 @@ ul.pagination {
   grid-template-columns: auto auto auto;
   grid-template-areas: "prev pages next";
   text-align: center;
+  margin: 0 0 0.75em 0;
 
 }
 
@@ -210,7 +211,7 @@ ul.pagination .arrow {
   align-self: center;
   justify-self: center;
   font-size:  large;
-  padding: .0625rem .25rem .0625rem;
+  margin: 0 .25em;
   &.disabled {
     opacity: 0.33;
     padding: .0625rem .25rem .0625rem;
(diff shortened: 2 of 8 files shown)


nfagerlund:
Here's the stuff I'm noticing so far: 

### Lynx

Screen Shot 2021-01-31 at Jan 31, 2 45PM

^^ Badness on Lynx -- ul.pages needs to set explicit padding (probably 0?) instead of getting the default for uls. 

There's probably some other stuff where we're not resetting enough defaults,

### Margins

ul.pagination's margin value is bad (foundation defaults), and pulls it over to the left, THROUGH the padding of the .inner that contains it:

Screen Shot 2021-01-31 at Jan 31, 2 58PM

If you set it to total 0, it looks like this: 

image

Which I don't mind, but you might want some margin on the bottom, idk. (`0 0 1rem 0`) Actually, maybe both top and bottom margin (`1rem 0`), and ought to do margin-collapsing on the top margin and the bottom margin of the `p` above. 

### Mismatched styles on inactive arrows

Inactive arrows aren't wrapped in an <a>, they're just a text node... and that makes them inconsistent in width/padding/margin/etc. Easiest to see it if you turn the grid lines on:

Screen Shot 2021-01-31 at Jan 31, 3 00PM

You could dink around with the styles until stuff lines up, or you could wrap the inactive arrow in an `a` without an `href`. That looks like this: 

Screen Shot 2021-01-31 at Jan 31, 3 01PM

It gets a faint background on hover, but imo that's fine?

### Text baselines

If there's an odd number of rows in the pagination list, the text baselines of the back/forward arrows don't line up with those in the middle row of links. 

I vaguely remember fighting this for... way too long... and not coming up with a satisfactorily reliable answer, so maybe it should wait until we have a chance to divorce this grid/flex pagination component from the `.pagination` class that Foundation reserves for itself and contaminates with a ton of extremely cursed float styles.

nfagerlund:
The code itself looks fairly solid, although there's some spots that could use some tiny tidying (extra line breaks right before a block of properties ends, like). 👍🏼 

My broader take in the channel, which I'll repeat here, is that building this on top of the same classes that are used by Foundation's pagination component isn't good or sustainable, and those styles will continue to be unexploded ordnance scattered around our back lawn. So, we should probably change the class names and finish turning this into a standalone implementation. But not in this PR.

nfagerlund (approved):

PR #2865, opened 2021-01-31
[#2861] convert WHILE to FOREACH in views/profile/_blocks.tt

### LINKED ISSUE #2861: converted profile page breaks with >1000 subscriptions?
This was always a problem (see `%LJ::FORCE_EMPTY_
SUBSCRIPTIONS`) but after pushing the converted profile display code, this error was seen in the logs: `va-web02: undef error - WHILE loop terminated (> 1000 iterations)`

From the TT documentation:

> The Template Toolkit uses a failsafe counter to prevent runaway "WHILE" loops which would otherwise never terminate. If the loop exceeds 1000 iterations then an "undef" exception will be thrown, reporting the error:
```
        WHILE loop terminated (> 1000 iterations)
```
>    The $Template::Directive::WHILE_MAX variable controls this behaviour and can be set to a higher value if necessary.

So a dirty fix would be to go into views/profile/_blocks.tt and up that WHILE_MAX limit to something like 5000, unless we can think of a different approach that wouldn't bang up against weird TT loop limits.

kareila:
WHILE has a hidden limit WHILE_MAX which will terminate a loop after 1000 iterations. FOREACH has no such restrictions.

Fixes #2861.

CODE TOUR: The logic loop for profile pages that would print the list of OpenID accounts on a user's access list was running up against a hidden limit where the loop would exit if the list was longer than 1000 items. This changes it to use a different type of logic loop instead, which avoids the problem.

### DIFF
--- a/views/profile/_blocks.tt
+++ b/views/profile/_blocks.tt
@@ -109,9 +109,7 @@
   [% dw.img( 'id_openid', '' ); "[$s] "; %]
     [%- site_users = [];
         i = 0;
-        i_max = openid_info.sites.$s.size;
-        WHILE i < i_max;
-          user = openid_info.sites.$s.$i;
+        FOREACH user IN openid_info.sites.$s;
           text = openid_info.shortnames.$s.$i;
           linked_u = linkify( { url => user.profile_url, text => text } );
           PROCESS format_userlink;


views/profile/_blocks.tt
```diff
@@ -109,9 +109,7 @@
   [% dw.img( 'id_openid', '' ); "[$s] "; %]
     [%- site_users = [];
         i = 0;
-        i_max = openid_info.sites.$s.size;
-        WHILE i < i_max;
-          user = openid_info.sites.$s.$i;
+        FOREACH user IN openid_info.sites.$s;
           text = openid_info.shortnames.$s.$i;
```
nfagerlund, line 113: Hmm, and we're sure the $i will always line up with the index of the user we're tracking? (It looks probably fine, I just have an innate heeby-jeeby reaction to tracking two arrays in parallel like that. I bet that's also just logic translated verbatim from the BML conversion.)

nfagerlund: Yeah, okay, I read the parse_openids function, it should be fine. 👍🏼

nfagerlund (approved):

views/profile/_blocks.tt
kareila: Yup, that's the one thing I don't like about this change - it makes the parallelism of the two arrays less obvious to the casual observer. But it doesn't seem worth reconstructing the entire underlying data structure to make it clearer.

PR #2871, opened 2021-02-05
Have /data/edges return an error if accessed without first logging in

kareila:
CODE TOUR: This is a rarely used API endpoint with the potential to slow down the site for everyone if a badly behaved program accesses it repeatedly. To reduce the likelihood of abuse, only return the requested data if the user is logged in.

(This is currently returning 404 for all requests in production, but we can lift that restriction once this gets deployed.)

### DIFF
--- a/cgi-bin/DW/Controller/Edges.pm
+++ b/cgi-bin/DW/Controller/Edges.pm
@@ -44,6 +44,9 @@ sub edges_handler {
         return $r->OK;
     };
 
+    # Don't allow access unless we have a logged-in user
+    return $error_out->( 404, 'Must log in first' ) unless LJ::get_remote();
+
     # Load the account or error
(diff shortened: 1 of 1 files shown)


cgi-bin/DW/Controller/Edges.pm
```diff
@@ -44,6 +44,9 @@ sub edges_handler {
         return $r->OK;
     };
 
+    # Don't allow access unless we have a logged-in user
+    return $error_out->( 404, 'Must log in first' ) unless LJ::get_remote();
```
zorkian, line 48: This is totally fine, I just am slightly weirded out by this function not invoking `controller(...)`, but maybe that's not super helpful or this kind of JSON thing? Weird tho, I thought all of our controllers called it... TIL.

PR #2873, opened 2021-02-06
Add PR template explaining the CODE TOUR: line

nfagerlund:
CODE TOUR: Meta: encourage devs to leave hints for whoever writes the next code tour post! When devs submit code, they include a description of what they did... but it's intended for other developers on the same project, so it can often be opaque even to devs that work on other code. This just adds a reminder to also include a brief summary for non-devs in there.

We've all started doing this, and it seems like a good idea, so let's go ahead and codify it!

This is a magic file that GitHub uses to auto-populate the PR description textarea, so that a new WIP PR looks like this: 

image

It also still does the thing where it includes the commit message if it's a single-commit PR; IIRC that goes below the template content?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,9 @@
+CODE TOUR: summarize here.
+
+<!--
+Above, please explain how these changes affect users. This summary must be a single line that starts with the text `CODE TOUR:`.
+
+The rest of your PR description is for other developers, but the CODE TOUR line is for whoever writes the next code tour post. See https://dw-dev.dreamwidth.org/tag/code+tour for examples.
+
+If this PR has no direct effect on users, write a summary anyway, but include the text `no-impact` as a hint for sorting.
+-->


zorkian:
Oh fancy. Didn't know about this!

Sophira:
Since this is a thing, it might be worth having someone modify the [lib/bin/codetour.py](https://github.com/dreamwidth/dreamhack/blob/master/lib/bin/codetour.py) file in the Dreamhack repository to include this hint in the generated code that can be copied and pasted!

nfagerlund:
@Sophira I think a couple of us are kind of diffusely WIP on a re-work of surrounding code tour tooling -- I've got like... idk, 20% of a replacement for the python script that takes advantage of these lines, and what would be best would be to integrate it into a discord bot that nags the channel every, say... month, or ten PRs, whichever comes first? something like that? and dumps a starter text file as an attachment. Uh, we'll see how it shakes out!

PR #2874, opened 2021-02-06
remove paragraph tags from explanatory CSS comments

kareila:
I don't know if this is identical to the emergency fix in prod, but some version of this needs to live on master since most styles are broken without it.

CODE TOUR: The fix from #2867 included some CSS comments that contained a paragraph tag, to explain the impact of the change. Unfortunately, our less-than-ideal CSS parser choked on rendering CSS that appeared to contain HTML tags. This rewrites the comments to avoid the problem.


nfagerlund (approved):
👍🏼😭

PR #2877, opened 2021-02-11
cleanup from latest changes

kareila:
CODE TOUR: no impact

Two small fixes: an ALTER TABLE statement was acting on the wrong table, and a test suite needed to be updated to accommodate recent refactoring.


zorkian:
As always, thank you!

PR #2879, opened 2021-02-12
Knock it off with the Foundation table styles

nfagerlund:
CODE TOUR: The new site skins were adding *too much style juice* to tables in entries and comments. (Messing with font/background colors, adding unexpected amounts of padding, etc.) Now they're not. Hopefully.

Yanks the default Foundation style for tables, and re-adds it as an opt-in style tied to `class="table"`, then adds that class on all of the templates in views/ that were _probably/maybe_ already depending on Foundation table styles. This mimics prior art with buttons, where to get the Foundation site skin look you usually need to add `class="button"`. 

- Yes, abusing tables for decorative markup is Bad '90s Nonsense. 
- Yes, using tables like that means your stuff will inevitably display wrong on the reading page anyway, so why are people surprised when it _also_ looks bad on entry pages?
- Yes, they should definitely be using divs to do this kind of stuff.

That said, may as well try not to break existing cultural practices. 

<details><summary>Screenshots</summary>

**Old styles:** 

Screenshot_2021-02-11 betagal time to do everything uaoedfythquxoeuiaeioeuiaeouuqjkuaibqkxbihfyefhdxjbkxthdncgrghthdlidhutd

**New styles after this patch:** 

Screenshot_2021-02-11 betagal time to do everything uaoedfythquxoeuiaeioeuiaeouuqjkuaibqkxbihfyefhdxjbkxthdncgrghthdlidhutd (1)

Actually this is even more minimalist than the old ones, because I didn't add back the extra grey backgrounds on `<th>` cells. The bold text seems sufficient.

**New styles before this patch:** 

Screenshot_2021-02-11 betagal time to do everything uaoedfythquxoeuiaeioeuiaeouuqjkuaibqkxbihfyefhdxjbkxthdncgrghthdlidhutd (2)

Too spicy for the RP userbase.

</details>

### DIFF
--- a/htdocs/scss/components/foundation-custom/_tables.scss
+++ b/htdocs/scss/components/foundation-custom/_tables.scss
@@ -1,9 +1,25 @@
-th[scope="row"] {
-    text-align: left;
+$include-html-table-classes: false;
+@import 'foundation/components/tables';
+
+// You can opt into Foundation-styled tables by adding class `table`, but
+// turning them on by default makes a dumb mess (just due to the site's nature).
+table.table {
+    @include table;
+
+    th[scope="row"] {
+        text-align: left;
+    }
+
+    th label {
+        color: $table-head-font-color;
+    }
 }
 
-table th label {
-    color: $table-head-font-color;
+// Alternately, here's the padding the old site skins used.
+table:not(.table) {
+    td, th {
+        padding: 0.2em 0.8em;
+    }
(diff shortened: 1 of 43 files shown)


zorkian:
I think that this is better for sure, but some small part of me just feels bad about `table class="table"`. I cannot think of a better way to solve this issue though...

Issue #2189, opened 2017-05-24
imzy shutting down at end of june


rahaeli:
https://www.imzy.com/imzy/post/imzy_is_shutting_down

And we just added it, too. :(

kareila:
What's the best way to handle this? It seems to me that people would find value in continuing to refer to Imzy usernames after the site shuts down, even though the associated links will stop working.

rahaeli:
With past shutdowns I believe we removed everything. Inksome is an example, maybe we kept the username tag?

kareila:
Yes, we kept the user tag, so just remove the profile field and related userprop.

The badge_icon links to https://www.imzy.com/favicon.ico so we'll have to cache that before it goes away.

pauamma:
Snarfed https://www.imzy.com/favicon.ico locally, in case no one did before.

kareila:
Pau, can you upload your copy of the image as an attachment on this issue?  Thanks.

pauamma:
[imzy_favicon.zip](https://github.com/dreamwidth/dw-free/files/1110328/imzy_favicon.zip)

There. (Had to ZIP it, because GHI wouldn't let me upload a .ico file.) Will keep it around until you tell me it's OK to delete.

kaberett:
Claim!

@pauamma, thank you -- I've grabbed a copy and am intending to deal with this tomorrow.

PR #2880, opened 2021-02-13
Update external site constructor to use https

### LINKED ISSUE #2627: update AO3 user links to use HTTPS
https://www.dreamwidth.org/support/see_request?id=41640

Someone points out that while http://archiveofourown.org/ redirects to HTTPS, http://www.archiveofourown.org does not, and we use the latter when constructing external username links. That suggests that the bare domain without the WWW should be considered more canonical, but in either case, we should adjust our link constructor to use HTTPS links.

woggy:
Fixes #2627

CODE TOUR: User links to external sites will now use https, instead of http.

### DIFF
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -254,7 +254,7 @@ sub journal_url {
     # AND YOU WILL SEE WHAT INFINITE RECURSION LOOKS LIKE.
 
     # override this on a site-by-site basis if needed
-    return "http://$self->{hostname}/users/" . $u->user . '/';
+    return "https://$self->{hostname}/users/" . $u->user . '/';
 }
 
 # returns an entry link for this user on this site.


zorkian:
Welcome back to Dreamwidth development! :D

So, this fix is right-one-maybe-wrong-place. The function you changed, `DW::External::Site->journal_url` is actually used by a whole bunch of different external sites, not just AO3! Do you know if every external site we use supports HTTPS? Or, did you want to only change this for AO3's links?

If you want to change it only for AO3, what you'll need to do is copy this whole `journal_url` function and put a copy of it into the file `cgi-bin/DW/External/Site/ArchiveofOurOwn.pm` -- that is the specific class for AO3 links. Today, that class doesn't set a custom `journal_url` function (it doesn't have one at all!) so it just uses the base class, which is the one you modified here.

Does that make sense? I'm happy to describe this further if it'd be helpful, class inheritance is a tricky thing since it involves stuff happening magically in places where no code actually exists :)

woggy:
Oh, no, that makes perfect sense! I think it makes sense to change it for everything that we can (I was in fact sorts hoping for the inheritance), and was sorta assuming that all of the external sites support HTTPS, in this day and age. I'll go and check to make sure that assumption is correct.

zorkian:
TBH, I'd be super happy to switch the default! And then if anything is broken we could just patch those particular sites to HTTP? Wdyt?

woggy:
Having gone through the list of every site that's listed in Site.pm; almost everything is good to go with HTTPS. There are only three exceptions:

* DeadJournal exists but has an invalid site certificate.
* Inksome appears to no longer exist as a site; the page is an (https) domain reseller splash.
* Imzy is giving me an ERR_CONNECTION_REFUSED, in both https and http addresses.

So I think we should be good to default to https, possibly with an override for DeadJournal. Or possibly not, since they're set *for* https, just not correctly right now.

(Also, as a curiosity/learning experience - why do the siteid numbers in that list start with 2? is 1 the "this site, not external" reserve, or something else?)

kareila:
We know Imzy and Inksome went away, but decided to keep the user tag constructors around for historical reasons. (See #2189.) So it makes sense to let those inherit whatever the current default is. And yeah, I wouldn't override DeadJournal either - that's on them to fix their certificate.

As for your other question, the siteids appeared with e81f04746e and thus appear to be linked to how the crossposter code is set up. So yeah, 1 is probably "the current site" since it is possible to crosspost from one DW account to another.

zorkian:
Awesome, thank you for checking these! Yay for having the default be https :)

PR #2885, opened 2021-02-28
Add errors for bad markup to beta create page

momijizukamori:
CODE TOUR: Previously, if you had bad markup in an entry, you wouldn't find out until after you had already posted the entry and said bad markup was potentially on display for the world to see. Now the beta create page will give you the error before the post is saved, so you can preemptively fix it (note: this doesn't catch *all* bad markup - but it gets the most common cases)

This will warn for irreparable invalid markup, and unclosed tags that require closing (other than `<p>` and `<cut>`, because it's common to use those unclosed and they have well-defined behavior). At the moment the error is blocking - I'm open to the possibility of adding a way to say 'post anyway' but I would prefer to see if anyone complains and in what scenario first, because that logic is potentially a bit complicated (as I don't think we should allow irreparable markup to go through).

I'm also open to feedback on error strings, I mostly just grabbed the existing ones from HTMLCleaner.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -713,19 +713,20 @@ sub _form_to_backend {
     my $clean_event = $post->{event};
     my $errref;
 
-    # TODO: accept editor prop and thread it through to the cleaner.
     my $editor = undef;
     my $verbose_err;
-    LJ::CleanHTML::clean_event( \$clean_event, { errref => \$errref, editor => $editor, verbose_err => \$verbose_err } );
+    LJ::CleanHTML::clean_event( \$clean_event,
+        { errref => \$errref, editor => $editor, verbose_err => \$verbose_err } );
 
-    if ($errors && $verbose_err) {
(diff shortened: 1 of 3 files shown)


zorkian:
I'm fine with it being blocking too, and we can see. It definitely makes more sense there.

Altho... how does this intersect with synsuck and such? Do those also get rejected if they have bad content? That is one case where I think it might be nice to allow thru the post so it doesn't just not show up?

kareila:
The only effect I see in the main HTML cleaner is to make more error data available to the caller. The changed behavior provided to the user is only in the beta Create Entries page. If we have any syndicated feeds using that, they need to go to their room and think about their choices.

momijizukamori:
Yep, kareila's got the right of it - cleaner changes are only to expose more data, and places that use the cleaner will have to explicitly ask for the more verbose errors and then decide what to do with it - otherwise they'll get the existing behavior. And the beta create page right now is the only one asking for and using that info.

(and I just triple-checked - the change is in the form-to-backend functionality, which I am 99% sure isn't on the post-by-email path)

PR #2886, opened 2021-02-28
Upgrade modernizr to 3.11.4

### LINKED ISSUE #2878: Possible issues from antique modernizr
Report: https://dw-beta.dreamwidth.org/14656.html?thread=500288#cmt500288

At least one report of weird visual glitches in chrome on the beta create entries page. They were able to dig into the inspector, and found positive proof that modernizr was (sometimes!) marking their browser as unable to handle `::before/after` (the `(no-)generatedcontent` class on the root element). 

I looked into it, and the current version of that check includes a margin of error for weird non-deterministic rounding glitches linked to zoom behavior. Our version does NOT. 

I’m not sure what all is on/off in our “modernizr-custom.js” but if we can recover that info we should probably upgrade it. 

(Obligatory note that the modern web has much better tools for this sort of thing and modernizr is kind of silly, but a bunch of foundation stuff relies on it, so. At some point it will be smarter to just lock all those classes to always-on, but probably not today.)

nfagerlund:
Fixes #2878 ......maybe. The story there is that modernizr checks for
`::before/after` support by measuring the height of something, and some browsers
give slightly sloppy pixel measurements of things if the zoom level is tweaked.
So the newest versions of modernizr account for that in their measurement, but
our "vintage" version didn't.

Anyway, the hope is that this'll deal with that for now, but really let's:

- Not increase our reliance on modernizr at all
- Transition stuff off it as the opportunity arises

In that spirit, this update is a custom build that pares modernizr back to the
veritable bones, _only_ including what we (or Foundation) appear to CURRENTLY be
using. The custom build config looks something like:

```javascript
modernizr.build({
  minify: true,
  enableJSClass: true,
  enableClasses: true,
  classPrefix: "",
  options: [
    'html5shiv',
    'addTest',
    'setClasses',
  ],
  "feature-detects": [
    "css/transforms",
    "css/transitions",
    "css/flexbox",
    "css/fontface",
    "css/generatedcontent",
    "touchevents",
  ],
}, output => {
  // Hack: alias (bad, deprecated) touchevents feature to its even-older name.
  output += "\n; Modernizr.addTest('touch', Modernizr.touchevents);";
  // something to write the file
})
```

Theoretically this update and the associated removal of unused code shouldn't
break anything... but keep an eye out.

CODE TOUR: Some Chrome users saw weird behavior with the interface for adding tags, in the beta "create entries" page -- the "x" button for removing a tag was replaced with a text label. This was because some browsers are now TOO MODERN for the javascript library we were using to test whether your browser was modern enough to show certain things. So we updated the version. // Heads up: We removed some code that seemed unused, but there's a small chance it could cause some visual breakage somewhere. So please let us know if you see something weird.


zorkian:
I'm going to lump this into "magical pixie dust" as far as tagging it in my brain goes.

nfagerlund:
Ahem, _pixel_ dust.

Issue #2817, opened 2020-08-03
break ljlib.pl into modules


martindemello:
From the comments in ljlib.pl:
```
    # ugly hack to shutup dependent libraries which sometimes want to bring in
    # ljlib.pl (via require, ick!).  so this lets them know if it's recursive.
    # we REALLY need to move the rest of this crap to .pm files.
```
Promoting it to a bug so it can be discussed.

martindemello:
I'm interested in working on this if someone more experienced can point me in the right general direction and answer the occasional question.

PR #2898, opened 2021-07-21
move some functions from ljlib into an LJ::Utils module

martindemello:
CODE TOUR: Started work on breaking ljlib.pl into modules.

Low-hanging fruit for #2817.

### DIFF
--- a/t/utils.t
+++ b/t/utils.t
@@ -2,19 +2,14 @@
 #
 # Test LJ::Utils module
 #
-# This code was forked from the LiveJournal project owned and operated
-# by Live Journal, Inc. The code has been modified and expanded by
-# Dreamwidth Studios, LLC. These files were originally licensed under
-# the terms of the license supplied by Live Journal, Inc, which can
-# currently be found at:
+# Authors:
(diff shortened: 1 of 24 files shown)


t/utils.t
```diff
@@ -0,0 +1,39 @@
+# t/utils.t
+#
+# Test LJ::Utils module
+#
+# This code was forked from the LiveJournal project owned and operated
```
zorkian, line 5: So since this is a net-new file, we should use the header that assigns copyright to Dreamwidth.

The LJ::Utils file I could go either way on... I don't know what we did with the LJ/User/FOOBAR.pm style modules, so I'd copy what we did there?

zorkian:
Hey this is awesome!

I made one comment about changing out the copyright header, but other than that, the code looks great and I'll merge once that's updated. (Or someone can press the merge button.)

t/utils.t
martindemello: done. the LJ/User/*.pm modules use the "forked from the livejournal project" headers so I kept that in Utils.pm too.

zorkian:
Perfect, thank you very much!

PR #2901, opened 2021-09-09
[WIP] Plack testing

zorkian:
DO NOT MERGE. Just posting for early visibility.

This makes the minimal changes necessary to enable supporting Plack as a
runtime environment instead of Apache.

The idea is to make running the code much, much simpler since we can
have a pure Perl web server and never need to deal with system modules
and the increasingly difficult to deal with Apache/mod_perl stuff.

However, Plack is a pluggable runtime so it still does support mod_perl
so we can do that for efficiency in production if we want. Or, we can
start to look at something like Starman.

CODE TOUR: Major, massive, sweeping backend changes just designed to make it easier to develop on Dreamwidth code, which, y'know, hopefully makes people want to work on it more. Because that'd be cool!

### DIFF
--- a/app.psgi
+++ b/app.psgi
@@ -1,0 +1,108 @@
+#!/usr/bin/perl
+#
+# app.psgi
+#
+# Dreamwidth entrypoint for Plack.
+#
+# Authors:
+#      Mark Smith <mark@dreamwidth.org>
+#
+# Copyright (c) 2021 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself. For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+use strict;
+use v5.10;
+use Log::Log4perl;
+my $log = Log::Log4perl->get_logger(__PACKAGE__);
+
+BEGIN { require "$ENV{LJHOME}/cgi-bin/ljlib.pl"; }
+
+use Plack::Builder;
+
+use DW::Request::Plack;
+use DW::Routing;
+
+# Initial configuration that happens on server startup goes here, these things don't
+# change from request to request, so only put things here that never change
+BEGIN {
+    # If we're a DEV server, do some configuration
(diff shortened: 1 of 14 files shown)


cgi-bin/ljlib.pl
```diff
@@ -129,6 +129,23 @@ BEGIN
 use DW::TaskQueue;
 use DW::BlobStore;
 
+# Load more modules so that we get as much advantage out of prefork
+# memory allocation as possible (as well as moving as much loading cost
+# to startup time)
+BEGIN {
+    # Do not run if we're in a test
+    unless ($LJ::_T_CONFIG) {
```
zorkian, line 137: These moved here because this is basically what mod_perl.pl was doing on startup and I needed to replicate it so that Plack would benefit from the preloading before startup/forking.

cgi-bin/modperl_subs.pl
```diff
@@ -74,9 +74,6 @@ BEGIN
 
 use MIME::Words;
 
-# Try to load DBI::Profile
-BEGIN { $LJ::HAVE_DBI_PROFILE = eval "use DBI::Profile (); 1;" }
```
zorkian, line 78: This module is never used anywhere.

app.psgi
```diff
@@ -0,0 +1,108 @@
...
+    my $r = DW::Request->get;
+
+    # Main request dispatch; this will determine what kind of request we're getting
+    # and then pass it to the appropriate handler. In the future, this should just
+    # be a call to DW::Routing and let it sort it out with all the controllers and
+    # such, but until then, we're having to dispatch between various generations
+    # of systems ourselves.
+    my $uri = $r->path;
+    $log->debug('Routing for URI: ', $uri);
+    if ( $uri =~ qr!^/api/v\d+/! ) {
```
zorkian, line 58: This only dispatches to API routes now, but this is where we need to add the rest of the dispatch engines... TT, BML, etc.

Although BML is probably a nightmare to pull off of Apache and maybe we should never support it here, which means Plack can't really be deployed until we finish the BML migration.

PR #2919, opened 2022-03-01
Cache API keys for faster lookup

momijizukamori:
CODE TOUR: A while ago, we added API keys! These have to be sent with every API request - right now, there's not a lot you can do with the API, so there isn't many of those, but in the future, this will hopefully increase. Right now, the server has to check the database every time to see if an API key is valid. Checking the database is expensive in terms of time and resources, so we ideally want to reduce the number of times that happens. This change makes the server cache successful key lookups, so if you make a bunch of API requests in a row, it will 'remember' the key after the first one and not have to check the database.

### DIFF
--- a/cgi-bin/DW/API/Key.pm
+++ b/cgi-bin/DW/API/Key.pm
@@ -57,12 +57,17 @@ sub new_for_user {
 sub get_key {
     my ( $class, $hash ) = @_;
     return undef unless $hash;
+    my $memkey = [ $hash, "api_key:" . $hash ];
 
-    my $dbh     = LJ::get_db_writer() or croak "Failed to get database";
-    my $keydata = $dbh->selectrow_hashref(
-        "SELECT keyid, userid, hash FROM api_key WHERE hash = ? AND state = 'A'",
-        undef, $hash );
-    carp $dbh->errstr if $dbh->err;
+    my $keydata = LJ::MemCache::get($memkey);
+    unless ( defined $keydata ) {
+        my $dbh = LJ::get_db_writer() or croak "Failed to get database";
+        $keydata = $dbh->selectrow_hashref(
+            "SELECT keyid, userid, hash FROM api_key WHERE hash = ? AND state = 'A'",
+            undef, $hash );
+        LJ::MemCache::set( $memkey, $keydata, 60 * 60 * 24 ); # cache for one day
+        carp $dbh->errstr if $dbh->err;
+    }
 
     if ($keydata) {
         my $user = LJ::want_user( $keydata->{userid} );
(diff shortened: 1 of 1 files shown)


momijizukamori:
Two quick notes:
1) I'm flexible on the cache length, it just seemed like a bad idea to have it be 'forever'
2) I think the storing of the arrayref for the row data should be okay? I checked it manually in cache via telnet, and also did a debug line to see if was doing db lookup or not, and I'm pretty sure it's writing and reading the data from memcache okay, but I would feel more comfortable if someone with more memcache experience had a quick look.

cgi-bin/DW/API/Key.pm
```diff
@@ -57,12 +57,17 @@ sub new_for_user {
...
+    my $memkey = [ $hash, "api_key:" . $hash ];
+
+    my $keydata = LJ::MemCache::get($memkey);
+    unless ( defined $keydata ) {
+        my $dbh = LJ::get_db_writer() or croak "Failed to get database";
+        $keydata = $dbh->selectrow_hashref(
+            "SELECT keyid, userid, hash FROM api_key WHERE hash = ? AND state = 'A'",
+            undef, $hash );
+        LJ::MemCache::set( $memkey, $keydata, 60 * 60 * 24 );    # cache for one day
+        carp $dbh->errstr if $dbh->err;
```
zorkian, line 69: I would move this above the memcache set, so that we don't put invalid data into memcache.

cgi-bin/DW/API/Key.pm
```diff
@@ -57,12 +57,17 @@ sub new_for_user {
 sub get_key {
     my ( $class, $hash ) = @_;
     return undef unless $hash;
-
-    my $dbh     = LJ::get_db_writer() or croak "Failed to get database";
-    my $keydata = $dbh->selectrow_hashref(
-        "SELECT keyid, userid, hash FROM api_key WHERE hash = ? AND state = 'A'",
-        undef, $hash );
-    carp $dbh->errstr if $dbh->err;
+    my $memkey = [ $hash, "api_key:" . $hash ];
```
zorkian, line 60: I would recommend just using

```perl
my $memkey = "api_key:$hash";
```

The reason you would use the `[ X, Y ]` format is only if you are going to be getting/setting multiple related Y's and you want them to map to the same underlying Memcache server.

For example, we frequently use something like `[ $userid, "thingy:$userid" ]` because that lets us put all of that user's data onto the same Memcache server. This makes it more efficient to do a multi-get or multi-set, also means that if you have a memcache outage, it affects 100% of data for X% of users rather than X% of all data for 100% of users.

zorkian (approved):
Approved with two comments. If you address those, please feel free to self-merge!

PR #2930, opened 2022-03-16
Clear "request" cache of db handles when an import is ending

alierak:
… before doing anything else that uses a db, in case a db connection has gone away.

CODE TOUR: Long-running imports should no longer fail with a "MySQL server has gone away" error

### DIFF
--- a/cgi-bin/DW/Worker/ContentImporter.pm
+++ b/cgi-bin/DW/Worker/ContentImporter.pm
@@ -158,6 +158,10 @@ sub fail {
     my ( $class, $imp, $item, $job, $msgt, @args ) = @_;
     $0 = 'content-importer [bored]';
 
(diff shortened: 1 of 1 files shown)


alierak:
This needs more work -- so far only addresses the end-of-import handlers (temp_fail, fail, and ok). An import can also end in "MySQL server has gone away" if it does 10 minutes worth of work that doesn't use a particular db handle and then suddenly it does (e.g., looks through 75000 comments that already exist and don't need updating, and then runs into one that does, at which point insert_comment calls LJ::get_prop, etc.  -- might be cleaner just to add a DBI::Role option to not use the "request" cache at all, and have the importer set that for entries / comments imports.

alierak:
Okay, not perfect, but it should cover the cases we've actually seen where the importer spends a long time fetching comments or whatever and then tries to use an idle db handle.

PR #2949, opened 2022-04-27
convert /manage/banusers to TT

momijizukamori:
CODE TOUR: Part of the ongoing crusade eliminate BML from the codebase! One page down, many more to go...


zorkian:
Woo! Thank you!

Every time I review one of these, I have these itches to rewrite the Perl so it's more modern and cleaner... but then I realize that would make this project take 10x as long. Soooo nope. Discretion is the better part of valor or something.

rahaeli:
Only one endless modernization project that everybody hates at a time, Mark.

Issue #2958, opened 2022-06-18
<details>/<summary> tags have no styling in sitescheme


analpi:
Using the < details > tag lets you hide text, which can then be expanded if you click on the block. On journal themes these blocks get preceded by little sideways triangles to let the reader know they're there, but on sitescheme pages they just look like plaintext. Can you add the little triangles there too?

momijizukamori:
So the little triangle is actually default browser styling, and something in one of our CSS files is stripping that out by changing the display type for those tags. Which is very silly in this case, and once I'm sure it won't make other bits of the site layout go haywire I'm gonna see about fixing that.

PR #2964, opened 2022-08-02
Don't override browser display for details and summary

momijizukamori:
CODE TOUR: The `details` HTML element is basically a nice little native expand/collapse that allows you to hide and unhide blocks of text. By default, browsers usually style it with a little arrow at the side to indicate that it can be expanded and collapsed. We, however, had some CSS fixes for old browsers (IE and very old versions of Firefox) that didn't support this newer element - which were probably added when those browsers weren't super old and it was actually a good idea - that caused the arrow to vanish entirely for all browsers. Very silly. Now that we are in the future, those workarounds can be removed!

### DIFF
--- a/htdocs/scss/foundation/normalize.scss
+++ b/htdocs/scss/foundation/normalize.scss
@@ -32,7 +32,6 @@ body {
 
 article,
 aside,
-details,
 figcaption,
 figure,
 footer,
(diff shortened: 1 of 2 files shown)


momijizukamori:
This closes #2958

PR #2970, opened 2022-09-08
Add Beta flag for new inbox redesign

momijizukamori:
CODE TOUR: After we accidentally pushed the inbox update live without planning to, and got a bunch of feedback, we realized we uhhhh should probably make it a beta, until we get all the bugs ironed out, as it's a pretty major rewrite of the code.

NB: This needs a new block added to `%BETA_FEATURES` in config-local.pl:
```
       "inbox" => {
            start_time => 0,
            end_time => "Inf",
        },
```

Also, when we finally remove the old inbox page, we will need to revert a few of the changes here (the `/new` bits, which are to keep paths from conflicting with the old inbox)

### DIFF
--- a/cgi-bin/DW/Controller/Inbox.pm
+++ b/cgi-bin/DW/Controller/Inbox.pm
@@ -27,9 +27,9 @@ use DW::FormErrors;
 use LJ::Hooks;
 use Data::Dumper;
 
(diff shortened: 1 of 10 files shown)


momijizukamori:
And for ref, here's the original MR for the new inbox: https://github.com/dreamwidth/dw-free/pull/2853

zorkian (approved):
LGTM

PR #2976, opened 2022-09-21
Add twitch to embed whitelist

### LINKED ISSUE #2638: add Twitch to embed whitelist
https://www.dreamwidth.org/support/see_request?id=41598

Sample code:

`<iframe src="https://player.twitch.tv/?autoplay=false&video=v582773417" frameborder="0" allowfullscreen="true" scrolling="no" height="378" width="620"></iframe><a href="https://www.twitch.tv/videos/582773417?tt_content=text_link&tt_medium=vod_embed" style="padding:2px 0px 4px; display:block; width:345px; font-weight:normal; font-size:10px; text-decoration:underline;">Watch #DIGITALDRAGSHOW FRIDAYS @ 7PM PST/10PM EST from biqtchpuddin on www.twitch.tv</a>`

chebegeek:
CODE TOUR: Added Twitch to the embed whitelist, and added tests to match

The addition of Twitch to the embed whitelist is not straight-forward;

* I couldn't find a way to match (past) the ? in twitch urls, so the regex isn't as precise as I'd like. If anyone has suggestions I'd love to hear them,

* Twitch embeds require https connections, so I am getting 'refused connections' on my dreamhack,

* Twitch embeds require &parent=website.host appended to the src in the iframe. Is this something we want to leave up to users to get right?

Attempts to fix #2638.


kareila:
Status update: tested this and the embed renders, but it doesn't actually work in dev because twitch requires a `parent` attribute containing the domain of the site doing the embedding. I tested it with the embed subdomain configured on my dev box with https enabled and still got an error, so I'm not sure what exactly is still going wrong.

chebegeek:
The error I'm getting in the browser console is;
> "Refused to frame 'https://player.twitch.tv/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors"

I have scoured the internet. Because we place the iframe inside an iframe inside the page, we need to include all with &parent= attributes on src. I have tried with embed.system.user.hack.dreamwidth.net and every combination of (www.)(system.)user.hack.dreamwidth.net I could think of. Some people report needing to add things like localhost as well? Other than that an SSL connection is required, which I can't get around. Haven't found anything else, yet.

rahaeli:
You get the Tedious Research Star this week, Chebe!

PR #2984, opened 2022-10-14
redirect hostnames with underscores before they reach apache

alierak:
CODE TOUR: Re-enables username.dreamwidth.org URLs for usernames containing underscores, after an Apache upgrade broke them


zorkian (approved):
Seems valid. Will test on canary before releasing to the wild, though.

PR #2986, opened 2022-10-17
add script dump-profile-editors.pl and associated hook

kareila:
CODE TOUR: This is an admin utility script for retrieving for review a list of users who have recently edited their profile. 

We were previously doing this by parsing server log files, but this provides an alternate logging implementation using memcache to store the list. A hook appends to the list whenever a save happens, and the utility script resets the list when it is run.

### DIFF
--- a/cgi-bin/DW/Hooks/ProfileSave.pm
+++ b/cgi-bin/DW/Hooks/ProfileSave.pm
@@ -1,0 +1,52 @@
+#!/usr/bin/perl
+#
+# DW::Hooks::ProfileSave
+#
+# This module implements a hook for lightweight logging of uids
+# who save profile edits, for later examination to detect
+# accounts being used for spam purposes.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
+#
+# Copyright (c) 2022 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself. For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Hooks::ProfileSave;
+
+use strict;
+use warnings;
+
+use LJ::Hooks;
+use LJ::MemCache;
+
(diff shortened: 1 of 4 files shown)


cgi-bin/DW/Hooks/ProfileSave.pm
```diff
@@ -0,0 +1,52 @@
...
+        my $oldurl = $saved->{url} // '';
+        my $newurl = $post->{url}  // '';
+
+        $log_edit = 0 if $oldurl eq $newurl;
+        $log_edit = 0 unless $newurl;
+
+        return unless $log_edit;
+
+        # set the new key - expires after a week if no activity
+        my $memval = LJ::MemCache::get('profile_editors') // [];
```
zorkian, line 46: It's probably OK at our size, but this is technically racy (since two profile edits will collide, and someone's edit won't end up in the memcache key.)

If you don't want it to be racy you can use ddlockd to get a lock (with some exponential backoff and maximum attempts) and only edit memcache while you have the lock.

cgi-bin/DW/Hooks/ProfileSave.pm
```diff
@@ -0,0 +1,52 @@
...
+        push @$memval, $u->id;
+        LJ::MemCache::set( 'profile_editors', $memval, 86400 * 7 );
```
zorkian, line 48: Memcache has a value size limit, I think? Around 1MB?

Assuming userids are 7 digits that's about 100,000 userids (if they were strings) so I guess this probably won't ever matter in this use case for our size... :)

cgi-bin/DW/Hooks/ProfileSave.pm
kareila: Yeah, we discussed that and decided it wasn't worth the extra complication in this case.

cgi-bin/DW/Hooks/ProfileSave.pm
kareila: When we were doing this by checking server logs we never saw more than 1000 in a week.

PR #2998, opened 2022-10-24
Enable full screen video in embeds

### LINKED ISSUE #2995: investigate allowing fullscreen option in videos embedded from YouTube
https://www.dreamwidth.org/support/see_request?id=46108
https://www.dreamwidth.org/support/see_request?id=46558

chebegeek:
CODE TOUR: iframes insides iframes, all must be in agreement to enable full screen video for embeds.

* Adds allowfullscreen="true" to the outer embed iframe. allowfullscreen must be set on both iframes (the user-created embed as well as the container iframe) to enable full screen video for embeds. If the user does not set it on each embedded video it will remain disabled.
* Adds test to ensure allowfullscreen="true" on the outer iframe.

With special thanks to alierak.

Fixes #2995

### DIFF
--- a/cgi-bin/LJ/EmbedModule.pm
+++ b/cgi-bin/LJ/EmbedModule.pm
@@ -233,7 +233,7 @@ sub parse_module_embed {
                     push @stack, $tag;
                 }
 
-                # append the tag contents to new embed buffer, so we can convert in to lj-embed later
+               # append the tag contents to new embed buffer, so we can convert in to lj-embed later
                 $embed .= $reconstructed;
             }
             else {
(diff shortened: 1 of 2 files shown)


alierak (commented):
This looks like it'll always add allowfullscreen to the outer iframe. Do we necessarily want to do that for every possible embed, or only those from trusted video sites, or only those from trusted video sites where the user has also set it on the inner iframe, or something else?

chebegeek:
The inner iframe goes through all the trusted site checks, if they fail the outer iframe will be empty and have nothing to display. I don't see any risk with this, but if we prefer something more selective I can do that

kareila:
At a glance, it seems okay to me too - the existing code already treats the `allowtransparency` attribute this way, so following suit with `allowfullscreen` seems logical. It was mentioned above that if the inner iframe didn't also set it, nothing would happen.

I do see some spurious whitespace edits in this diff - please make sure you retidy before submitting changes! Otherwise the `t/02-tidy.t` test will fail.

chebegeek:
Oh no, I thought I was helping tidy up more, I'll revert them

PR #3007, opened 2022-11-09
shop til you drop

### LINKED ISSUE #2328: currency formatting glitch when paying by check/MO in shop
https://www.dreamwidth.org/support/see_request?id=38431

After submitting an order, "Amount Due" is listed as "$XX.X USD" with one place after the decimal point instead of two.

### LINKED ISSUE #2974: minimum amount for mailed payments
Our remailer has raised their monthly rates *and* raised their reshipping rates over and above the post office rate raises. It's always cost more to accept check/money order but this latest rate raise has gotten it to the point where if we don't put a minimum amount for C/MO payments, it will cost us more to accept them than the amount of the payment. 

In a last ditch effort to avoid having to get rid of C/MO entirely as an option, a minimum floor of $20 for C/MO payments gets us to "still more expensive to accept but at least not actively losing money when combining cost of accepting payment with cost to support the paid features for the length of paid time it buys".

kareila:
CODE TOUR: Lots of configuration tweaks and minor fixes for the shop, most notably making it possible to enforce a minimum purchase amount when paying by check or money order.

More details in the individual commit messages.

Fixes #2974. Fixes #2328.

### DIFF
--- a/cgi-bin/LJ/Widget/ShopCart.pm
+++ b/cgi-bin/LJ/Widget/ShopCart.pm
@@ -191,8 +191,26 @@ sub render_body {
             $ret .= " &nbsp;&nbsp;";
 
             # check or money order button
+            my $cmo_threshold =
+                $LJ::SHOP_CMO_MINIMUM ? $cart->total_cash - $LJ::SHOP_CMO_MINIMUM : undef;
+            my $disable_cmo = defined $cmo_threshold ? $cmo_threshold < 0 : 0;
             $ret .= $class->html_submit(
-                checkout_cmo => $class->ml('widget.shopcart.paymentmethod.checkmoneyorder') );
+                checkout_cmo => $class->ml('widget.shopcart.paymentmethod.checkmoneyorder'),
+                { disabled => $disable_cmo }
+            );
+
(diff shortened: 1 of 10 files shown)


cgi-bin/LJ/Widget/ShopCart.pm
```diff
@@ -191,8 +191,29 @@ sub render_body {
             $ret .= " &nbsp;&nbsp;";
 
             # check or money order button
-            $ret .= $class->html_submit(
-                checkout_cmo => $class->ml('widget.shopcart.paymentmethod.checkmoneyorder') );
+            my $cmo_threshold =
+                $LJ::SHOP_CMO_MINIMUM ? $cart->total_cash - $LJ::SHOP_CMO_MINIMUM : undef;
+            my $disable_cmo = defined $cmo_threshold ? $cmo_threshold < 0 : 0;
```
zorkian, line 196: nit: could probably be simpler, altho I'm going to merge as-is.

```perl
my $disable_cmo = $cart->total_cash < ( $LJ::SHOP_CMO_MINIMUM // 0 ) ? 1 : 0;
```

PR #3014, opened 2022-11-11
Rename hangouts to chat

### LINKED ISSUE #3002: Google Hangouts is discontinued. Should we remove it from "Other Sites" on the profile?
Or is there some other doomed Google service we can rename it to that uses the same user info?

zorkian:
CODE TOUR: Rename Google Hangouts to Google Chat. Fixes #3002.


kareila:
This may need a translation string update in production since the strings weren't re-keyed.

PR #3018, opened 2022-11-12
Fix to order by jitemid, not journalid

### LINKED ISSUE #2883: paging through entries filtered by tag losing some entries and repeating others
Detailed report here: https://www.dreamwidth.org/support/see_request?id=43162

Summary: user keeps changing the "Skip N" value for their journal, and whichever amount is skipped, the Nth entry will be repeated at the top of the next page, and the N+1-th entry will not be shown.

zorkian:
CODE TOUR: This query was meant to use jitemid (judging from the commit message from 12 years ago which says `jitemid`), but looks like a typo. Fix it. This should address situations where entries with the same timestamp get ordered weirdly, leading to some odd behaviors on page boundaries. Fixes #2883.


zorkian:
On further thinking, I think this is wrong, please don't merge. I want to write some test for this behavior.

rahaeli:
We made a change to the page boundaries logic in open beta that wound up producing *really* weird and irregular results with some entries getting eaten entirely, so we backed it out again, which I mention only in case it was relevant because my memory coughed up the overview but not the exact details of the change

PR #3033, opened 2022-11-20
API tooling

momijizukamori:
CODE TOUR: A whole bunch of updates to the tools to build REST API routes, including better validation, reusable components, and a draft documents page!


momijizukamori:
Comments from rah on the docs page text:

> This API, and it's documentation

Should be its
Other than that, looks good to me! Although if you're fixing the its, I would also take this out of passive voice

> This document attempts to show what information can be requested

To "what information you can request"

zorkian:
GOLD STAR (also I need this since I've been cutting and pasting on top of it locally and I realized it'd be easier if I merged it, I'll fix up the text in a followup)

PR #3039, opened 2022-11-22
Update comments to match license stated, add link

alansill:
Although the license stated in the text is CC-by-SA-2.5, the comments in the header stated inconsistent information apparently left over from use of that header in a code file. This pull request updates the comments to be consistent with the license stated, and adds a link to the license as required by the license terms for its use. I would like to fix these issues before making a copy to use on our own site with appropriate alterations and credit as per the license.

CODE TOUR: no-impact


rahaeli:
We used to have a free and a nonfree branch and that was the header for the nonfree, yes! Technically we should include a request for people to modify 'em for their own details, but this is a good start, thank you

Issue #3068, opened 2023-02-13
fix syntax error in etc/config-local.pl.example


kareila:
@vivtek pointed out that our config-local is missing a close brace at the end of the file. I looked into it and confirmed it was accidentally deleted here: 9a7b1e697241b8cb56f692e02df883dd497c249e

While we're fixing syntax, we should also add a trailing semicolon to the IS_CANARY line to prevent the next variable that comes along from causing the parser to choke as well. (And maybe change the comment to read "every page" instead of "ever page"?)

Vivtek:
OK, I'm claiming this, since I found it. It'll be a good practice pull request.

Incidentally, that lack of a semicolon isn't a problem because it's the last thing before the closing bracket (if the closing bracket is still there). I think it's bad practice to leave it off, precisely because of situations like this, but it's not actually a syntax error.

kareila:
Yes, that's why I said for "the next variable that comes along" to not cause an error. Let us know if you run into any more trouble!

Vivtek:
Oh, duh, sorry, my reading comprehension isn't revved up yet this morning.

I guess I was just focused on that semicolon because that's what I saw first, put a semicolon on it, and only after restarting did I notice the lack of the closing bracket.

kareila:
Fixed in #3069.

PR #3069, opened 2023-02-13
Restored close-bracket in config-local.pl.example (#3068)

Vivtek:
CODE TOUR: fix config-local.pl.example so new Dreamhacks start correctly


kareila:
For future reference, it's best to include some variation on (in this case) `fixes #3068` in your PR description for the associated issue to close automatically when the PR is merged. I manually closed it, though, so we're good. Thanks!

Issue #2955, opened 2022-06-13
change DW's Google Analytics snippet to use GA4


rahaeli:
Not the user-facing GA feature (that's in #2954) but the DW production GA code on site-skinned pages needs to be updated. The code that needs to be swapped in for the old GA code is:

    
    <script async src="https://www.googletagmanager.com/gtag/js?id=G-4FH9Y1TDGT"></script>
    <script>
      window.dataLayer = window.dataLayer || [];
      function gtag(){dataLayer.push(arguments);}
      gtag('js', new Date());
    
      gtag('config', 'G-4FH9Y1TDGT');
    </script>

That's got the DW identifier in it already (G-4FH9Y1TDGT). I've already done the thingamajig with the Google account itself so we should be able to start getting data if we just update the code.

momijizukamori:
For ref in my digging - for site pages, we store the identifier value in the `SITE_PAGESTAT_CONFIG{google_analytics }` config value in `config-private.pl`.

This is then used in https://github.com/dreamwidth/dw-free/blob/950b1e9777e6f814fa518e74831943c3f98d7df6/cgi-bin/DW/PageStats/GoogleAnalytics.pm which actually has the code for building the rest of the script tag.

PR #3095, opened 2023-05-29
Add new GA4 google analytics script into pages

### LINKED ISSUE #2954: change old Google Analytics paid feature to GA4 since UA is being sunsetted 30 July 2023
Users will need to upgrade their Google account to create a new GA4 property and get the new tag ID, so it's probably best to make a new setting/feature for the new GA4 version, since the two user identifiers aren't interchangeable and users will have to take action with Google to set up the new system. The new GA4 JS (using DW's ID, G-4FH9Y1TDGT) is:

    
    <script async src="https://www.googletagmanager.com/gtag/js?id=G-4FH9Y1TDGT"></script>
    <script>
      window.dataLayer = window.dataLayer || [];
      function gtag(){dataLayer.push(arguments);}
      gtag('js', new Date());
    
      gtag('config', 'G-4FH9Y1TDGT');
    </script>

The G-4FH9Y1TDGT is the part that will need to plug in the "measurement ID" that the user is able to provide to us.

momijizukamori:
CODE TOUR: This is the second half of the code to update Google Analytics scripts to the newer Google format (the first half is #3094 ). This bit constructs the script tags using either the new user setting (for journals where users have enabled google analytics) or the site config variable (for site pages).

NOTE FOR DEPLOYERS (@zorkian or @alierak ): the private-config.pl file needs to be updated with the new site config key.

`    %SITE_PAGESTAT_CONFIG = (
        google_analytics => 'UA-xxxxxx-x',
        ga4_analytics => 'G-4FH9Y1TDGT'
    );`

Closes #2954 and #2955

### DIFF
--- a/cgi-bin/DW/PageStats/GoogleAnalytics4.pm
+++ b/cgi-bin/DW/PageStats/GoogleAnalytics4.pm
@@ -1,0 +1,68 @@
+#!/usr/bin/perl
+#
+# DW::PageStats::GoogleAnalytics4
+#
+# LJ::PageStats module for Google Analytics
+#
+# Authors:
+#      Andrea Nall <anall@andreanall.com>
+#
+# Copyright (c) 2009 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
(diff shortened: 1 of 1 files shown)


kareila:
I have updated config-private.pl in the appropriate places.

PR #3098, opened 2023-06-05
Clean up Perl -> JS dumping

momijizukamori:
CODE TOUR: This started with a support request and a mystery. The request was opened by a user who had renamed their journal to a username that was all numbers, starting with `0`, and then found they could no longer delete comments from their journal - they got a strange error when they tried. On investigation, this turned out be an issue ONLY with number-only usernames, and ONLY if they started with `0` - and only for comments in their own journal (they could comment in someone else's journal and those comments would behave normally). The reason for this turned out to be the unholy combination of 1) some really old code inherited from LJ and 2) the meeting of two different weakly-typed programming languages (Perl and Javascript).

In 2005, Brad wrote some Perl code to print Perl variables in a format Javascript understands. The code was a little janky, but in his defense this predates broader Perl support for doing this very thing. Unfortunately Perl doesn't really differentiate much between strings (ie, text) and numbers most of the time, and this code turned anything that was all numbers into a number. In printing out information about comments for JS, all-number usernames would end up as `{"journal": 007}` instead of `{"journal": "007"}`. This is exciting because, [per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_literals), Javascript interprets numbers starting with `0` as octal number (if all digits that follow are under 8) or decimal number (in which case it strips the leading `0`s). This means that a journal named `042` would be converted to the octal number `34`, and then a check later on to see if we were acting in the correct journal to manage a comment would check if `34` was the same as `"042"` and fail, blocking the action on the comment. If it got converted to a decimal number, this check would succeed (because unless you're using strict equality, `7` and `"007"` are actually equal in JS), but the JS would then ask the server to act on a comment in journal `7` and the server would look and see that the comment was in journal `007` and go "Can't let you do that, buddy".

(without the starting `0`, swapping back and forth between number and string in Javascript repeatedly produces the same thing, which is why this didn't affect *all* journals with numerical usernames)

Interestingly, Livejournal actually [switched from using this code to using a more standard conversion function](https://github.com/apparentlymart/livejournal/commit/8e20f0a23a423421c79d9a5f9a830fa70c2dd4a3), though after the point at which the Dreamwidth codebase diverges from it. This fix does the same thing, though I then had to add some fixes in other places to keep numbers that should be printed as numbers and not strings in the correct form.

* Replace all instances of `LJ::js_dumper` with `to_json`
* Force `LJ::is_enabled` to return `1` or `0` so that printing to JSON receives the correct format.
* Update the handling of site vars in JS to use more modern JS that makes it clearer what the function is actually doing.

### DIFF
--- a/cgi-bin/LJ/Web.pm
+++ b/cgi-bin/LJ/Web.pm
@@ -32,6 +32,7 @@ use LJ::Subscription::Pending;
 use LJ::Directory::Search;
 use LJ::Directory::Constraint;
 use LJ::PageStats;
+use LJ::JSON;
 
 # <LJFUNC>
 # name: LJ::img
@@ -1488,7 +1489,7 @@ RTE
             . LJ::ejs( BML::ml('fcklang.userprompt.user') ) . "\";\n";
         $out .= "FCKLang.UserPrompt_Site = \""
             . LJ::ejs( BML::ml('fcklang.userprompt.site') ) . "\";\n";
-        $out .= "FCKLang.UserPrompt_SiteList =" . LJ::js_dumper( \@sitevalues ) . ";\n";
+        $out .= "FCKLang.UserPrompt_SiteList =" . to_json( \@sitevalues ) . ";\n";
         $out .= "FCKLang.InvalidChars = \"" . LJ::ejs( BML::ml('fcklang.invalidchars') ) . "\";\n";
         $out .= "FCKLang.LJUser = \"" . LJ::ejs( BML::ml('fcklang.ljuser') ) . "\";\n";
         $out .= "FCKLang.LJVideo = \"" . LJ::ejs( BML::ml('fcklang.ljvideo2') ) . "\";\n";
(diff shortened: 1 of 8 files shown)


cgi-bin/LJ/Web.pm
```diff
@@ -2910,8 +2879,8 @@ sub res_includes {
             cmax_comment        => LJ::CMAX_COMMENT,
         );
 
-        my $site_params     = LJ::js_dumper( \%site );
-        my $site_param_keys = LJ::js_dumper( [ keys %site ] );
+        my $site_params     = to_json( \%site );
+        my $site_param_keys = to_json( [ keys %site ] );
```
alierak, line 2883: Looks like site_param_keys is no longer needed due to another change below that removed the one place it was used.

momijizukamori: Oh good catch!

PR #3099, opened 2023-06-10
build-static fixes

alierak:
CODE TOUR: The development script that builds static CSS and JS files could get confused with different developers' data on the dreamhack server, and was doing extra work most of the time.

- use a unique temp file and clear / remove it between runs


zorkian:
tfw multi-user environments. thanks!

PR #3100, opened 2023-06-11
add admin view to /file/list

kareila:
CODE TOUR: Denise mentioned needing a way for site admins to review uploaded images for possible TOS violations. This adds that capability to the /file/list page for users who have the canview:images admin priv.

### DIFF
--- a/cgi-bin/DW/Media/Base.pm
+++ b/cgi-bin/DW/Media/Base.pm
@@ -136,7 +136,11 @@ sub visible_to {
     # at this point, if we don't have a remote user, fail
     return 0 unless LJ::isu($other_u);
 
-    # private check.  if it's us, allow, else fail.
+    # private check.  if it's us or an admin, allow, else fail.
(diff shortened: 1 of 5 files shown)


cgi-bin/DW/Media/Base.pm
```diff
@@ -136,7 +136,11 @@ sub visible_to {
     # at this point, if we don't have a remote user, fail
     return 0 unless LJ::isu($other_u);
 
-    # private check.  if it's us, allow, else fail.
+    # private check.  if it's us or an admin, allow, else fail.
+    my $refer       = { DW::Request->get->headers_in }->{Referer};
```
zorkian, line 140: What is this syntax `{ foo }` what does that do? (Either TIL or I've forgotten.)

zorkian:
Seems reasonable to me, auth part looks solid. I'm trusting the syntax question I had is just a lack of knowledge on my part and you've tested this, so merging.

Issue #3067, opened 2023-01-31
treat user bio of non-validated accounts like anonymous comments for HTML cleaning purposes 


rahaeli:
Will be superseded by issue #3051 and can be closed if 3051 is implemented before this one, but I figured this might be a faster fix: until we can implement 3051, we should make the HTML cleaner treat the user-supplied bio the way it treats anonymous comments (no auto-linking, links replaced by bare unlinkified URLs, etc) until and unless the user confirms their email address.

PR #3104, opened 2023-06-12
Strip links from the userbio for accounts that aren't validated yet

momijizukamori:
CODE TOUR: Part of the endless war on spam - this delinkifies links in the bios of users who haven't validated their email addresses yet, making it slower for spammers to create a bunch of profiles and dump spam links in them (if they aren't actual links, it doesn't count towards page rank algorithms on Google etc)


momijizukamori:
This closes #3067

PR #3105, opened 2023-06-13
Add spam domain auto-suspend list

### LINKED ISSUE #3080: URL blocklist for HTML cleaner
I give up. I've resisted doing this for 15 years, because if you keep a list of "domains you will be suspended for posting links to" and anyone finds out any of the domains on the list they will make it a fun and exciting game to get other people to post the links that will get their account instant-suspended, but we are well past the point where cleaning up after accidental misfires will be signficantly, signifcantly less work than continuing to suspend the dozens of accounts daily that are spamming for the same domains over and over again.

The system must:

* attach "can edit the list" to a new priv siteadmin:spamblocklist 
* create a new admin system whereby someone with siteadmin:spamblocklist can add a domain to the blocklist or edit/remove previous items from the blocklist
* do wildcard matching so you can specify "spamdomain.com" on the list and it will match any URL at that domain
* add a hook that will scan entries, comments, profile bios, profile URLs in the "website" field of the profile and icon keywords (a depressing number of spam accounts add spam links in the icon keywords)
* if any of the scanned items contain any string that matches a domain on the blocklist, suspend the account immediately
* log to statushistory with a suspend message of "auto-suspend for matching domain blocklist: spamdomain.com"
* have the 'admin' field of the statushistory logger be some unique string that we can search for on /admin/statushistory like 'blocklistmatch' or something 

Not necessary but would be nice: if the suspend message could include where the blocklist match was ('entry: 11434.html', 'profile bio', 'icon keyword', etc). It should check for the bare string, not just in a &lt;a href> tag, because a lot of these spambots use Markdown, BBCode, etc, etc, even when we don't support the text markup format, or just use a bare URL that autolinks.

momijizukamori:
CODE TOUR: A number of our spammers are repeat customers. This change allows site admins to set up a list of domains that posting links to will cause your account to get auto-suspended, with a note in the admin log of what domain got flagged and where it got flagged.

This should check all field of entries, all fields of userbios, all fields of comments, and all fields of icons. Suspends will be from the `system` user. Fixes #3080

### DIFF
--- a/cgi-bin/DW/Hooks/SpamCheck.pm
+++ b/cgi-bin/DW/Hooks/SpamCheck.pm
@@ -30,7 +30,7 @@ LJ::Hooks::register_hook(
         my ( $user, $data, $location ) = @_;
         return unless defined $user && defined $data && defined $location;
         my $system = LJ::load_user('system');
-        my $blocked_links = [ 'google.com', 'yahoo.com' ];
+        my @blocked_links = grep { $_ } split( /\r?\n/, LJ::load_include('spamblocklist') );
         my $suspended     = 0;
         my $location_str  = $location; # same for everything but hashrefs
 
@@ -38,7 +38,7 @@ LJ::Hooks::register_hook(
             my $item = shift;
             return unless defined $item;    # don't waste time iterating over undefined items
 
-            foreach my $re (@$blocked_links) {
+            foreach my $re (@blocked_links) {
                 if ( $item =~ $re ) {
                     LJ::User::set_suspended( $user, $system,
                         "auto-suspend for matching domain blocklist: $re in $location_str" );
--- a/cgi-bin/DW/Hooks/SpamCheck.pm
+++ b/cgi-bin/DW/Hooks/SpamCheck.pm
@@ -29,6 +29,7 @@ LJ::Hooks::register_hook(
     sub {
         my ( $user, $data, $location ) = @_;
         return unless defined $user && defined $data && defined $location;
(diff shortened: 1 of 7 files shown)


cgi-bin/DW/Hooks/SpamCheck.pm
```diff
@@ -0,0 +1,68 @@
...
+use Scalar::Util qw/reftype/;
+
+use LJ::Hooks;
+
+LJ::Hooks::register_hook(
+    'spam_check',
+    sub {
+        my ( $u, $data, $location ) = @_;
+        return unless defined $u && defined $data && defined $location;
+        return if $u->has_priv('siteadmin');    # some users can be trusted
```
zorkian, line 31: I definitely misread this as "can't be trusted" and was like "you know, she's right."

cgi-bin/DW/Hooks/SpamCheck.pm
```diff
@@ -0,0 +1,68 @@
...
+                    $u->set_suspended( $system,
+                        "auto-suspend for matching domain blocklist: $domain in $loc" );
+                    return 1;
+                }
+            }
+        };
+
+        if ( reftype $data eq reftype [] ) {
+            foreach my $item (@$data) {
+                my $suspended = $check_item->( $item, $location );
```
zorkian, line 51: nit: could just be `last if $check_item->( ... );` right?

zorkian:
Seems like a pretty big hammer but, I suppose, that was the ask. Hammer has been built.

PR #3108, opened 2023-06-14
Add call to print_meta_tags() for venture layout

cmho:
CODE TOUR: This adds the function that includes useful meta tags to the Venture layout, which allows it to, among other things, display the layout responsively on mobile devices.


rahaeli:
HOORAY FIRST COMMIT!!! Thank you for fixing this!

Issue #3088, opened 2023-04-20
add userhead support for ko-fi.com


kareila:
Request here: https://www.dreamwidth.org/support/see_request?id=47323

Issue #3063, opened 2023-01-21
update userhead icon for fanfiction.net users


kareila:
https://www.dreamwidth.org/support/see_request?id=42983

The current image URL, `https://www.fanfiction.net/static/images/favicon_2010_site.png`, has disallowed hotlinking, but according to one of our support volunteers, the correct link should actually be https://www.fanfiction.net/static/icons3/ff-icon-180.png (although that's too large to use as a userhead...)

PR #3119, opened 2023-06-17
Add kofi userheads; update FFN userheads

cmho:
CODE TOUR: Adds Ko-Fi userhead support and updates the userhead for Fanfiction.net.

Addresses #3088 and #3063


zorkian:
Merging. One note, it looks like ko-fi supports HTTPS URLs, so it would be nice to use `https://` so we don't leak what accounts someone is clicking out to -- but that's pretty minor, so I'm merging. :)

Issue #3110, opened 2023-06-14
Add Media Query support to all base layouts


momijizukamori:
The core2base layouts have functionality to expose a media query setting in the wizard, which allows users to easily set media breakpoints for collapsing the layout tidily on smaller screens. It would be great to have this support in all our base layouts.

Relevant bits of code:
```
    property use medium_breakpoint_width;
    property use large_breakpoint_width;
```
which goes in `propgroup presentation` and then
```
    var string small_media_query = generate_small_media_query();
    var string medium_media_query = generate_medium_media_query();
    var string large_media_query = generate_large_media_query();
``` 
in ` Page::print_default_stylesheet()`. These can then be used in `@media` CSS rules, such as [this one](https://github.com/dreamwidth/dreamwidth/blob/main/styles/core2base/layout.s2#L442). Styles that don't use the typical columns structure will probably also need additional CSS so they style nicely at smaller sizes.

cmho:
todo:
[x] Venture
[ ] EasyRead (sort of—it's already pretty mobile friendly, but just to check)
[ ] Goldleaf
[ ] Zesty (it's one column but also needs refurbishment probably)

PR #3120, opened 2023-06-17
Media query support for more layouts

cmho:
CODE TOUR: Venture now actually uses media queries to collapse the layout on smaller devices, and Easy Reader and Zesty got a couple little quality of life tweaks for mobile.

Addresses #3110

### DIFF
--- a/styles/venture/layout.s2
+++ b/styles/venture/layout.s2
@@ -1222,7 +1222,10 @@ $module_layout_css
 .one-column #content .module-content { text-align: left;}
 
 .two-columns-left #content,
-.three-columns-sides #content { border: none; }
+.three-columns-sides #content { border-left: $*sidebar_width solid transparent; }
+
+.two-columns-right #content,
+.three-columns-sides #content { border-right: $*sidebar_width solid transparent; }
 
 .three-columns-left #content { border-left: $*sidebar_width_doubled solid transparent; }
 
@@ -1232,19 +1235,14 @@ $module_layout_css
 
 #content { padding: 1em; }
 
-@media $medium_media_query {
-    .two-columns-left #content,
-    .three-columns-sides #content { border-right: $*sidebar_width solid transparent; }
-
-
-}
-
 
 /* ---------- PRIMARY COLUMN ---------- */
 #primary {
-    width: 100%;
-    margin: 0;
-}
+    float: left;
+    margin-right:-100%;
+    margin-left: 0;
+    width:100%;
+    }
 
 .one-column #primary { float: none; }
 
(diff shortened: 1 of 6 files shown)


rahaeli:
I am so excited for all this work you are doing to improve things, you have no idea! *\o/*

styles/zesty/layout.s2
```diff
@@ -753,13 +781,21 @@ h3 {
 }
 .userpic {
     margin-bottom: 5px;
+    width: 80%;
```
momijizukamori, line 784: This would need to go on `.userpic img` as userpics have their width and height hardcoded, and I feel like there was a reason we don't do that??? But of course I can't remember it. That said this is Zesty so I'm down for just putting the resize on the image itself and seeing what happens.

styles/venture/layout.s2
```diff
@@ -1217,118 +1228,138 @@ $module_layout_css
...
+    .three-columns-sides #content { border-left: $*sidebar_width solid transparent; border-right: $*sidebar_width solid transparent; }
 
-.column-right #content { min-width: $*sidebar_width; }
+    .three-columns-left #content { border-left: $*sidebar_width_doubled solid transparent; }
+
+    .two-columns-right #content,
+    .three-columns-right #content { border-right: $*sidebar_width_doubled solid transparent; }
+
+    .column-right #content { box-sizing: content-box; }
+}
```
momijizukamori, line 1248: I think something has gotten kind of off here, possibly because Venture's existing CSS seems like it might be wonky? The outer margins don't seem to be applying properly on any layout on desktop, and the two column versions have way too much space on the right-hand side of the page

momijizukamori (changes requested):

momijizukamori:
For ref on Venture:
image
new code

image
existing code

styles/venture/layout.s2
momijizukamori: wtf??? can you check http://momijizukamori.momiji.hack.dreamwidth.net/ and see if you get the same thing? I'm seeing it on both my 1920x1200 monitor and laptop (in both FF and Chrome), and I'm pretty sure I was looking on my other laptop originally. When I shrink the page enough to flip it into one-column mode it looks fine but before that, no.

cmho: okay there is *something* up here because it does in fact look like it does in your screencap on your dreamhack but it still looks like *my* screencap on *my* dreamhack

what. what is happening. is the computer haunted

cmho: okay it looks like the styles are in fact different between the two dreamhack instances so mmmmaybe i'll recompile styles once i'm back on my personal machine and in the meantime when you get the chance recompile yours and see if anything changes???

cmho: ok never mind i was the fool who needed to recompile. fix incoming shortly

styles/zesty/layout.s2
cmho: If you're willing to let me fuck around and find out with zesty I'm down for it, haha. I went back and checked and tweaked a couple things with userpic display on mobile (and also double-checked that it looks right on my phone) and also realized that we hadn't added the meta tags so I did that also.

t/directorysearch.t
```diff
@@ -72,13 +72,6 @@ $is->(
     "Russia", "loc_cn=RU&opt_sort=ut", LJ::Directory::Constraint::Location->new( country => 'RU' )
 );
 
-$is->(
-    "Age Range + last week",
-"loc_cn=&loc_st=&loc_ci=&ut_days=7&age_min=14&age_max=27&int_like=&fr_user=&fro_user=&opt_format=pics&opt_sort=ut&opt_pagesize=100",
-    LJ::Directory::Constraint::Age->new( from => 14, to => 27 ),
```
momijizukamori, line 78: pst I think you maybe meant to put this change on #3137

styles/zesty/layout.s2
momijizukamori: Yeah I'm willing to YOLO it up with Zesty, and it looks good now even with non-square icons for me.

styles/venture/layout.s2
```diff
@@ -1217,118 +1228,138 @@ $module_layout_css
...
 
-.three-columns-right #content { border-right: $*sidebar_width_doubled solid transparent; }
+@media $medium_media_query {
+    .two-columns-left #content,
+    .three-columns-sides #content { border-left: $*sidebar_width solid transparent; border-right: $*sidebar_width solid transparent; }
 
-.column-right #content { min-width: $*sidebar_width; }
+    .three-columns-left #content { border-left: $*sidebar_width_doubled solid transparent; }
+
+    .two-columns-right #content { border-right: $*sidebar_width solid transparent; }
```
momijizukamori, line 1243: I think something similar to this needs to be done but for `.two-columns-left #content` - it's showing up with the space still

styles/venture/layout.s2
```diff
@@ -1217,118 +1228,138 @@ $module_layout_css
...
 .two-columns-left #content,
-.three-columns-sides #content { border-left: $*sidebar_width solid transparent; }
+.three-columns-sides #content { border: none; float: none; }
 
-.two-columns-right #content,
-.three-columns-sides #content { border-right: $*sidebar_width solid transparent; }
+.column-right #content { min-width: $*sidebar_width; box-sizing: border-box; }
 
-.three-columns-left #content { border-left: $*sidebar_width_doubled solid transparent; }
+#content { padding: 1em; }
```
momijizukamori, line 1235: This is messing with the margins being set on line 1201 - you may need to change that one to just left and right and pull it into a media query (so desktop sizes get the margins but they collapse down on mobile)

t/directorysearch.t
cmho: oh shit you're right

styles/venture/layout.s2
cmho: oh good point... okay, i think i've unfucked this

PR #3122, opened 2023-06-17
Update colorpicker to something more modern

### LINKED ISSUE #3035: Update color picker element to something more modern
Form element should
- include a text element that updates with the picker value (and which in turn can update the picker value)
- work on all major browsers and versions from the last two years (Edge, Chrome, Firefox, Safari - including mobile versions)
- be keyboard/screenreader accessible

Code will need to be updated in:
- `cgi-bin/DW/Template/Plugin/FormHTML.pm`
- `cgi-bin/LJ/HTMLControls.pm`

momijizukamori:
* Replace terrible old colorpicker with Coloris (https://coloris.js.org/)
* Add CSS for dark theme for Coloris to Gradation CSS
* Nuke files for old colorpicker
* NB: this is a copy of Coloris that I rebuilt with a few minor changes - better handling of input clearing, and mirroring of cursor position/selection between the page and picker inputs

Fixes #3035

CODE TOUR: You may or may not know that if you go to customize your style (https://www.dreamwidth.org/customize/options?group=colors), if you click on the little color swatches, you get a colorpicker? It opens in a new window, and is 1) ancient, 2) terrible, 3) inaccessible, and 4) apparently difficult for a lot of people to find. It has now been nuked to oblivion and replaced with a nice little library that pops up in the page and which has way better support for screenreaders and keyboard navigation. It also comes in dark mode to match all you Gradation users out there.


cgi-bin/LJ/HTMLControls.pm
```diff
@@ -439,27 +422,14 @@ sub html_color {
             'maxlength' => 7,
             'name'      => $htmlname,
             'id'        => $htmlname,
-            'onchange'  => "setBGColorWithId(findel('${htmlname}_disp'),'${htmlname}');",
             'onfocus'   => $opts->{'onchange'},
             'disabled'  => $opts->{'disabled'},
             'value'     => $opts->{'default'},
             'noescape'  => 1,
-            'raw'       => $opts->{'raw'},
+            'raw'       => $opts->{'raw'} . "data-coloris",
```
zorkian, line 429: If there is raw here, will there be a trailing space or do we need to add one? Or is it okay if whatever is in raw is smashed up with `data-coloris`?

zorkian:
One comment, feel free to merge if it's not an issue.

cgi-bin/LJ/HTMLControls.pm
momijizukamori: Oof good catch. I don't think the places we call the color picker right now pass in any raw arguments buuuut I'd rather be safe for the future.

Issue #3038, opened 2022-11-21
Add syntax highlighting to the Custom CSS field


momijizukamori:
We already have CodeMirror in the codebase (for the advanced customization editor), so this should be fairly trivial - I think the only real work would be to make sure we have color themes in use that mesh well with the site scheme, and to double-check accessibility of the field.

PR #3123, opened 2023-06-17
Add CodeMirror to Custom CSS field

cmho:
CODE TOUR: Adds magical syntax highlighting to the Custom CSS field, so it's a little easier to work on code!

Screen Shot 2023-06-17 at 4 51 19 PM

Addresses #3038

### DIFF
--- a/cgi-bin/LJ/Widget/CustomizeTheme.pm
+++ b/cgi-bin/LJ/Widget/CustomizeTheme.pm
@@ -224,7 +224,7 @@ sub render_body {
         );
         $ret .= "</div>";
         $ret .= "\n\n<script type='text/javascript'>\n";
-        $ret .= "   \$(window).load(function() {\n";
+        $ret .= "   window.addEventListener('load', function() {\n";
         $ret .= "       let codeMirror = CodeMirror.fromTextArea(document.querySelector('[name=\"Widget[S2PropGroup]_custom_css\"]'), {\n";
         $ret .= "           mode:  'css',\n";
         $ret .= "           lineWrapping: true,\n";
--- a/cgi-bin/LJ/Widget/CustomizeTheme.pm
+++ b/cgi-bin/LJ/Widget/CustomizeTheme.pm
@@ -224,7 +224,7 @@ sub render_body {
         );
         $ret .= "</div>";
         $ret .= "<script type='text/javascript'>";
-        $ret .= "   $(document).ready(function() {";
+        $ret .= "   \$(window).load(function() {";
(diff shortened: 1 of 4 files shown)


momijizukamori:
I am not gonna merge this in just yet, because I think we need a dark mode to not accidentally blind people on Gradation, but thank you for getting this going!

momijizukamori:
Okay for the dark theme - it looks like it should be sufficient to include another CodeMirror theme file, and then before we load CodeMirror in, check the body tag for the `gradation` class and if it's there, set the `theme` key in the CodeMirror config to the name of the dark mode, and other set it to `default`. I will consult with people who actually use Gradation to see what they think meshes best. 

Also for some reason the tab-complete isn't working which is not the end of the world but slightly confusing. Time go to refresh my memory as to how I implemented it on the layer editor ahahaha.

momijizukamori:
Aha - for hinting, we need to include 1) `/js/vendor/codemirror/addon/show-hint.js`, , 2) `/stc/s2edit/show-hint.css` ,and 3) a hint file because *apparently* I only included S2 hinting. There's a [CSS one available](https://codemirror.net/5/addon/hint/css-hint.js) so we can just plop that in the same folder as the hint addon file and go from there.

momijizukamori:
Theme opinions from Discord:

> darcula and twilight I think vibe closest to gradation, but darcula I think is closest. that said, if you want the code text to have better differentiation of types of text, my picks would be material-darker and moxer.

cgi-bin/LJ/Widget/CustomizeTheme.pm
```diff
@@ -223,6 +223,27 @@ sub render_body {
             groupprops => $groups{groupprops}->{customcss},
         );
         $ret .= "</div>";
+        $ret .= "\n\n<script type='text/javascript'>\n";
```
zorkian, line 226: For this section, could we do something like this instead:

```perl
$ret .= <<EOF;

<script type='text/javascript'>
    ...
</script>

EOF
```

The reason for doing this is just to make the Perl cleaner / easier to read the underlying code, instead of having to stick in \n's and quotes and stuff. It's also faster to execute, although that's probably not something that matters heh.

cmho:
Hinting now works, and dark mode now uses the Twilight theme. And also thank you for the hot multiline string tip because I had no idea how to do that in perl 👍

momijizukamori:
Okay there is unfortunately not an easy answer here, but I think what makes the most sense to me is to change the binding from tab to tab + space, and add some help text like 'To insert indentation or open the code hint menu, press Tab + Space' above or below the field.

PR #3125, opened 2023-06-18
stop suspending yourself, stop suspending yourself

### LINKED ISSUE #3111: allow suspended accounts to use the admin console if they have the suspend priv
Way back in the dawn of LJ, you used to be able to unsuspend yourself if you accidentally suspended yourself! sadly, you can no longer, womp womp. ("Your account status prevents you from using the console.") With Discord discussion about possible splash damage from the autosuspender, we should put in a suspenders-and-belt check: if your account has the suspend priv, you should not be prevented from using the console to unsuspend yourself. (In the extremely unlikely case of needing to suspend someone with the suspend priv, sites will need to remove that priv first.)

kareila:
@rahaeli says this is the line of code that prevents an admin from reversing their own suspension if they accidentally suspend themselves. (Which with the amount of spam accounts we've been dealing with lately, is not that hard to do.)

It prevents any console command from being executed by a user who doesn't have a statusvis of V.

There are a LOT of other statusvis codes! Looking at the available helper methods, `is_inactive` seems like the right one to use. It encompasses the S (suspended), D (deleted), and X (expunged) codes. But it still needs to be overridden.

So: don't allow a deleted/suspended/purged user to run a console command... unless that user has the suspend priv.

Fixes #3111.

### DIFF
--- a/cgi-bin/LJ/Console/Command.pm
+++ b/cgi-bin/LJ/Console/Command.pm
@@ -125,7 +125,7 @@ sub execute_safely {
             if $cmd->requires_remote && !$remote;
 
         return $cmd->error("Your account status prevents you from using the console.")
-            if $cmd->requires_remote && !$remote->is_visible;
+            if $cmd->requires_remote && $remote->is_inactive && !$remote->has_priv("suspend");
 
         return $cmd->error("You are not authorized to run this command.")
             unless $cmd->can_execute;


zorkian:
Heh. Seems reasonable, I guess if we do have a situation where a stafff-with-suspend needs to be suspended, we just have to remove the privilege first.

rahaeli:
Yes, that was my thinking too! also now we can go back to doing the thing where you all suspend me when I'm allegedly on vacation and I can unsuspend myself when I get back

PR #3137, opened 2023-06-22
Remove references to search-by-age and remove capability on backend

### LINKED ISSUE #3065: Directory search: remove "age" as a search option
I've been looking at the directory search page lately for the first time pretty much since we forked from LJ because it's been helpful in spammer fighting, and having actually looked at the search options, it's made me realize that we absolutely need to stop including "age" as an option at all, on user safety grounds. (The directory search runs off the age specified on the profile, which is 100% optional to include and does not have to be a user's real age, and often isn't thanks to RP accounts and the like, but it's still a ridiculously basic violation of user-safe design 101 to include the search option at all.)

cmho:
CODE TOUR: Removes search-by-age in the directory, because that's a weird holdover from the LJ days that we don't actually want on account of That's Weird.

Closes #3065.


kareila:
I think this is missing the changes to `directorysearch.t` I saw you accidentally push elsewhere? Let us know if I'm wrong and this is considered complete as is.

cmho:
> I think this is missing the changes to `directorysearch.t` I saw you accidentally push elsewhere? Let us know if I'm wrong and this is considered complete as is.

@kareila I was considering changing it but I just ended up deleting it so I've pushed the new change here now 👍🏼

kareila:
So I think what you have here looks good! I would just add a few more things:

* delete the `cgi-bin/LJ/Directory/Constraint/Age.pm` file

* delete the `cgi-bin/LJ/Directory/SetHandle/Age.pm` file

* remove references to `age_min` and `age_max` in t/directorysearch-extra.t

cmho:
@kareila re-requesting review since i did the changes you mentioned!

kareila:
I don't see any changes since my last review - did you forget to push the updated branch?

cmho:
@kareila these changes? https://github.com/dreamwidth/dreamwidth/pull/3137/commits/b2d997e9a305e85da7d7a73acfb6d61465da107d they're from last night

kareila:
... maybe my tab didn't refresh? ??? I see it now...

kareila:
Oops! I found two more references to LJ::Directory::Constraint::Age in t/directorysearch.t. Sorry I didn't notice them on my last review!

If you will remove the tests labelled "test ages" and "test ages after filter" starting around line 235, we should be good here. 🤞

PR #3138, opened 2023-06-23
Convert /manage/circle/invite to TT

momijizukamori:
CODE TOUR: Taking a 🔪 to more BML

### DIFF
--- a/cgi-bin/DW/Controller/Manage/Circle/Invite.pm
+++ b/cgi-bin/DW/Controller/Manage/Circle/Invite.pm
@@ -1,0 +1,187 @@
+#!/usr/bin/perl
+#
+# DW::Controller::Manage::Circle::Invite
+#
+# /manage/circle/invite
+#
+# Authors:
+#      Cocoa <momijizukamori@gmail.com>
+#
+# Copyright (c) 2023 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.  For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+#
+
+package DW::Controller::Manage::Circle::Invite;
+
+use strict;
+
+use DW::Controller;
+use DW::Routing;
+use DW::Template;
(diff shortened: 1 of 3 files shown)


kareila (changes requested):
Haven't tested this, but I saw a few minor Perl things on a quick readthrough that I thought should be addressed before merging.

cgi-bin/DW/Controller/Manage/Circle/Invite.pm
```diff
@@ -0,0 +1,184 @@
...
+        if ($email) {
+            my @errs;
+            LJ::check_email( $email, \@errs, $POST, \$email_checkbox );
+            $errors->add( "email", @errs ) if @errs;
+
+            if ( $LJ::USER_EMAIL && $email =~ /$LJ::USER_DOMAIN$/ ) {
+                $errors->add( "email", '.error.useralreadyhasaccount' );
+            }
+
+            unless ($LJ::USE_ACCT_CODES) {
```
kareila, line 79: Hmm, another thing I just noticed. This block is nested within a `if ($LJ::USE_ACCT_CODES)` block, so it will never execute. I would suggest taking a top level view of your conditionals and making sure they're properly aligned.

cgi-bin/DW/Controller/Manage/Circle/Invite.pm
```diff
@@ -0,0 +1,184 @@
...
+
+                $u->log_event(
+                    'friend_invite_sent',
+                    {
+                        remote => $u,
+                        extra  => $email,
+                    }
+                );
+
+                if ($LJ::USE_ACCT_CODES) {
```
kareila, line 128: Similar to above, nested `if ($LJ::USE_ACCT_CODES)` block.

cgi-bin/DW/Controller/Manage/Circle/Invite.pm
kareila: Oh okay, never mind then! My bad, I was just reviewing it in the browser window and didn't see where the cutoff was.

kareila (approved):

PR #3139, opened 2023-06-23
Create styling for dead services that indicate that they're no longer available

cmho:
CODE TOUR: I went on an arduous journey and learned a lot about arrays in Perl to tweak username display so that dead services aren't linked out and display a default userhead.

### DIFF
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -221,7 +221,7 @@ sub get_site {
     return $mapped || undef;
 }
 
-sub get_deadsites { return \@deadsites; }
+sub get_deadsites { return @deadsites; }
 
 # returns a list of all supported sites for linking
 sub get_sites { return @all_sites_without_alias; }
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -160,6 +160,8 @@ $domaintosite{"artstation"}      = $domaintosite{"artstation.com"};
 $domaintosite{"substack"}        = $domaintosite{"substack.com"};
 $domaintosite{"itch"}            = $domaintosite{"itch.io"};
 
+@my deadsites = qw(delicious delicious.com diigo imzy inksome journalfen);
+
 foreach my $value (@all_sites_without_alias) {
     $idtosite{ $value->{siteid} } = $value;
 }
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -111,7 +111,7 @@ $domaintosite{"furaffinity.com"} =
 $domaintosite{"artstation.com"} =
     DW::External::Site->new( "33", "www.artstation.com", "artstation.com", "ArtStation", "artstation" );
 
-@deadsites =  qw(delicious delicious.com diigo imzy inksome journalfen);
+@deadsites = ("del.icio.us", "diigo.com", "imzy.com", "inksome.com", "journalfen.net");
 
 @all_sites_without_alias = values %domaintosite;
 
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -221,7 +221,7 @@ sub get_site {
     return $mapped || undef;
 }
 
-sub get_deadsites { return @deadsites; }
+sub get_deadsites { return \@deadsites; }
 
 # returns a list of all supported sites for linking
 sub get_sites { return @all_sites_without_alias; }
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -160,7 +160,7 @@ $domaintosite{"artstation"}      = $domaintosite{"artstation.com"};
 $domaintosite{"substack"}        = $domaintosite{"substack.com"};
 $domaintosite{"itch"}            = $domaintosite{"itch.io"};
 
-@my deadsites = qw(delicious delicious.com diigo imzy inksome journalfen);
+my @deadsites = qw(delicious delicious.com diigo imzy inksome journalfen);
 
 foreach my $value (@all_sites_without_alias) {
     $idtosite{ $value->{siteid} } = $value;
--- a/cgi-bin/DW/External/Site.pm
+++ b/cgi-bin/DW/External/Site.pm
@@ -31,6 +31,7 @@ LJ::ModuleLoader->require_subclasses("DW::External::Site");
 my %domaintosite;
 my %idtosite;
 my @all_sites_without_alias;
+my @deadsites;
 
 ### static initializers
 # with tld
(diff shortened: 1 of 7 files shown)


cgi-bin/DW/External/Site/Imzy.pm
```diff
@@ -53,16 +53,13 @@ sub profile_url {
...
-# argument: DW::External::User
-# returns info for the badge image (userhead icon) for this user
 sub badge_image {
     my ( $self, $u ) = @_;
     croak 'need a DW::External::User'
         unless $u && ref $u eq 'DW::External::User';
 
-    # for lack of anything better, let's use the favicon
     return {
-        url    => $LJ::IMGPREFIX . '/external/imzy.png',
```
kareila, line 65: Since this is the only place in the code using `htdocs/img/external/imzy.png`, we should also remove that file from the repository.

cgi-bin/DW/External/Site/Inksome.pm
```diff
@@ -48,18 +48,15 @@ sub journal_url {
...
-# argument: DW::External::User
-# returns info for the badge image (head icon) for this user
 sub badge_image {
     my ( $self, $u ) = @_;
     croak 'need a DW::External::User'
         unless $u && ref $u eq 'DW::External::User';
 
-    # Inksome went away, so just assume every account is personal
     return {
-        url    => "$LJ::IMGPREFIX/external/ink-userinfo.gif",
```
kareila, line 60: Same for `htdocs/img/external/ink-userinfo.gif`.

cgi-bin/DW/External/Site/JournalFen.pm
```diff
@@ -35,25 +35,15 @@ sub accepts {
...
 sub badge_image {
     my ( $self, $u ) = @_;
     croak 'need a DW::External::User'
         unless $u && ref $u eq 'DW::External::User';
 
-    my $type = $self->journaltype($u) || 'P';
-    my $gif  = {
-        P => [ '/external/lj-userinfo.gif',   17, 17 ],
-        C => [ '/external/lj-community.gif',  16, 16 ],
-        Y => [ '/external/lj-syndicated.gif', 16, 16 ],
```
kareila, line 49: ... but not these, because they are also used in LiveJournal.pm.

cgi-bin/DW/External/User.pm
```diff
@@ -69,13 +71,17 @@ sub ljuser_display {
     my $display_class = $opts{no_ljuser_class} ? "" : " class='ljuser'";
     my $domain = $self->site->{domain} ? $self->site->{domain} : $self->site->{hostname};
 
+    $nolink = 1 if (first {$domain eq $_} @DW::External::Site::deadsites);
```
kareila, line 74: I appreciate that you went on a journey 😂 but it's much more typical to use a hash than an array here:
```
my %deadsites;
...
$nolink = 1 if exists $deadsites{$domain};
```
Let us know if you need a more detailed explanation!

kareila (changes requested):
Thanks so much for working on this! I hope you are enjoying the process 😅

cgi-bin/DW/External/User.pm
```diff
@@ -69,13 +70,17 @@ sub ljuser_display {
...
+
     return
-          "<span style='white-space: nowrap;'$display_class>"
-        . ( $opts{no_link} ? '' : "<a href='$profile_url'>" )
+        "<span style='white-space: nowrap;'$display_class>"
+        . ( $nolink ? '' : "<a href='$profile_url'>" )
         . "<img src='$badge_image->{url}' alt='[$domain profile] ' style='vertical-align: text-bottom; border: 0; padding-right: 1px;' width='$badge_image->{width}' height='$badge_image->{height}'/>"
-        . ( $opts{no_link} ? '' : "</a><a href='$journal_url'>" )
+        . ( $nolink ? '' : "</a><a href='$journal_url'>" )
         . "<b>$user</b>"
```
kareila, line 82: Revisiting the original issue, it looks like you should be printing something like `"<b>$user [$site]</b>" ` here if `exists $DW::External::Site::deadsites{$domain}` is true?

cgi-bin/DW/External/User.pm
```diff
@@ -69,13 +70,17 @@ sub ljuser_display {
     my $display_class = $opts{no_ljuser_class} ? "" : " class='ljuser'";
     my $domain = $self->site->{domain} ? $self->site->{domain} : $self->site->{hostname};
 
+    $nolink = 1 if exists $DW::External::Site::deadsites{$domain};
```
kareila, line 73: I would suggest combining this line and the initial nolink declaration as follows:

`my $nolink = exists $DW::External::Site::deadsites{$domain} ? 1 : 0;`

This is called the ternary operator 😄

cgi-bin/DW/External/User.pm
```diff
@@ -69,13 +70,17 @@ sub ljuser_display {
...
+
+    $nolink = $nolink || $opts{no_link};
```
kareila, line 75: This can also be shortened to:

`$nolink ||= $opts{no_link};`

cgi-bin/DW/External/User.pm
```diff
@@ -20,6 +20,7 @@ package DW::External::User;
 use strict;
 use Carp qw/ croak /;
 use DW::External::Site;
+use List::Util qw(first);
```
kareila, line 23: Is this still being used after changing to a hash?

cgi-bin/DW/External/Site.pm
```diff
@@ -218,6 +225,8 @@ sub get_site {
     return $mapped || undef;
 }
 
+sub get_deadsites { return %deadsites; }
```
kareila, line 228: If this is going to be a function, I would just declare the hash inside the function since we're not using it in any other context:

```
sub get_deadsites {
  return (
              "del.icio.us" => 1,
              "diigo.com"   => 1,
              "imzy.com"    => 1,
              "inksome.com" => 1,
              "journalfen.net" => 1
  );
}
```

kareila (changes requested):
I feel like this is getting close! Adding [site] to the printed username is something requested from the original issue that I just noticed was missing; the rest of my comments are mere syntax suggestions. 👍

cgi-bin/DW/External/User.pm
kareila: Since we're now using a `get_deadsites` function, this syntax needs to be updated:

```
my %deadsites = DW::External::Site::get_deadsites();
my $nolink = exists $deadsites{$domain} ? 1 : 0;
```

Sorry for not noticing that in my previous review! I **think** that's the last update necessary for this to work.

I would also highly encourage you to make sure your Perl edits are tidy going forward. @momijizukamori posted a pre-commit hook in dw-dev recently to help make this easier!

zorkian:
Updated and merged, thank you @cmho and @kareila! :)

PR #3140, opened 2023-06-23
Fix overflow of some HTML elements from /latest entries

### LINKED ISSUE #3086: HTML overflow on /latest
Noticed while spamhunting. Something about the HTML in https://lyssabn.dreamwidth.org/2534.html (I think it's that they have a &lt;nav> tag in there) is causing elements of the entry to show up at the top of the /latest page (see screenshot); we should add that to the extra HTML cleaning that /latest gets 
Screen Shot 2023-04-16 at 6 46 49 PM

cmho:
CODE TOUR: Some HTML elements were picking up site style CSS due to overly-broad style scoping which led to some elements appearing wildly outside of the container, so this anchors absolutely-positioned items to the entry itself and also narrows the scope of some site styles in the Tropospherical layouts.

(also if someone could test this on an env where /latest works i'd be much obliged)

Closes #3086.

### DIFF
--- a/ext/dw-nonfree/htdocs/stc/tropo/tropo-base.css
+++ b/ext/dw-nonfree/htdocs/stc/tropo/tropo-base.css
@@ -137,7 +137,7 @@ body {
  * The menu follows the content in the markup source, so we absolutely position it underneath the masthead
  */
 
-nav {
+nav[role="navigation"] {
     position: absolute;
     top: 8.833333em;
     left: 0;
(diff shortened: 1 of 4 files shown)


kareila:
I have /latest working in my dev environment. Can you provide some examples of entries that would cause the overflow behavior?

rahaeli:
The ones i definitely remember are unclosed divs and unclosed ul/ol

momijizukamori:
The original triggering post for this issue for this has a `<nav>` tag in it, which is properly closed but we guessed might be getting unwanted CSS styling from the site scheme CSS. Probably want to try an entry with a nav element in it before and after this patch.

PR #3142, opened 2023-06-24
Hide website link for accounts under ten days old

### LINKED ISSUE #2795: restrict visibility of URL link for 10 days on newly created account profiles
The vast majority of our spam is SEO spammers creating an account, wodging their spammy URL into the URL field of the profile, posting a spammy entry or two, and never touching the account again. We are generally very good at finding them within a reasonable amount of time, but a simple change that might make us less appealing for spammer account creation: people should be able to add the URL to their account profile whenever, but it shouldn't display to anyone other than the owner of the account for the first 10 days after account creation. (Ideally, with a note next to the field so if the user looks at their profile and doesn't see the URL, they don't think it's a bug: "Other people will not see this link on your profile until your account is at least 10 days old.")

cmho:
CODE TOUR: To help discourage spammers from registering accounts here, we want to hide the website link on the profile until the account is ten days old.

Closes #2795.

### DIFF
--- a/cgi-bin/DW/Logic/ProfilePage.pm
+++ b/cgi-bin/DW/Logic/ProfilePage.pm
@@ -535,7 +535,7 @@ sub _basic_info_website {
         $urlname = LJ::ehtml( $urlname || $url );
         if ($url) {
             $ret->[0] = _profile_ml('.label.website');
-            my $spam_time_threshold = (localtime() - (ONE_DAY * 10)) > LJ::mysql_time( $u->timecreate );
+            my $spam_time_threshold = scalar (localtime() - (ONE_DAY * 10)) <= scalar LJ::mysql_time( $u->timecreate );
             if ($spam_time_threshold) {
                 $ret->[1] = { url => $url, text => $urlname };
             } else {
--- a/cgi-bin/DW/Logic/ProfilePage.pm
+++ b/cgi-bin/DW/Logic/ProfilePage.pm
@@ -21,6 +21,8 @@ package DW::Logic::ProfilePage;
 use strict;
 use DW::Countries;
 use DW::Logic::UserLinkBar;
+use Time::Piece;
+use Time::Seconds 'ONE_DAY';
 
 # returns a new profile page object
 sub profile_page {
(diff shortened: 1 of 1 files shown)


kareila:
This is a good start, but it doesn't entirely meet the requested spec. Just at a glance:

* it seems to show the "your profile" message to everyone who views the profile, not just the journal owner?
* it should still show the actual link to logged-in users with suspend privileges.

cmho:
@kareila I think this should hit all the requirements now?

kareila:
Okay, deep breaths. This is like, 70% of the way there! You're doing great.

I am going to break this review into three segments.

1.

`my $spam_time_threshold = scalar (localtime() - (ONE_DAY * 10)) <= scalar LJ::mysql_time( $u->timecreate );`

^^ This line gave me an error: `Argument "2022-10-18 22:31:21" isn't numeric in localtime`

I spent half an hour trying different variations on this to get the right output, so you don't have to. Here's what I ended up with:

`my $spam_time_threshold = (DateTime->now->epoch - (ONE_DAY * 10)) > $u->timecreate;`

Notice I reversed the direction of the comparison, because the link will display normally if the expression is true.

2.

`if ($remote->has_priv('suspend') || $spam_time_threshold) {`

If the page is being viewed logged out, this will fail because $remote will be unset. Also, you've already calculated $spam_time_threshold, and we expect it to usually be true, so it makes sense to check that first. So here's the way I recommend rewriting that:

`if ( $spam_time_threshold || ( $remote && $remote->has_priv('suspend') ) ) {`

3.

As written, the "Other people will not see this link..." message will always display to the user if their account is less than 10 days old, but I think we only want to display it if the user has saved a URL. So I would move all this back into the `if ($url) {` conditional - the first one, line 527 in the original file.

... Phew, I think that's everything! Thanks for hanging in there <3

alierak:
For what it's worth, I'd never heard of DateTime->now->epoch, so I checked what it does, and it's equivalent to just "time", as in `my $spam_time_threshold = (time - (ONE_DAY * 10)) > $u->timecreate;`

cmho:
ok i thiiiiink that should do it (also tbh don't worry about the big paragraph of feedback; this is still fewer comments than i get from work lmao and that's in a language i'm *good* at)

kareila:
> For what it's worth, I'd never heard of DateTime->now->epoch, so I checked what it does, and it's equivalent to just "time", as in `my $spam_time_threshold = (time - (ONE_DAY * 10)) > $u->timecreate;`

Well sure, and `ONE_DAY * 10` is equivalent to 864000, but there's an argument to be made that there are times when a bit of additional complexity makes the code's intent clearer to the casual observer.

kareila:
Okay, three things:

1. The direction of the inequality in the time comparison row is still backwards from what I think it should be. Have you tested it?
2. This backs out the change from #3212
3. Please, please, please tidy your code changes before submitting.

PR #3148, opened 2023-06-24
Add nofollow to all user-supplied links on profile

### LINKED ISSUE #3066: Add "nofollow" to all user-supplied links on profiles
I have been *strongly* resisting doing this for 14 years, but it's time, sigh: we need to start adding the nofollow attribute to all user-supplied links on a user profile, including the Website field, all social media profiles (from the "other sites" section), any user-constructed links in the profile bio (using an &lt;a href> tag), and any URLs that have been auto-linked by the HTML cleaner in the profile bio.

Grr, arrgh, etc.

If it's *considerably* easier, we can just add a nofollow to the headers on the profile page, but I'd prefer to keep on-site links (subscribers, etc) not having nofollow. If the amount of effort it would take to delve into the HTML cleaner would be greater than an extra hour or two, though, just add the nofollow to the header on all /profile pages.

cmho:
CODE TOUR: We're adding the rel="nofollow" attribute to all user-supplied links to try and cut down on spam.

Closes #3066

(Also can someone who can verify their account email on their instance check this for me >_>)

### DIFF
--- a/cgi-bin/DW/Controller/Profile.pm
+++ b/cgi-bin/DW/Controller/Profile.pm
@@ -163,7 +163,7 @@ sub profile_handler {
         if ( $l->{text} ) {
             my $ret = "";
             $ret .= $l->{secimg} if $l->{secimg};
-            $ret .= $l->{url} ? qq(<a href="$l->{url}">$l->{text}</a>) : $l->{text};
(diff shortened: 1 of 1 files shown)


zorkian:
Looks reasonable!

PR #3161, opened 2023-06-28
convert and reorganize /manage/emailpost

### LINKED ISSUE #1853: Advanced Setup Options could be more visible on /manage/settings/?cat=mobile
It's relatively easy to see when you set up your account for the first time but easy to miss when you just want to check your options because you have to scroll down (and not so much of a reason to do so if you don't know advanced is down there), and I missed it even when I did. I decided to click on 'View more posting instructions.' on the odd chance that it would lead me to the correct place. Is there a way to make it more visible? I'd love to have all options there actually (I don't see the need for more clicking) but this would be the next best thing. Thoughts?

### LINKED ISSUE #3021: Make "Generate API key" more discoverable
At the moment it's a bit difficult to navigate to /manage/emailpost to find where to generate an API key. If you use the menu (Organize / Manage Account / Mobile tab), that takes you to /manage/settings?cat=mobile, where there is a link to "Advanced Setup Options" at the bottom of the Email Posting section. The layout and location of the link sort of implies that it's only going to be related to email posting, and it's not at all obvious that will be the place to generate an API key.

kareila:
CODE TOUR: This converts /manage/emailpost (including the help pages) away from BML, removes the existing form from the Mobile tab of Account Settings (because it was entirely duplicating what was on /manage/emailpost), and moves the API Key and Reset Reply Address settings out onto the Mobile tab (because they weren't email-related, people were having trouble locating them - now they are more discoverable, yay!)

Closes #1853. Closes #3021.

Screenshot of new settings tab:

Screenshot 2023-06-28 at 3 11 29 PM


cgi-bin/DW/Controller/Manage/EmailPost.pm
```diff
@@ -0,0 +1,191 @@
...
+
+sub emailpost_handler {
+    my $ml_scope = "/manage/emailpost.tt";
+    return error_ml("$ml_scope.error.sitenotconfigured") unless $LJ::EMAIL_POST_DOMAIN;
+
+    my ( $ok, $rv ) = controller( anonymous => 0, form_auth => 1 );
+    return $rv unless $ok;
+
+    my $u = $rv->{u};
+    return DW::Template->render_template( 'error.tt', { message => $LJ::MSG_READONLY_USER } )
```
zorkian, line 45: vague thought that is totally out of scope of this PR, I wonder if this should just be part of calling `controller`. Most of the time, we don't want to allow readonly users, particularly if form_auth is 1 and anonymous is 0, so maybe this could be done automatically... I know I never remember to check whether a user is readonly or not, so I bet there's a zillion paths where we don't even try.

cgi-bin/DW/Controller/Manage/EmailPost.pm
```diff
@@ -0,0 +1,191 @@
...
+
+    return error_ml("$ml_scope.error.acct") unless $u->can_emailpost;
+
+    my @emailpost_props = qw/
+        emailpost_pin emailpost_allowfrom
+        emailpost_userpic emailpost_security
+        emailpost_comments
+        /;
+    $u->preload_props(@emailpost_props);
+    $rv->{u} = $u;
```
zorkian, line 56: why set this? I don't think `$u` can change and above this came from `my $u = $rv->{u}` so this seems redundant.

cgi-bin/DW/Setting/ApiKeyDelete.pm
```diff
@@ -0,0 +1,104 @@
...
+    return 1;
+}
+
+sub save {
+    my ( $class, $u, $args ) = @_;
+
+    my $apikeys = DW::API::Key->get_keys_for_user($u);
+    my $numkeys = scalar @{$apikeys};
+
+    for ( my $idx = 0 ; $idx < $numkeys ; $idx++ ) {
```
zorkian, line 94: it's a nit, but I prefer the `foreach my $idx ( 0 .. $numkeys )` syntax in general in Perl.

cgi-bin/DW/Setting/ApiKeyDelete.pm
```diff
@@ -0,0 +1,104 @@
...
+sub label {
+    return '<nobr>' . $_[0]->ml('setting.apikeydelete.label') . '</nobr>';
+}
+
+sub option {
+    my ( $class, $u, $errs, $args ) = @_;
+
+    my $ret = '<p>' . $class->ml('setting.apikeydelete.des') . '</p>';
+
+    my $apikeys = DW::API::Key->get_keys_for_user($u);
```
zorkian, line 39: It's been a long time since I've thought about the settings page modules... I assume that authorization, authas, all that is handled at the layer above? i.e., we don't need to worry about "is $remote actually $u, or do they have authorization to act as $u"?

cgi-bin/DW/Setting/ApiKeyGenerate.pm
```diff
@@ -0,0 +1,88 @@
...
+            name     => $class->pkgkey . "keygen",
+            value    => 1,
+            selected => $error ? 1 : 0,
+            label    => $class->ml('setting.apikeygenerate.act'),
+        }
+    );
+
+    my $lastkey;
+    my $apikeys = DW::API::Key->get_keys_for_user($u);
+    if (@$apikeys) {
```
zorkian, line 51: nit: this can print out one of those warnings you love so much ("coercing undef to arrayref" kind of thing) since we don't know if $apikeys is undef or not.

cgi-bin/DW/API/Key.pm
```diff
@@ -99,7 +99,8 @@ sub get_keys_for_user {
         push @keylist, $new;
     }
 
-    return \@keylist;
+    # sort oldest to newest (predictable ordering for management page)
+    return [ sort { $a->{keyid} <=> $b->{keyid} } @keylist ];
```
zorkian, line 103: nit: in principle, I don't love putting presentation logic down into the model layer, but this definitely doesn't hurt anything and does mean we don't have to repeat sorting logic in different areas... :handwaving:

zorkian:
Merging, I didn't see anything that would be broken or show-stoppers, just some nits. I did not hard-review the templates though, but the settings code looks good.

cgi-bin/DW/Controller/Manage/EmailPost.pm
kareila: I did that because calling `preload_props` adds hash keys and values to `$u` and I wanted to make that more evident.

cgi-bin/DW/Controller/Manage/EmailPost.pm
kareila: I can open a new issue for that.

cgi-bin/DW/Setting/ApiKeyGenerate.pm
kareila: Yeah, it looks like it should - I tested the empty keys case, though, and didn't see an error. Go figure.

PR #3171, opened 2023-07-02
Don't impose a text-align on header navlinks in TR

momijizukamori:
CODE TOUR: The CSS for header navlinks in Tabula Rasa was being too opinionated and causing problems with other child layouts, so we fixed it.


kareila:
This and #3170 are already live.

PR #3180, opened 2023-07-06
[#3029] allow listing of multiple accounts for a service on user's profile

### LINKED ISSUE #3029: Allow adding more than one of the same type of site to the 'other services'
Particularly relevant with the discussion in #3003 (tl;dr, if we add fediverse stuff, there's a number of very diverse site types which increases the likelihood of a user wanting to list more than one). I thiiiink the way we store userprops means this would require some db rejiggering, and we would want to cap the max number of listed services overall.

UI-wise, I think my ideal would be a 'add another' sort of UI where each row has a dropdown to select the site, and then a text input for the username. Probably best implemented whenever we get around to the profile page conversion (someday.............)

kareila:
CODE TOUR: This rewrites the "Other Sites" section of /manage/profile to allow for specifying more than one account per service type. Closes #3029.

The form UI could use more polish, but it's does the job. I left the preexisting table layout intact and just changed the labels to dropdowns.

Before:
Screenshot 2023-07-05 at 8 16 33 PM

After:
Screenshot 2023-07-05 at 8 15 45 PM


bin/upgrading/base-data.sql
```diff
@@ -1,5 +1,4 @@
-# This file is automatically generated from MySQL by $LJHOME/bin/dumpsql.pl
-# Don't submit a diff against a hand-modified file - dump and diff instead.
+# This file is no longer automatically generated by $LJHOME/bin/dumpsql.pl
```
zorkian, line 1: interesting, I guess I'm not opposed, but I'm curious on this. did the maintenance burden of managing dumpsql to add the profile services table feel too high?

bin/upgrading/update-db-general.pl
```diff
@@ -3120,6 +3120,31 @@
...
+
+# clustered
+register_tablecreate( "user_profile_accts", <<'EOC');
+CREATE TABLE user_profile_accts (
+    userid INT(10) UNSIGNED NOT NULL,
+    name VARCHAR(40) NOT NULL,
+    value VARCHAR(255) NOT NULL,
+
+    PRIMARY KEY (userid, name, value),
+    INDEX (userid)
```
zorkian, line 3144: this is implied by having the primary key (since a primary key is an index), so this line is unnecessary (I don't think it does anything?)

bin/upgrading/update-db-general.pl
```diff
@@ -3120,6 +3120,31 @@
 )
 EOC
 
+register_tablecreate( "profile_services", <<'EOC');
+CREATE TABLE profile_services (
+    name VARCHAR(40) NOT NULL,
```
zorkian, line 3125: we should use an index as the primary key so that services can rename themselves. with the table design as-is, we can't rename a service.

the traditional method here is to use an `auto_increment` id column, then you don't really have to think about it, look at `priv_list` as an example.

your keys would then be a primary key on profile_service_id and a unique key on name (since presumably we want to enforce uniqueness, but also allow to lookup by name)

```sql
create table profile_services (
    profile_service_id mediumint(8) unsigned NOT NULL auto_increment,
    ...

    primary key (profile_service_id),
    unique key (name)
)```

you would then change the `user_profile_accts` table to contain the profile_service_id instead of the name

Comment (zorkian) [reply bin/upgrading/update-db-general.pl:3125]: this is what I get for doing a top down review and commenting. I see you already considered this and are putting the ML string into the db in a separate column. so, the "need" for using an integer is less and makes this more of a nit. I guess it's okay.

File: cgi-bin/DW/External/ProfileServices.pm [hunk 4]
```diff
@@ -0,0 +1,175 @@
...
+    my $memkey   = 'profile_services';
+    my $services = LJ::MemCache::get($memkey);
+    return $services if $services;
+
+    # load services from database and add to memcache (expiring hourly)
+    my $dbr  = LJ::get_db_reader();
+    my $data = $dbr->selectall_hashref(
+        "SELECT name, userprop, imgfile, title_ml, url_format, maxlen FROM profile_services",
+        "name" );
+    die $dbr->errstr if $dbr->err;
```
zorkian, line 37: nit: confess vs die?

cgi-bin/DW/External/ProfileServices.pm
```diff
@@ -0,0 +1,175 @@
...
+
+    # expire memcache after updating db
+    my $uid    = $u->userid;
+    my $memkey = [ $uid, "profile_accts:$uid" ];
+
+    return unless $u->writer;
+
+    my %sites = map { $_->{name} => 1 } @{ DW::External::ProfileServices->list };
+
+    # if %$old_accts is empty, we need to clear out the user's legacy userprops
```
zorkian, line 105: this is technically not safe to use as a trigger for writes, since this value can come purely from memcache (and memcache is a loosely consistent cache, there's nothing that guarantees it's in sync with the database)

while it probably never will matter, if you want to be safe, the pattern here is to add a force parameter to this loader so that it bypasses/refreshes memcache and returns the database value.

cgi-bin/DW/External/ProfileServices.pm
```diff
@@ -0,0 +1,175 @@
...
+                    $ret{$name} //= [];
+                    push @{ $ret{$name} }, $val;
+                }
+            }
+        }
+        return %ret;
+    };
+
+    my %del = $process->( \%old, \%new );
+    my %add = $process->( \%new, \%old );
```
zorkian, line 152: this should probably enforce some kind of limit to how many external services you can link?

cgi-bin/DW/External/ProfileServices.pm
```diff
@@ -0,0 +1,175 @@
...
+
+    foreach my $name ( keys %del ) {
+        foreach my $val ( @{ $del{$name} } ) {
+            $u->do( "DELETE FROM user_profile_accts WHERE userid = ? AND name = ? AND value = ?",
```
zorkian, line 156: confess if err?

cgi-bin/DW/External/ProfileServices.pm
```diff
@@ -0,0 +1,175 @@
...
+                undef, $uid, $name, $val );
+        }
+    }
+
+    foreach my $name ( keys %add ) {
+        foreach my $val ( @{ $add{$name} } ) {
+            $u->do( "INSERT INTO user_profile_accts (userid, name, value) VALUES (?,?,?)",
```
zorkian, line 163: confess if err?

htdocs/manage/profile/index.bml
```diff
@@ -403,45 +405,64 @@ body<=
...
+            my @services = map $service_info->($_), @{ DW::External::ProfileServices->list };
+            my @dropdown = ( '' => '' );
+            push @dropdown, ( $_->{name} => $_->{title} ) foreach @services;
+
+            my $service_td = sub {
+                my ( $p, $val ) = @_;
+                $oddeven = !$oddeven;
+                $ret .= "<tr class='field_block'>" if $oddeven;
+                $ret .= "<td class='field_name' width='20%'>$p->{title}</td><td width='30%'>";
+                $ret .= LJ::html_hidden( "extservice_name_$ct", $p->{name} );
```
zorkian, line 425: this puts an implicit requirement on the name field that it can only be "stuff that's safe to put in a field name." So, no hyphens or periods, other symbols. we should call this out up in the database schema/inserts that the name field has to be `[a-z0-9_]` or something.

htdocs/manage/profile/index.bml
```diff
@@ -693,12 +714,29 @@ body<=
...
 
+            # update external services
+            my %services = map { $_->{name} => $_ } @{ DW::External::ProfileServices->list };
+            my %new_accts;
+
+            foreach my $ct (1..26) {
+                my $name = $POST{"extservice_name_$ct"};
+                next unless $name;
+                my $val = $POST{"extservice_val_$ct"} // '';
+                # this isn't great but it's better than not enforcing maxlen at all
```
zorkian, line 732: should we should use the `LJ::text_trim` (I think that's the name) so that we don't cut multibyte characters in half? (I think that's a risk here?)

zorkian:
Okay, so overall, I think the code would work. I would like to suggest though two changes to the schema which will make it more robust against future issues, plus be more in line with general best practice.

First is the switch to using an integer as the primary key for the profile services, instead of using the name. Since the name is later embedded as a key, and it's a string-as-index, these will definitely work but they're probably going to be surprising design patterns to folks coming along and looking at this, particularly compared to the rest of the codebase. I think it would also be worth avoiding any issues with "a profile with a name that contains a symbol ends up breaking profiles" situations.

The second is that, for the user storage, I would suggest using the `alloc_user_counter` system and giving an ID per row. That way adding a new reference to users is like:

```perl
my $pid = $u->alloc_user_counter( 'whatever' );
$u->do( "INSERT INTO user table (userid, $pid, $profile_service_id, $value)
```

This means that the delete logic is simpler, since it's just a row of checkboxes that have IDs and you can delete "where userid = ? and profile_id = ?"). Actually, if it helps, the privilege system uses this exact structure (except it uses auto_increment on the priv_map table, instead of making this a per-user field.)

But anyway, look at priv_list (equivalent to the external service list) and priv_map (equivalent to "users who have these, including multiple, with custom values per.")

bin/upgrading/base-data.sql
alierak: We found that dumpsql.pl would no longer have any effect since you removed schematables 9 years ago.

zorkian: love it. good bye, dumpsql, good bye.

cgi-bin/DW/External/ProfileServices.pm
kareila: Right now, the limit is enforced by having a static number of form fields on /manage/profile as shown in the screenshot - I think it's 26. @momijizukamori mentioned wanting to make that section of the form dynamic in the future, so we can come back and decide what the future limit should be at that point.

htdocs/manage/profile/index.bml
kareila: I did use that slightly further down, in the save_profile_accts method that gets called right after this, but I guess it would make sense to do it here instead.

bin/upgrading/update-db-general.pl
kareila: Yeah, I was following the pattern of the userprop names being fixed forever, but I see your point.

kareila:
If I were to use alloc_user_counter, would that ID be the only primary key, or would it be in combination with the userid?

kareila:
Okay, I think all of the feedback has been addressed. The only thing the `name` field is used for any more is sort order. I also left an index on `value` because that gets ordered.

I noticed when looking at `alloc_user_counter` that G and P were in the regex but unused, and N was used but not in the regex, so I cleaned that up and noted which remaining letters were unused. (Not many.)

zorkian:
Exactly what I was thinking! Thank you for doing this :)

PR #3183, opened 2023-07-07
Convert Homepage to Foundation/TT

momijizukamori:
CODE TOUR: More modernization! This time, the homepage, both logged in and logged out.

* Converts all widgets used on the homepage to TT
* Adds a series of `.smaller` heading element defs so that we can visually reduce headers in a standard way without semantically reducing them
* Takes advantage of the ability to use multiple background layers on a single div to massively clean up the code for the logged out home page.

This one may need to sit on canary for a bit to work out any visual bugs - the only one of these widgets that actually does anything is the quick-reply widget, and I did do some testing on that, but I definitely haven't hit the full range of edge cases for display widgets, heh.


momijizukamori:
I apparently missed that there was already a check in the entry controller to just show the form with content - https://github.com/dreamwidth/dreamwidth/blob/a6990cad48a33b5145b5aad81f2a16e615995813/cgi-bin/DW/Controller/Entry.pm#L197

cgi-bin/DW/Widget/LatestInbox.pm
```diff
@@ -28,24 +28,9 @@ sub render_body {
...
-        . $class->ml('widget.latestinbox.links.compose')
-        . "</a></li>";
-    $ret .=
-          "<li><a href='/manage/settings/?cat=notifications'>"
-        . $class->ml('widget.latestinbox.links.manage')
-        . "</a></li>";
-    $ret .= "</ul></div>";
-
-    $ret .= "<div class='contents'>";
+    my $vars->{limit} = $opts{limit} || 5;
```
kareila, line 33: These two lines can be combined as:
`my $vars = { limit => $opts{limit} || 5 };`

As written, the second `my` is a syntax error.

cgi-bin/DW/Widget/QuickUpdate.pm
```diff
@@ -18,59 +18,36 @@ package DW::Widget::QuickUpdate;
 
 use strict;
 use base qw/ LJ::Widget /;
+use DW::Template;
 
-sub need_res { qw( stc/widgets/quickupdate.css ) }
```
kareila, line 22: This is the only place I see this CSS file included, so you might as well remove it completely.

cgi-bin/DW/Widget/QuickUpdate.pm
```diff
@@ -18,59 +18,36 @@ package DW::Widget::QuickUpdate;
...
+    my @accounts = DW::External::Account->get_external_accounts($remote);
+    @accounts = grep { $_->xpostbydefault } @accounts;
+
+    my @journallist = ( $remote, $remote->posting_access_list );
+    my $sidebar     = LJ::Hooks::run_hook( 'entryforminfo', $remote->user, $remote );
+    my @security    = (
+        ( "public", LJ::Lang::ml("/entry/form.tt.select.security.public.label") ),
+        ( "private", LJ::Lang::ml("/entry/form.tt.select.security.private.label") ),
+        ( "access",  LJ::Lang::ml("/entry/form.tt.select.security.access.label") ),
+    );
```
kareila, line 40: JSYK, a list of lists is syntactically identical to a list that just contains the list items. The extra parens are a little distracting here.

Better would be to use the "arrow comma" syntax, which also lets you use the "key" strings as barewords:
```
    my @security    = (
        public => LJ::Lang::ml("/entry/form.tt.select.security.public.label"),
        private => ...
```

cgi-bin/LJ/Widget.pm
```diff
@@ -146,6 +146,9 @@ sub render {
 
         #        $class->handle_error;
     }
+    else {
+        return;
```
kareila, line 150: I'm curious as to why you added a null return here. Shouldn't you be returning an empty string if you don't expect to see any results, like when `$class->should_render` is false above?

momijizukamori: Hmmm empty string might work there, yeah - I added it because otherwise LJ::Widget passes back a bunch of wrapper HTML with no inner content, which makes Panel wrap it in more wrapper HTML at https://github.com/dreamwidth/dreamwidth/pull/3183/files#diff-73f66f3da82438b1bba42bd02af26110ad3fe2556794b7bb2af891bb7db2928dR113 , and then you get weird display because some of the CSS works on the wrapper classes.

kareila:
I think this is ready to go, it just needs to be rebased on main. I did a force push on my copy of the branch to update it if you want to use that.

One thing I don't think any of us have thought to check on, though, is replacing the stock dw-free index.bml. I'm happy to work on that separately, though.

momijizukamori:
Rebased version merged in  #3237

PR #3184, opened 2023-07-08
convert /register from BML to Foundation

kareila:
CODE TOUR: modernizes the /register (aka /confirm) page.

I think I managed to test every possible success path?

Also:

* pulls /confirm redirect out of Apache/LiveJournal.pm

* activates Foundation on views/success.tt (views/error.tt already had it)


views/register.tt.text
```diff
@@ -67,17 +59,13 @@ Sincerely,
 
 .new.101.title=[[sitename]] 101
 
-.new.bodyuser2=Thanks!  The email address [[address]] for [[user]] has now been verified.  Now that you're registered, what to do first?
+.new.bodyuser=Thanks!  The email address [[address]] for [[user]] has now been verified.  Now that you're registered, what to do first?
```
momijizukamori, line 62: Do strings for TT have a different prefix or something to the old BML strings? I'd be worried about a collision with a previous version otherwise.

kareila: They do! The old one is register.bml.new.bodyuser2 and the new one is register.tt.new.bodyuser. Obviously there used to be a register.bml.new.bodyuser but it was added to deadphrases already.

momijizukamori: Okay cool, that was the only concern I had with this!

PR #3221, opened 2023-07-21
new page /dev/embeds for documenting whitelisted embed domains

kareila:
CODE TOUR: Ever wondered which domains were allowed for embed codes? Wonder no more!

I've been discouraged for years from writing a FAQ that included a list of the whitelisted domains because I knew it would be impossible to keep up to date.

I finally had a brainwave (inspired by the documentation @nfagerlund did for /dev/formats) and now we will have a page we can refer people to which will always have the most current list of supported sites.


rahaeli:
[ this is good ]

PR #3230, opened 2023-07-24
link to full logout page from entry post/edit/delete success links

### LINKED ISSUE #2553: Link to full /logout page from...... somewhere.
Once we make the logout button act like people expect it to again (#2549), it's worth making sure people can still find the /logout page if they actually want to log out of all their sessions. 

Where, tho? Candidates:

- The "You can edit your account settings here. You can also..." intro, on the settings pages. 
- The "Account" settings tab.
- The "Privacy" settings tab.
- The "History" settings tab. (kind of unintuitive, but I can see why it'd make sense to have it near the login sessions link.)
- Next to the logout button in the site header. (too heavyweight, IMO. that's precious real estate.)

kareila:
CODE TOUR: add /logout link to entry success links

I considered Nick's suggestions in #2553 but I didn't like the idea of putting the logout link anywhere near the settings page, because (a) crowded and (b) the Manage Logins page prominently linked there already has logout form buttons. Putting it right next to the sitewide logout button didn't make sense to me, either.

So I thought, when would someone be likely to want to be _prompted_ to log out of the site? And the answer that came to me was... after posting (or editing, or deleting) an entry. So this adds it to the success links for the beta version of the entry form.

Closes #2553, unless someone strenuously objects to my reasoning.

### DIFF
--- a/cgi-bin/DW/Controller/Entry.pm
+++ b/cgi-bin/DW/Controller/Entry.pm
@@ -1219,6 +1219,10 @@ sub _do_post {
             {
                 url       => "$LJ::SITEROOT/editjournal",
                 ml_string => '.links.manageentries',
+            },
+            {
+                url       => "$LJ::SITEROOT/logout",
+                ml_string => '.links.logout',
             }
             );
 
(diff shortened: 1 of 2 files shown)


kareila:
Screenshot 2023-07-23 at 8 25 15 PM

kareila:
I'm going to go ahead and merge this since @rahaeli gave it a thumbs up on Discord.

PR #3233, opened 2023-07-25
convert /manage/invites

kareila:
CODE TOUR: convert /manage/invites to Foundation, plus a few other minor tweaks.


views/manage/invites.tt
```diff
@@ -0,0 +1,134 @@
...
+         END %]</dd>
+    [% END %]
+</dl>
+<br />
+  [% END %]
+
+  [% IF responses.rejected.size %]
+<h2>[% '.head.rejected' | ml %]</h2>
+<p>[% '.body.rejected' | ml %]</p>
+<dl>[% FOREACH comm IN responses.rejected %]
```
momijizukamori, line 57: I know you're just pulling the `<dl>`/`<dd>` from the existing BML template, but I think this isn't really an appropriate use of them, and they can be weird to style -  I think this should probably be `<ul>`/`<li>` instead (possibly with `list-style: none;` CSS)

views/manage/invites.tt
```diff
@@ -0,0 +1,134 @@
...
+[%- END -%]
+
+[% IF responses %]
+  [% IF responses.accepted.size %]
+<h2>[% '.head.accepted' | ml %]</h2>
+<p>[% '.body.accepted' | ml %]</p>
+<dl>[% FOREACH row IN responses.accepted %]
+  <dd>[% row.0.ljuser_display %]:
+      [% FOREACH attr IN row.1;
+           IF loop.count != 1; ", "; END; ".label.$attr" | ml;
```
momijizukamori, line 47: Hmm. This took me a bit to figure out what it was doing, so I'm wondering if it might be better to have a more verbose (but clearer) variant here like
```
[%- attrs  = [] -%]
[% FOREACH attr IN row.1;
attrs.push(dw.ml(".label.$attr");
END %]
[% attrs.join(' ,') %]
```
But I'm willing to be overruled on this one. (if only TT had a `map` function, sigh)

views/manage/invites.tt
momijizukamori: OR ACTUALLY. If you put the comm name inside of `<dt>` tags, and the attributes list in a `<dd>` tag, that would be appropriate use of the tags, semantically speaking. But may change the visual look. (this assumes the attributes are in some way descriptions of the comm)

views/manage/invites.tt
kareila: Heh, if you look at the original BML file, it's concatenating `"item1, item2, item3, "` and then chopping off the last two characters of the string. 🙀

momijizukamori:
Looks good! And the comment is good enough for that second, I think :)

PR #3261, opened 2023-08-02
July cleanup

kareila:
CODE TOUR: no intended impact, just the usual silencing of warnings and such.

### DIFF
--- a/cgi-bin/LJ/Widget/CustomizeTheme.pm
+++ b/cgi-bin/LJ/Widget/CustomizeTheme.pm
@@ -222,7 +222,7 @@ sub render_body {
             propgroup  => "customcss",
             groupprops => $groups{groupprops}->{customcss},
         );
-        $ret .= <<EOF
+        $ret .= <<EOF;
             </div>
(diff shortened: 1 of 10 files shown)


cgi-bin/LJ/Widget/CustomizeTheme.pm
```diff
@@ -222,7 +222,7 @@ sub render_body {
             propgroup  => "customcss",
             groupprops => $groups{groupprops}->{customcss},
         );
-        $ret .= <<EOF
+        $ret .= <<EOF;
```
momijizukamori, line 225: Does this construction actually work or did Tidy do something weird?

alierak: Yes. The only reason omitting the semicolon wasn't a syntax error was that the here doc is followed by a close curly brace.

momijizukamori: Okay cool, I've been jumping between languages so much lately I'm losing track of line termination rules, haha.

PR #3303, opened 2023-11-11
Inbox - read styling, no-JS fix, and imageshrink

momijizukamori:
CODE TOUR: More small tweaks and bugfixes for the beta inbox!

### DIFF
--- a/htdocs/js/jquery.imageshrink.js
+++ b/htdocs/js/jquery.imageshrink.js
@@ -7,14 +7,21 @@
 jQuery(function($) {
     // Entry-like things that might have images in them we want to possibly shrink. Should be in CSS-selector format
     const imgHolders = ['.entry-content', '.comment-content', '.InboxItem_Content .Body'];
-    // containers to observe for changes. Should be a plain id name.
+    // Containers to observe for changes. Should be a plain id name.
     const observedContainers = ['entries', 'comments', 'inbox_message_list'];
 
-    // This just turns imgHolders into a single CSS query string.
-    const imgQuery = imgHolders.join(" img, ") + " img";
+    // The next two constants turn imgHolders into single CSS selector strings.
+    const imgSelector = imgHolders.map(function(holder) {return `${holder} img`}).join(", ");
+
+    // Only shrink casual pics; leave artisanal HTML alone. Can't predict
+    // everything, but 99% of the time that's a <div style="..."> or a table.
+    // Expanded cut tags have style='display: block', but aren't decorations.
+    const exemptSelector = imgHolders.map(function(holderName) {
+            return `${holderName} div[style]:not(.cuttag-open) img, ${holderName} table img`;
+        }).join(", ");
 
     // First: Basic click-to-zoom. (.imageshrink-expanded on/off)
-    $(document).on('click', imgQuery, function(e) {
+    $(document).on('click', imgSelector, function(e) {
         var $that = $(e.target);
         if ( ! $that.is('a img, .poll-response img, .imageshrink-actualsize') ) {
             $that.toggleClass('imageshrink-expanded');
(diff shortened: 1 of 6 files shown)


htdocs/js/jquery.imageshrink.js
```diff
@@ -7,7 +7,7 @@
 jQuery(function($) {
 
     // First: Basic click-to-zoom. (.imageshrink-expanded on/off)
-    $(document).on('click', '.entry-content img, .comment-content img', function(e) {
+    $(document).on('click', '.entry-content img, .comment-content img, .InboxItem_Content .Body img', function(e) {
```
nfagerlund, line 10: Main problem here is that this isn't the only spot in the file where we rely on some selectors to identify which content we're wrangling images within — 

- Under `// Second`, there's the exemption rules in `protectTheCodes()`
- In `observeImages`, there's the zones we're watching for window resize changes
- Under `// Fifth`, there's the zones we're watching for newly-introduced images with the mutation observer

So, you'll need to go through those and make appropriate additions. If you see a good chance to abstract some of this out and enable us to declare top-level selectors for zones up top, that would be the bee's knees, but that might be dicey because of the way we need to exploit knowledge about different entry types and the outer _collection_ containers that can have new content bodies or entry bodies (or inbox item bodies, presumably) spawned into them by ajax updates. (That's the getElementByIds for `#entries` and `#comments` in the far south of here.)

htdocs/js/jquery.imageshrink.js
```diff
@@ -5,9 +5,16 @@
 // - Hiding zoom cursors for images that won't zoom
 // - Marking tall images so they get modified squish behavior
 jQuery(function($) {
+    // Entry-like things that might have images in them we want to possibly shrink. Should be in CSS-selector format
+    const imgHolders = ['.entry-content', '.comment-content', '.InboxItem_Content .Body'];
+    // containers to observe for changes. Should be a plain id name.
+    const observedContainers = ['entries', 'comments', 'inbox_message_list'];
+
+    // This just turns imgHolders into a single CSS query string.
+    const imgQuery = imgHolders.join(" img, ") + " img";
```
nfagerlund, line 14: This is purely 💅 , but could we use the more standard term "selector" in these names instead of "query"?

nfagerlund: Unrelatedly but much more urgently: Is the JS minimizer used by build-static able to cope with `const` and `let` yet, or does it choke on them and demand `var`? I think it's not used in dev, so you'd have to specifically check somehow.

htdocs/js/jquery.imageshrink.js
```diff
@@ -18,19 +25,19 @@ jQuery(function($) {
...
-        // Feeds (.journal-type-Y) always make a mess, so no mercy.
-        var exemptImages = container.querySelectorAll(
-            '.journal-type-P .entry-content div[style]:not(.cuttag-open) img, ' +
-            '.journal-type-P .entry-content table img, ' +
-            '.journal-type-C .entry-content div[style]:not(.cuttag-open) img, ' +
-            '.journal-type-C .entry-content table img, ' +
-            '.comment-content div[style] img, ' +
-            '.comment-content table img'
-        );
+        let queryStrings = imgHolders.map(function(holderName) {
```
nfagerlund, line 28: The selector string this builds is completely static, just like the "imgHolders + img" one above, so I suggest pulling it out of this more active code path and constructing it only once in that outer scope. (And, pre-baking the `.join(', ')` from below into it at the same time, so the pre-computed value is just one string.) 

Also, repeat same notes from above about "let" and calling em "selectors".

htdocs/js/jquery.imageshrink.js
```diff
@@ -18,19 +25,19 @@ jQuery(function($) {
...
+            return `${holderName} div[style]:not(.cuttag-open) img, ${holderName} table img`;
+        });
+
+        var exemptImages = container.querySelectorAll( queryStrings.join(', ') );
         for (var i = 0; i < exemptImages.length; i++) {
             exemptImages[i].classList.add('imageshrink-exempt');
         }
+
+        // Feeds (.journal-type-Y) always make a mess, so no mercy.
+        container.querySelectorAll('.journal-type-Y img').forEach( function(img) { img.classList.remove('imageshrink-exempt')})
```
nfagerlund, line 39: Well.................... okay. 😂 I guess it's better than the other things I might suggest.

nfagerlund (approved):
A nice improvement. 👍🏼

PR #3305, opened 2023-11-14
Entries API

momijizukamori:
CODE TOUR: summarize here.


cgi-bin/LJ/Entry.pm
```diff
@@ -2911,4 +2911,32 @@ sub currents_div {
...
+    $entry->{subject_html} = $self->subject_html();
+    $entry->{body_html}    = $self->event_html(0);
+    $entry->{poster}       = $self->poster()->{user};
+    $entry->{url}          = $self->url();
+    $entry->{security}     = $self->security();
+    $entry->{datetime}     = $self->{eventtime};
+    my @entry_tags = $self->tags();
+    $entry->{tags}     = ( \@entry_tags );
+    $entry->{icon}     = $self->userpic_kw || '';
+    $entry->{entry_id} = delete $self->{ditemid};
```
zorkian, line 2929: why is this deleting the ditemid from $self?

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,568 @@
+#!/usr/bin/perl
+#
+# DW::Controller::API::REST::Entries
+#
+# API controls for entries
+#
+# Authors:
+#      Allen Petersen <allen@suberic.net>
+#
+# Copyright (c) 2016 by Dreamwidth Studios, LLC.
```
zorkian, line 10: is this actually a 2016 file by Allen?

momijizukamori: Hilariously for once it actually is - he wrote the base code for the entry fetch and JSON transform of entry objects.

zorkian:
Do you want me to do anything on this one right now?

cgi-bin/DW/Entry.pm
```diff
@@ -0,0 +1,372 @@
...
+        if ( $security eq "private" ) {
+            $sec = "private";
+        }
+        elsif ( $security eq "access" ) {
+            $sec   = "usemask";
+            $amask = 1;
+        }
+        elsif ( $security eq "custom" ) {
+            $sec = "usemask";
+            foreach my $bit ( $post->get_all("custom_bit") ) {
```
zorkian: I think this trips up the API path: the web form passes a `Hash::MultiValue` (so `get_all` works), but the API hands `_form_to_backend` the decoded JSON body, which is a plain hashref with no `get_all` method — so a custom-security post reaches this line and dies with `Can't call method "get_all" on unblessed reference`. The same thing happens with crosspost down at line 310. In JSON these arrive as arrays, so something like `$is_api ? @{ $post->{custom_bit} || [] } : $post->get_all("custom_bit")` would keep the web path identical while letting the API path through. (public/private/access posts are all fine — it's just these two branches.)

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,353 @@
...
+            dateformat    => 'S2',
+            order         => $journal->is_community ? 'logtime' : '',
+            err           => \$err,
+            posterid      => $journal->is_community && $poster ? $poster->id : undef,
+        );
+
+        my @entries;
+        foreach my $it (@items) {
+            my $item  = LJ::Entry->new( $journal, jitemid => $it->{itemid} );
+            my $entry = json_from_entry( $journal, $item );
```
zorkian: Small thing, but possibly security-relevant: the single-entry path above passes `$remote` to `json_from_entry`, while the list path passes `$journal`. Since that arg drives the `editable_by` check that decides whether to include `text_raw` / `subject_raw` / `allowmask`, passing the journal owner means those raw fields and the allowmask get included for anyone who can see the list (the owner can always "edit" their own entries). I think this just wants to be `$remote`, to match the single-entry path.

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,353 @@
...
+
+    # oops errors when posting: show error, fall through to show form
+    return $self->rest_error( 500, $edit_res->{errors} ) if $edit_res->{errors};
+}
+
+sub _do_edit {
+    my ( $ditemid, $form_req, $auth, %opts ) = @_;
+
+    my $res = DW::Entry::_save_editted_entry( $ditemid, $form_req, $auth );
+    return { { success => 0, errors => { $res->{errors} } } } if $res->{errors};
```
zorkian: I think the nested braces here aren't quite doing what's intended — this builds a hashref-keyed-by-a-hashref rather than `{ success => 0, errors => ... }`, so on an edit failure `$edit_res->{success}` comes back undef and the real error gets dropped before the caller can surface it. The `_do_post` shape just above (`{ success => 0, errors => [...] }`) is probably the one to mirror.

cgi-bin/DW/Entry.pm
```diff
@@ -0,0 +1,372 @@
...
+            $security = "access";
+        }
+        else {
+            $security      = "custom";
+            @custom_groups = grep { $amask & ( 1 << $_ ) } 1 .. 60;
+        }
+    }
+
+    # allow editing of embedded content
+    my $event = $entry->event_raw;
```
zorkian: `$event` is already declared at the top of the sub (line 38), and the legacy-markdown detection mutates that copy to strip the `!markdown` marker. Re-`my`ing it fresh from `event_raw` here means that stripped version gets discarded, so on the web edit path old markdown entries would show the raw `!markdown` marker in the editor again. Dropping the `my` (reusing the existing `$event`) should restore the old behavior. Only affects the web UI, not the API.

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,353 @@
...
+
+        # if user has timezone, use it!
+        if ( $remote && $remote->prop("timezone") ) {
+            my $tz = $remote->prop("timezone");
+            $tz  = $tz ? eval { DateTime::TimeZone->new( name => $tz ); } : undef;
+            $now = eval { DateTime->from_epoch( epoch => time(), time_zone => $tz ); }
+                if $tz;
+        }
+
+        $datetime = $now->strftime("%F %R"),
```
zorkian, line 87: I might be misreading this, but it looks like `$datetime` and `$trust_datetime_value` get computed here and then never make it onto `$post` (or `$form_req`) before `_form_to_backend` runs — and that function only pulls the date out of `$post->{entrytime_date}` / `trust_datetime`. So a client-supplied `datetime` seems to get dropped and the entry falls back to server "now". Did the wiring for this land, or is it still in flight from the datetime fix?

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,353 @@
...
+    $entry->{poster} =
+        { username => $item->poster()->{user}, display_name => $item->poster()->{name} };
+    $entry->{url}      = $item->url();
+    $entry->{security} = $item->security();
+    $entry->{datetime} = $item->{eventtime};
+    my @entry_tags = $item->tags();
+    $entry->{tags}         = ( \@entry_tags );
+    $entry->{icon_keyword} = $item->userpic_kw || '(default)';
+    $entry->{icon}         = $item->userpic;
+    $entry->{entry_id}     = delete $item->{ditemid};
```
zorkian: Echoing the old comment on this one from a while back — the `delete` mutates the cached `LJ::Entry` object by pulling `ditemid` out of it, which can bite later code that expects it to still be there. `$item->{ditemid}` without the `delete` reads it without the side effect. (The `as_json` you added on `LJ::Entry` does it the non-deleting way, fwiw.)

cgi-bin/DW/Controller/API/REST/Entries.pm
```diff
@@ -0,0 +1,353 @@
...
+
+package DW::Controller::API::REST::Entries;
+
+use strict;
+use warnings;
+use DW::Routing;
+use DW::Request;
+use DW::Controller;
+use JSON;
+use Data::Dumper;
```
zorkian: Tiny cleanup: `Data::Dumper` (and a couple of commented-out `print Dumper(...)` lines further down, ~252/268) look like debugging leftovers that can come out. Also looks like there are a few `console.log`s left in `htdocs/js/api.js` too.

zorkian (changes requested):
Thanks for picking this back up! The big structural move here — pulling the form/backend/save helpers out of the Entry controller into `DW::Entry` so the web UI and API can share them — is the right call, and the `$is_api` split reads cleanly.

I focused on the endpoint behavior (per your note above) rather than the known gaps: the rate-limiting question, community edit handling, the datetime-format spec, and the `opt_screening` / "remove before beta" bits all look like reasonable things to defer, so I left those alone.

I did hit a handful of things on the API path worth sorting before this goes in — mostly a couple of spots where the shared helpers assume the web form's data shape and trip over the JSON body, plus one access-control thing on the list endpoint. Details inline. Nothing structural; the bones are good.

momijizukamori:
Okay, got all of those, I think. Can you have a look at the datetime stuff and see if that seems reasonable? I had not actually wired it up all the way previously, lol. 

I will probably take one more pass at any of the TODOs/FIXMEs (though I think one of those was just copied from the beta entry pages controller and thus I have no idea what it means lmao), and then double-check that the currents data comes out in the entry JSON, and that the api handles custom security group stuff properly as input.

momijizukamori:
Okay, I think the datetime stuff should be good, and I fixed the issue with printing entry currents (and cleaned up some duplicate code in the proccess). Still need to do some stuff around community admins, and the output format for locked posts is messy rn.

momijizukamori:
Okay, have
- cleaned up the security display for returned data
- ported over Nick's changes for entry validation
- fixed the URL link on the API docs page to point to the correct settings page for key management
- determined that the admin entry editing is an issue with the new entries page too and we will need a single solution for all that so I'm fine punting it

I *think* that should be all the outstanding bugs.

zorkian:
Thank you! I thiiiink we're good. :D

PR #3317, opened 2023-11-24
Convert /manage/circle/edit to Foundation

momijizukamori:
CODE TOUR: Converts /manage/circle/edit to Foundation and gives it a bit of sprucing-up, so it's nice on mobile and has a real color picker for circle colors.


cgi-bin/DW/Controller/Manage/Circle/Edit.pm
```diff
@@ -0,0 +1,313 @@
+#!/usr/bin/perl
+#
+# DW::Controller::Manage::Circle::PopSubscriptions
```
zorkian, line 3: copy pasta, this isn't the right filename... or the copyright authors below?

cgi-bin/DW/Controller/Manage/Circle/Edit.pm
```diff
@@ -0,0 +1,313 @@
...
+package DW::Controller::Manage::Circle::Edit;
+
+use strict;
+
+use DW::Routing;
+use DW::Template;
+use DW::Controller;
+use LJ::JSON;
+
+my $popsub_path = '/manage/circle/edit';
```
zorkian, line 31: unused variable, remove

cgi-bin/DW/Controller/Manage/Circle/Edit.pm
```diff
@@ -0,0 +1,313 @@
...
+sub edit_handler {
+    my ( $ok, $rv ) = controller();
+    return $rv unless $ok;
+
+    my $r = DW::Request->get;
+
+    my $remote = $rv->{remote};
+
+    my $GET  = $r->get_args;
+    my $POST = $r->post_args;
```
zorkian, line 44: It is not safe to use post_args (or take actions, like below) without checking form_auth (and I can't find a call to check_form_auth.)

anyway, controller does not do this automatically, but you have to tell it to with form_auth => 1.

cgi-bin/DW/Controller/Manage/Circle/Edit.pm
```diff
@@ -0,0 +1,313 @@
...
+    my $vars = {};
+    my $ret  = "";
+
+    my $authas      = $GET->{'authas'} || $remote->user;
```
zorkian, line 48: this is handled by controller, we shouldn't duplicate the logic (it's error prone)... you can pass authas => 1 so controller knows to look for this and then sets the $rv->{u} appropriately (so you can skip asking for remote).

zorkian (changes requested):

PR #3320, opened 2023-11-28
Convert /manage/profile to Foundation/TT

momijizukamori:
CODE TOUR: Conversion of one of the remaining Big Pages, with a handful of additional tweaks to make it work nicely on smaller screens.


cgi-bin/DW/Controller/Manage/Profile.pm
```diff
@@ -0,0 +1,460 @@
...
+#
+# https://github.com/apparentlymart/livejournal/blob/master/LICENSE-LiveJournal.txt
+#
+# In accordance with the original license, this code and all its
+# modifications are provided under the GNU General Public License.
+# A copy of that license can be found in the LICENSE file included as
+# part of this distribution.
+#
+# Authors:
+#      Jen Griffin <kareila@livejournal.com>
```
zorkian, line 19: really? :)

zorkian:
do you need to delete the BML file with this?

cgi-bin/DW/Controller/Manage/Profile.pm
momijizukamori: At least I remembered to change the package name in the header this time?

momijizukamori:
> do you need to delete the BML file with this?

I mean we don't *need* to, but it's satisfying to do so, heh (new routing takes precedence over BML routing)

zorkian:
Me: "November 28th, I thought this was ol..."
Brain: "202**3**, sir."
Me: "...oh, um, right then."

PR #3321, opened 2023-11-29
Convert Customize widgets to TT

momijizukamori:
CODE TOUR: Slowly working our way through re-templating old code....

Note for reviewers - widgets are split fairly cleanly by commit, to help with review.


zorkian:
Looks pretty straightforward, did a very light review by mostly just looking for auth/DB kinds of things. Figure we can eyeball most of it in canary!

PR #3354, opened 2024-04-30
Bug 3337 messaging age barrier

### LINKED ISSUE #3337: add child safety restriction: no PMs between over-18 and under-18 accounts
It's becoming more common for social media sites with a PM/DM function to prevent people whose accounts are flagged as being under 18 from being contacted privately by people whose accounts have a birthdate making them over 18 or people whose accounts don't have a birthdate specified, for child safety purposes. This restriction does have some serious downsides, because there can be some legitimate reasons for that kind of contact (ie, queer/trans teenager living with unsympathetic/unsafe parents who can't have conversations publicly because that unsafe parent monitors their public posts/comments), but the legitimate reasons are outnumbered by the ways it can lead to bad things (and even the legitimate reasons can wind up in the bad place really easily: vulnerable teens can be taken advantage of really easily).

I've been taking a good hard look at our child safety features lately (because we've been taking such a stand in legal cases) and I think it's time to add this one. 

* Someone whose init_bdate makes them over 18 should not be able to private message someone whose init_bdate makes them under 18. (Treat it exactly like PMs are off: grey out the link, and don't provide any clues that it's an age-related restriction, to avoid leaking information about age.)
* Someone whose init_bdate makes them over 18 should not be able to private message someone whose init_bdate makes them under 18. (Again, treat it exactly like PMs are off: grey out the link, and don't provide any clues that it's an age-related restriction, to avoid leaking information about age.)
* Two people whose init_bdate make them both under 18 should still be able to PM each other.
* If someone whose init_bdate used to make them under 18 turns 18, they should not be able to continue replying to any outstanding PM conversation. Ditto someone over 18 who had an outstanding PM conversation with someone under 18 prior to this feature going into place. Error message should be something non-leaky, along the lines of "Error: You cannot currently private message this account."
* This should only run on init_bdate: the birthdate provided on the profile should have nothing to do with it.
* This should not affect things like granting access, subscribing, or commenting to someone, just private messaging. 
* There's probably something else I'm forgetting to add here.

pauamma:
CODE TOUR: Prevent users under 18 from PM'ing those 18 and over, and conversely.

Fixes #3337

(Note: this probably needs some style fixes. I'm rusty on that.)

### DIFF
--- a/cgi-bin/LJ/User/Message.pm
+++ b/cgi-bin/LJ/User/Message.pm
@@ -307,6 +307,14 @@ sub can_receive_message {
     return 0 if $opt_usermsg eq 'M' && !$u->mutually_trusts($sender);
     return 0 if $opt_usermsg eq 'F' && !$u->trusts($sender);
 
+    my $u_age = $u->init_age;
+    my $s_age = $sender->init_age;
+    # init_age returns undef for init_bdate year 0000.
+    return 0 if defined($u_age) && $u_age < 18
+		&& (!defined($s_age) || $s_age >= 18);
+    return 0 if (!defined($u_age) || $u_age >= 18)
+		&& defined($s_age) && $s_age < 18;
+
     return 1;
 }
 
--- a/t/pm-age-barrier.t
+++ b/t/pm-age-barrier.t
@@ -1,0 +1,120 @@
+# t/pm-age-barrier.t
+#
+# Tests the PM age barrier at 18yo.
+#
+# Authors:
+#
+#      Pau Amma <pauamma@dreamwidth.org>
+#
+# Copyright (c) 2024 by Dreamwidth Studios, LLC.
+#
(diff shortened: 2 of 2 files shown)


pauamma:
I'll do a style conformance pass at the end, but a substance review on the current state would be nice.

pauamma:
Assumption check: are you waiting for me to do something before you review, or should I wait patiently until someone has time to look at this?

zorkian:
You don't need to do anything; I am just slow and I guess nobody else has had a chance.

momijizukamori:
At a quick look-over it seems good to me, but because it's a safety feature I'm not quite comfortable signing off on this by myself and would prefer actual staff have a look too.

zorkian:
Thank you for writing excellent tests for this and covering all the edge cases! It looks good to me.

PR #3365, opened 2024-05-22
Fix escaping for edittags page

momijizukamori:
CODE TOUR: TIL the HTML cleaner function in Template Toolkit expects HTML attributes to be double-quoted, and thus does not escape single quotes, leading to chaos if your attribute is single-quoted.


momijizukamori:
Fixes #3365

PR #3373, opened 2024-05-26
Show session messages on S2 pages, too.

momijizukamori:
CODE TOUR: We've had session message flashing (basically little temporary messages like 'Edit successful!' that show after taking an action, and which vanish when you navigate to a new page) for a while now, but when we upgraded the tag editing page, they got added there, too, and we discovered a bug: they didn't show in journal-styled pages, causing them to appear on the next site-styled page you visited instead, which was generally very confusing. This adds those messages to journal-styled pages, currently with very minimal styling.

This will require reloading all the core2 layers in the DB. I'm sorry in advance. I apparently [put styling in](https://github.com/dreamwidth/dreamwidth/blob/eff51877ee43aa1fb983d99ee46849220193831a/htdocs/stc/lj_base.css#L188) when I first implemented this, so messages have a 1px border with the color inherited from whatever the current text color is in the scope. No alert level colors yet but at  the moment the tag update success is the only thing that'll print here.

### DIFF
--- a/cgi-bin/LJ/S2.pm
+++ b/cgi-bin/LJ/S2.pm
@@ -2432,7 +2432,9 @@ sub Page {
         _styleopts            => LJ::viewing_style_opts(%$get),
         timeformat24          => $remote && $remote->use_24hour_time,
         include_meta_viewport => $r->cookie('no_mobile') ? 0 : 1,
+        session_msgs          => $r->msgs
     };
+    $r->clear_msgs;
 
     if ( $opts && $opts->{'saycharset'} ) {
         $p->{'head_content'} .=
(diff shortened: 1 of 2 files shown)


cgi-bin/LJ/S2.pm
```diff
@@ -2432,7 +2432,9 @@ sub Page {
         _styleopts            => LJ::viewing_style_opts(%$get),
         timeformat24          => $remote && $remote->use_24hour_time,
         include_meta_viewport => $r->cookie('no_mobile') ? 0 : 1,
+        session_msgs          => $r->msgs
```
zorkian, line 2435: I'm surprised Tidy didn't add a trailing comma, usually it does I thought

PR #3381, opened 2024-08-05
No longer using Freenode.

pauamma:
Reported by thesecretmaster in IRC.

CODE TOUR: https://www.dreamwidth.org/about wasn't updated when Dreamwidth IRC moved from Freenode to Libera Chat.


zorkian:
Derp, nice catch. Thank you!

PR #3396, opened 2024-09-19
Convert /customize/viewuser to TT

momijizukamori:
CODE TOUR: Another BML page down! Was anyone using this one? Who knows! It was short enough to convert either way


cgi-bin/DW/Controller/Customize.pm
```diff
@@ -0,0 +1,149 @@
...
+
+sub viewuser_handler {
+    my ( $ok, $rv ) = controller( authas => 1 );
+    return $rv unless $ok;
+
+    my $r      = $rv->{r};
+    my $remote = $rv->{remote};
+    my $get    = $r->{get_args};
+
+    my $dbh    = LJ::get_db_writer();
```
zorkian, line 39: wonder why this uses dbh, it seems to only do reads... ancient code is ancient

Issue #3379, opened 2024-06-14
update Gift Paid Time profile link to use new shop url


kareila:
From support: "It looks like when I click to gift a user paid time from their profile, the link still points to dreamwidth.org/shop and winds up just redirecting to the shop itself instead of bringing over the parameters and filling everything out for me."

alierak:
This is coming from cgi-bin/DW/Logic/UserLinkBar.pm, sub buyaccount. It specifies a url starting with "shop/account" and then calls fix_link to turn it into an actual link. fix_link always prepends $LJ::SITEROOT/ if the url doesn't already start with $LJ::SITEROOT.

PR #3397, opened 2024-09-19
Fix profile page link to shop

momijizukamori:
CODE TOUR: We had to move the shop to shop.dreamwidth.org for extremely unexciting compliance reasons, and most of the references to the old dreamwidth.org/shop got updated, but we missed the one on the profile page - it would get you to the shop after a redirect, but it dropped the info about the account you wanted to buy paid time for along the way which was declared 'suboptimal'. This fixes the issue.


momijizukamori:
Fixes #3379

PR #3400, opened 2024-09-25
Convert /polls to TT

momijizukamori:
CODE TOUR: ♫ And another one gone, and another one gone // Another one bites the dust (yeah)♫ The poll creator and it's associated pages have now been converted to TT!


zorkian:
I skimmed but didn't go deep, it seems like a conversion! We can have folks bang on it in canary as per usual, hopefully.

Issue #3405, opened 2024-10-30
Display error in layout chooser after picking new layout


momijizukamori:
The new layout setting seems to be applying properly, so I think this is just a parsing error between the frontend JS and backend RPC endpoint return

![image](https://github.com/user-attachments/assets/61b6289b-4bcc-47b7-a14c-8a84397bbc56)

momijizukamori:
Fixed b y #3407

PR #3407, opened 2024-11-25
Remove debug code causing errors

momijizukamori:
CODE TOUR: It turns out debug print messages do not make valid JSON! Who knew, right? The layout chooser widget now returns valid JSON that does not cause it to spew errors when you select a new layout.


zorkian:
Give it ye olde yeeto burrito!

momijizukamori:
Fixes #3405

PR #3408, opened 2024-11-25
Convert login

momijizukamori:
CODE TOUR: Part of the unending conversion of old pages to the new templates, this time with the side-bonus of cutting down on the number of different codepaths we had to handle logins (which has the bonus of unblocking future work around login options)

-------
Two notes:
1) this has one slightly nasty hack, which is a symlink between `/views/components/login.tt` and `/schemes/login.tt` - this is because we have two different loaders for TT, which are called at different points, one for page contents and one for sitescheme wrappers. They look at different folders, and trying to combine the folders results in some weird circular dependency issues which would be a major effort to untangle. The other option was to make a copy of `login.tt` but that kind of went against the principle of trying to consolidate the code, so symlink it is.

2) there is a small chance this may break some custom controlstrip styling, because the input ids have changed - I think most styles probably just use `input` rather than specific ids, but worth keeping an eye out for in support requests.


momijizukamori:
Draft for the moment because I realized there's some slightly weird behavior on /login directly due to the order in which things get set - but everything else is ready for review.

schemes/login.tt
```diff
@@ -0,0 +1 @@
+/dreamhack/home/8088-momiji/dw/views/components/login.tt
```
zorkian, line 1: The symlink will need to be relative, so it doesn't have your hack path in it.

zorkian:
Aside from the symlink needing to be relative, the rest looks fine. I read through the module code and it seems reasonable, so I think it's good to go once you've figured out the other piece.

schemes/login.tt
momijizukamori: Fixed in c389dee79682360bade65c3d15e9ba2f7e970371

PR #3416, opened 2024-12-03
Sirilyan/add steam to services

### LINKED ISSUE #3387: add Steam to profile/usertag
given a user [username], both profile link and username body can go to https://steamcommunity.com/id/[username] (there's an "all posts by X" option that would probably work better for the username body, but you can't predict the URL from the username alone, alas)

sirilyan:
CODE TOUR: add Steam usernames to profile

fixes #3387


zorkian:
Thank you! Looks right.

PR #3419, opened 2024-12-03
Fix current-theme widget in gradation, and tweak font size

### LINKED ISSUE #3265: "current theme" widget got updated for homepage but is now harder to read on /customize
First reported here: https://dw-maintenance.dreamwidth.org/92733.html?thread=4201277#cmt4201277

Homepage version:
Screenshot 2023-08-04 at 11 08 53 PM

Style select version: 
Screenshot 2023-08-04 at 11 09 07 PM

I believe the primary complaint is that the text on the lower right is smaller than before. The /customize page doesn't have Foundation styling, but perhaps some equivalent CSS could be applied in this one spot to make that text more legible...?

momijizukamori:
CODE TOUR: When we converted the current theme widget to use Template Toolkit, we missed a CSS issue with Gradation that caused to display.... artistically on that site scheme. This also makes the link text on the widget slightly larger, as we got feedback that it was too small.

Fixes #3265

### DIFF
--- a/htdocs/stc/gradation/gradation.css
+++ b/htdocs/stc/gradation/gradation.css
@@ -959,6 +959,11 @@ div#cf-edit select {
     margin-bottom: -2em;
 }
 
+#canvas .appwidget-currenttheme .row {
+    margin: 0;
+    padding: 0;
+}
+
 /** Dark theme for color picker **/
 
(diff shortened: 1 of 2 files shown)


zorkian:
_artistically_

I lol'd.

PR #3421, opened 2024-12-12
Since the DW shop has its own domain, reserve it as a username.

pauamma:
CODE TOUR: The URL for the Dreamwidth shop (where you can buy points, paid time, etc...) now uses "shop.dreamwidth.org", so "shop" can no longer be a username.

### DIFF
--- a/htdocs/inc/reserved-usernames
+++ b/htdocs/inc/reserved-usernames
@@ -1,4 +1,5 @@
 ^www$
+^shop$
 ^dw_
 ^ex_
 ^ext_


pauamma:
While on that topic, there seem to be other missing entries, like for "comments" and "va_*" (or at least "va_mail01"). Is there a complete list of special-purpose hostnames that should be reserved as usernames?

alierak:
This file is a starter list for new sites using the codebase. Remember that include files can be overridden in production through admin/fileedit which stores updated include files in the database. The production list has a few hundred username patterns which are not suitable for github, mostly antispam. I do believe the va_ hostnames were left off, though.

pauamma:
> Remember that include files can be overridden in production through admin/fileedit which stores updated include files in the database.
Uh. I completely forgot that even existed.
> The production list has a few hundred username patterns which are not suitable for github, mostly antispam. I do believe the va_ hostnames were left off, though.
If I'm correctly reading between the lines that "shop" and "comments" are in there too, feel free to close this without action.

alierak:
Right, thanks!

PR #3441, opened 2025-01-13
fix some issues with converted /manage/profile page

### LINKED ISSUE #3440: new Edit Profile page does not load birthday month
I managed to accidentally give myself a January birthday the last time I edited my profile. The saved month isn't being loaded into the form. I can adjust it and save it back as a non-January month, but it reverts to January in the form whenever I try to make another edit.

kareila:
CODE TOUR: When using the new version of the Manage Profile page, I managed to accidentally change my birth month to January, because the previously saved value wasn't being loaded into the form correctly. This fixes that issue plus a couple of others I uncovered during testing.

Fixes #3440.

### DIFF
--- a/cgi-bin/DW/Controller/Manage/Profile.pm
+++ b/cgi-bin/DW/Controller/Manage/Profile.pm
@@ -255,7 +255,7 @@ sub profile_handler {
         }
 
         # bio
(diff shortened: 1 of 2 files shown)


zorkian:
January is a perfectly reasonable month!

Issue #3445, opened 2025-01-17
user-suggested style fixes for /manage/profile in Gradation


kareila:
Summarizing from https://dw-maintenance.dreamwidth.org/96103.html?thread=4340839#cmt4340839

- The user's default profile picture slightly overlaps the border of the white box it's in; it needs to be precisely 2 pixels wider. The icon in question measures 98x62, making me think there's a 100px width assumption being made somewhere.
-  The black arrows on the dropdowns are hard to see on Gradation. (I also noticed this; see screenshot attached to #3442)

momijizukamori:
Turns out the visual bug around the icon is actually present in other site schemes too, it's just harder to spot there - this is on Tropo Red for example:
![Image](https://github.com/user-attachments/assets/9bcf0a5c-bac3-4d4f-99eb-9ce0506d9bd3)

I did also see the dropdown issue - I suspect that any dropdowns on new pages on Gradation have that issue, not just this one, but it needs to be fixed either way, heh

Issue #3443, opened 2025-01-14
remove anchor from URL when submitting update to profile


kareila:
If I click the direct link to edit my profile bio, after saving, the form redirects back to profile#bio, which means I don't see the nice new box at the top of the page with the success links unless I scroll up for some reason. In other words, there's a baked in assumption with showing those links (as opposed to the old interstitial success) that a page always loads with the top in view.

I think it would make sense to remove the anchor tag from the link when submitting the form so that the view snaps back to the top, at least on that one page.

momijizukamori:
writing down the solution to this before it slips out of my mind again - on this line: https://github.com/dreamwidth/dreamwidth/blob/28569d85cb3b045bd3ca06196b22013b3d041a81/views/manage/profile.tt#L28 we need to add an explicit `action` that uses the version of page URL without the fragment. Otherwise it defaults to POSTing to the page url, fragment included, and then we have to do some weird redirect stuff on the backend that I think would result in slower requests.

PR #3450, opened 2025-01-18
Tweaks to profile pages

momijizukamori:
- return to the top of the page after successful form saves
- fix border around userpic
- fix dropdown arrow color on Gradation

CODE TOUR: Some small visual tweaks to the profile editing page, and to forms in Gradation in general!


kareila:
Fixes #3443 
Fixes #3445

kareila:
The style changes look good, but in my testing, edits on the Manage Profile page are no longer being saved - the page just redirects? If I back out that part of the changes, the page saves normally.

Issue #3456, opened 2025-01-24
at least one report of polls no longer working in "older" browsers


kareila:
https://www.dreamwidth.org/support/see_request?id=49713

"Any attempt to answer a poll now does nothing but redirect back to a cleared input form." This is specifically since the code push that deployed the converted poll pages.

According to the user agent string, they are using Firefox 38 on MacOS Catalina, which is pretty old. The same user also said in dw-maint, "Polls don't work on any version of Firefox or Safari I have access to."

momijizukamori:
There's no JS included in the conversion, so presumably it's something to do with form-handling - I'll poke a little, but Firefox 38 is now ten years old.

momijizukamori:
AHAHA I take it back I think this is a genuine bug in all browsers that this person may just be hitting because of lack of JS support or something. The JS endpoint for filling out a poll (used when you're filling it out inside an entry) works fine. The plain HTTP post to `/poll/?id=32575` doesn't record answers, however, and it looks like it's because of our redirect handling for trailing slashes - posting to that returns a redirect to `/poll?id=32575`, and if I manually edit my post request to go to that URL instead, it appears to actually record properly and send me back to the results

PR #3458, opened 2025-01-25
Keep '/poll/' from redirecting to '/poll', which breaks form post

momijizukamori:
CODE TOUR: We have code that makes it so if a user navigates to `dreamwidth.org/somewhere/` , it will redirect them to the 'official' address of `dreamwidth.org/somewhere`. And mostly this works fine, but it *doesn't* work when trying to send a form response to the server, because the 'official' address endpoint never actually sees the data. This resulted in polls failing to work when not submitted via Javascript from within a journal entry, and has now been fixed.


momijizukamori:
Fixes #3456

PR #3490, opened 2025-06-15
Rate Limiting

zorkian:
CODE TOUR: This adds basic rate limiting functionality that will leverage memcached to enable us to keep track of how frequently certain things are happening and then return the appropriate error code. This is primarily being built for the API, but will probably be useful in other contexts as well, or could be? Maybe?

Still WIP, don't merge yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,6 @@ jobs:
 
       - name: Module compilation (t/00-compile.t)
         run: prove -v t/00-compile.t
+
+      - name: Rate limiting (t/rate-limit.t)
+        run: prove -v t/rate-limit.t
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install CPAN dependencies
+        run: cpm install -v --show-build-log-on-failure --no-color --resolver metadb -L /opt/dreamwidth-extlib/ - < doc/dependencies-cpanm
+
       - name: Setup environment
         run: |
           # Config symlink (needed for Perl modules to load site config)
--- a/app.psgi
+++ b/app.psgi
@@ -179,5 +179,8 @@ builder {
     # Middleware for doing sysban blocking (IP bans, uniq bans, tempbans, noanon_ip)
     enable 'DW::Sysban';
 
+    # Rate limiting (after auth and sysban, before request dispatch)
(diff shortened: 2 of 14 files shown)


momijizukamori:
Looks good overall to me! I had a moment of pause as to whether per-endpoint rate limits would be more confusing, but I think given stuff varies in the DB load it causes, and that the results are probably going to be handled programmatically anyway, it's probably fine.

zorkian:
### Code review

Found 2 issues:

1. **API mismatch in Apache/LiveJournal.pm**: The rate limiting integration calls `$limit->exceeded(...)` and `$limit->time_remaining(...)`, but `DW::RateLimit::Limit` only exposes a `check()` method that returns a hashref with `exceeded` and `time_remaining` keys. Additionally, `DW::RateLimit->get()` is called with `max_count`/`per_interval_secs` parameters, but the method requires a `rate` parameter (e.g., `rate => "100/60s"`) and returns `undef` without it. The correct usage pattern is shown in `DW::API::RateLimit->wrap()`. As written, rate limiting in the Apache handler will be silently disabled (`get()` returns undef), and if the constructor is fixed, the missing methods will cause runtime errors.

https://github.com/dreamwidth/dreamwidth/blob/19415ab382e92bb2837c727a35d4af5285aeaff9/cgi-bin/Apache/LiveJournal.pm#L23-L57

Compare with the actual API:

https://github.com/dreamwidth/dreamwidth/blob/19415ab382e92bb2837c727a35d4af5285aeaff9/cgi-bin/DW/RateLimit.pm#L69-L78

https://github.com/dreamwidth/dreamwidth/blob/19415ab382e92bb2837c727a35d4af5285aeaff9/cgi-bin/DW/API/RateLimit.pm#L31-L45

2. **Rate limiting runs before subrequest/OPTIONS check and before `LJ::start_request()`**: The new rate limiting block is inserted before the existing `return DECLINED if defined $apache_r->main || $apache_r->method_number == M_OPTIONS` guard and before `LJ::start_request()`. This means subrequests and CORS preflight requests will be rate-limited (they should be skipped), and `LJ::get_remote()` is called before per-request state is initialized (`LJ::start_request()` calls `LJ::unset_remote()`), potentially returning stale data from a previous request.

https://github.com/dreamwidth/dreamwidth/blob/19415ab382e92bb2837c727a35d4af5285aeaff9/cgi-bin/Apache/LiveJournal.pm#L329-L383

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3514, opened 2026-02-11
Add mysql-client to dependencies-system.

jjbarr:
It turns out that almost every container pulls in mysql-client as an "exclusive dependency". And if everyone's doing it, it's not exclusive, and it should be added to the global dependency list.

CODE TOUR: no-impact. added mysql-client to the list of dependencies needed to install dreamwidth, since it seems that it's required in all cases anyways.


zorkian:
Thank you! :)

PR #3524, opened 2026-02-20
Add support for @mentions and <user> mentions for bluesky.

jjbarr:
Bluesky usernames are FQDNs, and prior to now the @user.site syntax has used the first dot to indicate where the username ends. Since this won't work for services where the username is the domain name, a special case has been introduced: if the final dotted segment in an @mention is shorthand for an atproto-based site, everything else is assumed to be a username. This does introduce an irregularity: @user.example.com.bsky works to reference the bluesky user @user.example.com, but @user.example.com.bsky.app does NOT. This is done to prevent the mention as a whole from being a valid domain name (and because future atproto sites may eventually allow usernames that make such constructs ambiguous).

In addition to "bsky.app" and the "bsky" shorthand being added, "bsky.social" has been added to make referencing users with the default domain bluesky uses easier. An "Atproto" site,
`DW::External::Site::Atproto`, as also been added as a place to store common code. It is complete and links to an "account overview" page listing services associated with the account, but has not yet been exposed to users.

At present, bluesky accounts are linked to by username, not be DID. This may be subject to future change.

CODE TOUR: It's now possible to reference bluesky accounts with `@mentions`! `@username.bsky.social` works the way you would expect. If a user has a custom username, you can suffix it with `.bsky` to make the link work (eg, `@user.example.com.bsky`). `<user>` tags also work as you would expect (`<user name=user.bsky.social site=bsky.app>`, `<user name=user site=bsky.social>`, and `<user name=user.example.com site=bsky.app>` all do what you expect). However, `@user.example.com.bsky.app` doesn't work, as allowing this could create ambiguity in the future as more sites are added.

### DIFF
--- a/cgi-bin/DW/External/Site/Atproto.pm
+++ b/cgi-bin/DW/External/Site/Atproto.pm
@@ -1,6 +1,6 @@
 #!/usr/bin/perl
 #
-# DW::External::Site::Bluesky
+# DW::External::Site::Atproto
 #
 # Support class for elements shared between atproto-based sites. Links to
 # aturi.to to provide a whole-account overview for atproto..
@@ -46,7 +46,7 @@ sub journal_url {
     # sending the user to aturi is the correct way to handle that since it
     # presents the well-known sites the account DOES have a profile on in and
     # end-user friendly way.
-    return 'http://aturi.to/' . $u->user;
+    return 'https://aturi.to/' . $u->user;
 }
 
 # argument: DW::External::User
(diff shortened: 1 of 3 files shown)


cgi-bin/DW/External/Site/Atproto.pm
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/perl
+#
+# DW::External::Site::Bluesky
```
zorkian: nit: this is Site::Atproto

zorkian (changes requested):
Great PR! Mostly nits, the two real requests are to use `https` instead of `http`. If you want to push an update, I'll merge it in! :)

zorkian:
Perfect, thank you!

PR #3572, opened 2026-06-08
Convert /customize/preview_redirect from BML to Template Toolkit

zorkian:
Converts the theme-preview redirect to the modern architecture. The handler moves into `DW::Controller::Customize` (the existing hub for miscellaneous customization routes, alongside `/customize/viewuser`) as `preview_redirect_handler`: it resolves a `themeid`/`layoutid` to a preview style for the viewer's own journal and redirects there with `?s2id=`. The three error cases (`error.user`/`error.id`/`error.preview`) use `error_ml`. It's a controller-only page (no view), so only the `.bml.text` strings move to `views/customize/preview_redirect.tt.text` and the old keys are deadphrased. The two callers (the theme chooser and the admin themes list) link to the URL with GET params, so nothing needed repointing.

Verified live under Starman: anonymous (invalid account), logged-in with no params (no theme id), and a real themeid (303 redirect to the journal with the preview style).

CODE TOUR: Clicking "preview" on a journal theme bounces you through a small redirect page that figures out the right preview style and sends you to your journal to see it. This moves that page off the old BML format; the preview links work exactly as before.


KainosSLS:
Ok.

PR #3573, opened 2026-06-08
Convert /support/append_request from BML to Template Toolkit

zorkian:
Converts the support reply/action processor. `htdocs/support/append_request.bml` becomes `DW::Controller::Support::AppendRequest` + `views/support/append_request.tt`. It's the POST target behind the (already-TT) `see_request` reply form and the `act.tt` reopen form: it appends a user-facing and/or internal response and performs the requested staff actions — touch/untouch the queue, change category, approve a screened response, change the summary, or bounce to email and close — then shows a confirmation with navigation links. The many input-validation branches become `error_ml` (and `bml.requirepost` for the POST guard); the page keeps the support auth-token model (`anonymous => 1`, no `form_auth`) because requesters reach it via emailed auth links, not a session. The `.bml.text` strings move to `views/support/append_request.tt.text` and the old keys are deadphrased.

Verified live under Starman: unknown request, closed request, and a real answer POST (confirmation page with the nav links, and the answer persisted to supportlog).

CODE TOUR: When a support volunteer (or a user replying to their own request) submits the reply form, this is the page that actually records the response and runs any action — closing, reopening, moving categories, and so on. It moves that processor off the old BML format onto the modern controller/template split, with no change to the forms that feed it.


KainosSLS:
Ok.

PR #3574, opened 2026-06-08
Convert /manage/tags from BML to Template Toolkit

zorkian:
Converts `/manage/tags` from BML to the `DW::Controller` + Template Toolkit architecture. This is a JS-driven page (`js/tags.js`) that manages the tags in a journal you own or maintain.

**Controller (`cgi-bin/DW/Controller/Manage/Tags.pm`):**
- Registers `/manage/tags` with `no_cache => 1`; uses `controller( anonymous => 0, authas => 1, form_auth => 1 )` so a community's admin can manage its tags via `?authas=`.
- Dispatches the page's several POST actions by submit-button name — add (the image submit sends `add.x`/`add.y`), `delete`, `rename`, `merge`, `show posts` (redirects to the journal's `/tag/...`), and `save_levels` — collecting any failures in a `DW::FormErrors` object.
- Builds the tag list, the per-tag usage histogram levels (for the cell-bar icons), and the permission-level `<select>` option lists, and passes them to the template.

**Template (`views/manage/tags.tt`):**
- Reproduces the markup; every element id/class that `js/tags.js` hooks is preserved.
- Injects the `ml.*` label object (via `sections.head`) and the per-tag `var tags` JS array that `js/tags.js` reads, and sets `onLoad="initTagPage()"` via `sections.bodyopts`.
- Errors render through `components/errors.tt`; the rename/merge/delete confirm dialogs are JS-escaped (`LJ::ejs`, precomputed in the controller since TT's `|` filter only works at a directive's top level, not inside `form.*` arg expressions).

**Other:** resources (`js/tags.js`, `stc/tags.css`) load in the controller so the CSS can use `$LJ::OLD_RES_PRIORITY`. The "tags disabled" message is now a self-contained `.disabled` key instead of borrowing `/edittags.bml.disabled` (edittags is already on TT). Old `/manage/tags.bml.*` keys are retired in `deadphrases.dat`.

Verified live under Starman as a logged-in user with several tags: full GET render (authas switcher, tag `<select>` with usage-level option classes, the `ml.*` labels, the `var tags` array, the permission selects), and every POST action — add, rename, delete, merge, save-levels, and the show-posts 303 redirect — plus the comma-in-rename error bar. `t/02-tidy.t` (993) and `t/00-compile.t` (1570) pass.

CODE TOUR: The "manage tags" page (where you rename, merge, delete, and set permissions on your journal's tags) was written in BML, the old format that mixes Perl and HTML in one file. This rebuilds it the modern way — logic in one file, layout in another — with the page working exactly as before for both personal journals and communities.


KainosSLS:
Ok.

PR #3576, opened 2026-06-08
Convert /support/help from BML to Template Toolkit

zorkian:
Converts the support board — the last BML page in the support cluster. `htdocs/support/help.bml` becomes `DW::Controller::Support::Help` + `views/support/help.tt`. The controller handles the state (open/closed/green/you-replied) and category filtering, sorting, the abstracts mode, status classification + point calculation, and builds the per-request display rows and the filter dropdowns; the template renders the table, the filter form, and — for users who can close requests in the selected category — the mass close/move/close-with-points form. That form posts to the already-migrated `/support/actmulti` and emits exactly the fields it consumes (`check_<spid>`, `ids`, `spcatid`, `ret` with its `%s` slot, `changecat`, the `action:*` buttons, and `form_auth`). The page-local row CSS and the row-toggle JS move into the head section. The `.bml.text` strings move to `views/support/help.tt.text` (with one value un-wrapped from embedded BML `<?p?>` tags) and the old keys are deadphrased.

Verified live under Starman: the anonymous board (heading, counts, headers, no leftover BML tags), and the logged-in staff view with the mass-action form, the per-request checkbox rows, the `doClick` JS hooks, and the hidden `ids`/`spcatid`/`ret` fields.

CODE TOUR: The support board — the big list of open requests that volunteers work from, with the filters and the bulk "close these" controls — was the last piece of the support area still on the old BML page format. This moves it onto the modern templates, looking and working the same, including the checkboxes and bulk-close buttons.


KainosSLS:
Ok.

PR #3593, opened 2026-06-20
Fix blank page when saving /manage/profile

zorkian:
Saving `/manage/profile` returned a **blank page** and logged `Argument "<?errorbar <strong>There was an error processing your re..." isn't numeric in numeric eq (==) at app.psgi line 80` whenever a setting failed to save — e.g. leaving find-by-email at its default on a dev server (where `opt_findbyemail` is enabled).

**Root cause** (introduced in the original BML→TT conversion, `122355df5` / #3320, Dec 2024 — `git blame` puts both error returns on that commit): the controller returned `LJ::error_list(...)` / `bad_input(...)` on error, which produce a BML `<?errorbar?>` **string**. Under DW::Routing a handler must return a numeric status / Plack response, so the string fell through to `app.psgi:80` (`$ret == 0`) → the "isn't numeric" warning + a blank page. The page's error handling was broken several ways at once:

- the save gate checked an always-empty `@errors` array instead of the `$errors` `DW::FormErrors` object, so field validation never actually blocked the save;
- `views/manage/profile.tt` never rendered `$errors`;
- an operator-precedence bug (`!$x =~ /re/` parses as `(!$x) =~ /re/`) made the findbyemail check a no-op;
- post-save interest-validation errors still reported success.

**Fix:**
- Render errors through `DW::FormErrors` + `components/errors.tt` and re-render the form (no save) when `$errors->exist`.
- Route `LJ::Setting->save_all` errors into `$errors` instead of returning a string.
- Fix the gate (`@errors` → `$errors->exist`) and the findbyemail precedence bug.
- Guard the success redirect so interest-validation errors re-render instead of falsely reporting success.
- Silence a pre-existing `"" + 0` numeric warning in `profile.tt` (empty birth month).

**Verified** in the devcontainer (as the seeded `test_user`): invalid input now re-renders the form with the error shown (HTTP 200, no blank page); a valid save redirects (303 → profile); and there are **no more "isn't numeric" warnings**. `t/02-tidy.t` passes.

CODE TOUR: Editing your profile and hitting save would sometimes just show a blank page instead of telling you what went wrong. This fixes it so the page properly shows the error message and lets you correct it, and so a valid save goes through cleanly.

Fixes the blank-page-on-profile-save report.

### DIFF
--- a/cgi-bin/DW/Controller/Manage/Profile.pm
+++ b/cgi-bin/DW/Controller/Manage/Profile.pm
@@ -262,13 +262,26 @@ sub profile_handler {
             $errors->add( 'bio', "$scope.error.bio.toolong" );
         }
 
-        # FIXME: validation AND POSTING are handled by widgets' handle_post() methods
-        # (introduce validate_post() ?)
+        # Field validation is done; bail before any saving if it failed. The
+        # Location widget below both validates AND persists immediately, so
+        # running it after a validation error would partially save.
+        if ( $errors->exist ) {
+            $vars->{errors} = $errors;
+            return DW::Template->render_template( 'manage/profile.tt', $vars );
+        }
+
+        # FIXME: validation AND POSTING are handled by widgets' handle_post()
+        # methods (introduce validate_post()?). The Location widget's own errors
+        # accumulate in @BMLCodeBlock::errors rather than being returned, so
+        # capture them and surface them through $errors.
+        @BMLCodeBlock::errors = ();
         my $save_search_index = $POST->{'opt_showlocation'} =~ /^[YR]$/;
         LJ::Widget->handle_post( $POST, 'Location' => { save_search_index => $save_search_index } );
-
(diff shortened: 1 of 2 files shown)


zorkian:
Thanks — both addressed in the latest commit:

- **Partial save on validation error (Profile.pm:274):** moved the \`$errors->exist\` gate *above* \`LJ::Widget->handle_post\`, so a submission with field errors now re-renders before the Location widget (which validates-and-persists immediately) runs — no partial location/timezone save. Also capture the widget's own errors from \`@BMLCodeBlock::errors\` (it returns results, not errors) and surface them through \`DW::FormErrors\`.
- **Single arbitrary save error (Profile.pm:322):** now iterates every field in each setting's \`save_errors\` and adds them all, instead of picking one via non-deterministic hash order.

Verified in the devcontainer: missing-name POST re-renders with the error and does not save; a valid save still redirects (303); no "isn't numeric" warnings.

PR #3606, opened 2026-06-22
Add syn_delete console command to delete/undelete feed accounts

zorkian:
Adds two console commands (priv: `syn_edit`) for managing syndicated (RSS/Atom feed) accounts: `LJ::Console::Command::SynDelete` (`syn_delete <feed>`) and `LJ::Console::Command::SynUndelete` (`syn_undelete <feed>`). Previously the only options were `syn_merge` (which requires a merge target) or a direct DB purge. Delete calls `set_deleted` (statusvis `D`); undelete calls `set_visible` and resets `checknext`/`failcount` so checking resumes promptly. No scheduler change is needed because `bin/worker/schedule-synsuck` already only picks up feeds where `statusvis='V'`, so a deleted feed stops refreshing and then gets purged through the normal expunge lifecycle. Test coverage in `t/console-syndelete.t`.

CODE TOUR: Site admins can now cleanly retire an RSS/Atom feed account from the admin console with `syn_delete <feed>`. This marks the feed as deleted so the site stops fetching new posts for it, and it eventually gets cleaned up like any other deleted account. If it was deleted by mistake, `syn_undelete <feed>` brings it back and resumes feed checking.

🌒 Run on Niteshift: [View Task](https://niteshift.dev/repo/dreamwidth/dreamwidth/task_j9kgeatgpkfn04yd)

### DIFF
--- a/cgi-bin/LJ/Console/Command/SynDelete.pm
+++ b/cgi-bin/LJ/Console/Command/SynDelete.pm
@@ -1,0 +1,96 @@
+# This code was forked from the LiveJournal project owned and operated
+# by Live Journal, Inc. The code has been modified and expanded by
+# Dreamwidth Studios, LLC. These files were originally licensed under
+# the terms of the license supplied by Live Journal, Inc, which can
+# currently be found at:
+#
+# http://code.livejournal.org/trac/livejournal/browser/trunk/LICENSE-LiveJournal.txt
+#
+# In accordance with the original license, this code and all its
+# modifications are provided under the GNU General Public License.
+# A copy of that license can be found in the LICENSE file included as
+# part of this distribution.
+
+package LJ::Console::Command::SynDelete;
+
+use strict;
+use base qw(LJ::Console::Command);
+use Carp qw(croak);
+
+sub cmd { "syn_delete" }
+
+sub desc {
+    "Deletes a syndicated (RSS/Atom feed) account, marking it for purging and stopping the "
+        . "syndication system from refreshing it. Pass 'undelete' to reverse this and resume "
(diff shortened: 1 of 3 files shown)


cgi-bin/LJ/Console/Command/SynDelete.pm
```diff
@@ -0,0 +1,96 @@
+# This code was forked from the LiveJournal project owned and operated
```
zorkian: this code was not forked, we wrote it new - fix the header

cgi-bin/LJ/Console/Command/SynDelete.pm
```diff
@@ -0,0 +1,96 @@
...
+
+use strict;
+use base qw(LJ::Console::Command);
+use Carp qw(croak);
+
+sub cmd { "syn_delete" }
+
+sub desc {
+    "Deletes a syndicated (RSS/Atom feed) account, marking it for purging and stopping the "
+        . "syndication system from refreshing it. Pass 'undelete' to reverse this and resume "
```
zorkian: I don't know that I love 'syn_delete feed undelete' as the way to undelete.. are you mirroring another admin command or trying to make this only a single command?

should this have been part of syn_edit itself? syn_edit feed "delete or undelete or new URL"? I don't remember how that command works

t/console-syndelete.t
```diff
@@ -0,0 +1,113 @@
+# t/console-syndelete.t
+#
+# Test LJ::Console syn_delete command (delete and undelete of feed accounts).
+#
+# This code was forked from the LiveJournal project owned and operated
```
zorkian: incorrect header

PR #3607, opened 2026-06-22
Fix logged-out comment expansion: exempt AJAX from the captcha view-bounce

zorkian:
Logged-out readers couldn't expand comment threads ("Expand" links and "Expand all"). Root cause: `DW::Captcha->should_captcha_view` — wired into `DW::Controller::Journal::render` (the Plack/web22 entry-view path) by #3553/`3162df7c0` — redirects anonymous visitors to `/captcha` once their per-IP/uniq token bucket is spent. The expander's AJAX is a plain GET to that same entry-view path (`…/NNN.html?thread=…` / `?expand_all=1`), so it gets a `303 → /captcha` as well. jQuery's XHR silently follows the redirect and hands `jquery.threadexpander.js` the captcha page instead of the comment HTML, so expansion does nothing; the burst of redirected XHRs can also trip the fraud counter and tempban the reader's IP. Logged-in users were unaffected because `should_captcha_view` returns early for any `$remote`.

Fix: skip the captcha bounce for XHR requests (`X-Requested-With: XMLHttpRequest`, which jQuery sets on same-origin AJAX). These are sub-requests of a page whose initial load already cleared the captcha gate, so re-gating them only breaks interactive features. Full-page navigation stays gated.

Verified in the devcontainer with captcha enabled: anonymous page navigation still `303`s to `/captcha`, while the expand AJAX now returns `200` with the comment HTML; confirmed via a server-side header probe that the jQuery expander actually sends `X-Requested-With`.

CODE TOUR: If you weren't logged in, clicking "Expand" on a collapsed comment thread (or "Expand all") did nothing — the site was quietly redirecting those background requests to the "are you a robot?" page instead of sending back the comments, and could even get your connection temporarily blocked. Now those background requests skip the robot check (the page you're already reading already passed it), so expanding comments works whether or not you're logged in.

🌒 Run on Niteshift: [View Task](https://niteshift.dev/repo/dreamwidth/dreamwidth/task_7e2jfogakbayp5pu)

### DIFF
--- a/cgi-bin/DW/Captcha.pm
+++ b/cgi-bin/DW/Captcha.pm
@@ -267,6 +267,17 @@ sub should_captcha_view {
     # If the user is logged in, no captcha
     return 0 if $remote;
 
+    # Never captcha-bounce an AJAX request. These (e.g. comment thread expansion
+    # from jquery.threadexpander.js, ajax cut-tag expansion) are sub-requests of
+    # a page the visitor already loaded -- and that initial page load is what the
+    # captcha gate is meant to stop. An interstitial redirect can't be followed
+    # by an XHR the way a browser navigation can: jQuery silently follows the
+    # 30x, hands the JS the captcha page instead of the expected HTML fragment,
+    # and the feature just breaks. Worse, the stream of redirected XHRs looks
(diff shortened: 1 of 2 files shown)


zorkian:
wait but why

PR #3617, opened 2026-06-26
Occupancy-based autoscaling for the web tier

zorkian:
Scales the web tier on Starman worker-pool occupancy instead of CPU/request-count, which both move the *wrong* way when workers are blocked on I/O. Each request records busy/idle milliseconds into sharded memcached counters (`DW::WorkerOccupancy`, wired in `RequestWrapper`). A new singleton worker `bin/worker/web-autoscaler` reads those counters each tick, computes occupancy `busy/(busy+idle)` via the pure `DW::Autoscaler::{Counters,Window,Decision}` modules (per-shard reset detection, trailing-window smoothing, proportional target with asymmetric hysteresis/step-caps/cooldowns), and sets ECS `desiredCount` via Paws unless `dry_run`. Every decision is emitted via `DW::Stats`. It ships as a normal worker: a `config/workers.json` entry (singleton, on-demand) drives the generated firelens→Loki task def and the deploy workflow; `update-workflows.py` gains additive `task_role`/`environment` overrides so it can carry its dedicated least-privilege IAM role. Tuning knobs live in an EFS `autoscaler.json` re-read each tick, so target/windows/caps and dry-run are adjustable live. Rollout is observe (dry-run) → shadow → enforce, starting with `web-unauthenticated`.

CODE TOUR: When the site gets busy, the web servers can stall waiting on the database while their CPU still looks idle — so the old autoscaling sometimes added too few servers, or even removed some, at the worst moment. This teaches the system to measure how *busy the request workers actually are* and add or remove servers based on that. It rolls out carefully in watch-only mode first, and a matching infra change (the IAM role and worker service) goes in the dw-terraform repo.


bin/worker/web-autoscaler
```diff
@@ -0,0 +1,183 @@
...
+#
+# Occupancy-based autoscaler for the web tier. Each tick: read sharded
+# busy/idle counters from memcached, compute smoothed pool occupancy per
+# service, decide a desiredCount within configured limits, and (unless dry_run)
+# apply it via the ECS API. Emits everything it sees via DW::Stats.
+#
+# Run as a singleton ECS service (desiredCount = 1). Config is read from
+# DW_AUTOSCALER_CONFIG every tick, so it can be retuned live via EFS.
+#
+# This program is free software; you may redistribute it and/or modify it under
```
zorkian, line 19: Fix the header to include our standard copyright language.

bin/worker/web-autoscaler
```diff
@@ -0,0 +1,183 @@
...
+use JSON ();
+use Time::HiRes ();
+use Paws;
+use DW::Stats;
+use DW::Autoscaler::Counters;
+use DW::Autoscaler::Window;
+use DW::Autoscaler::Decision;
+
+my $CONFIG_PATH = $ENV{DW_AUTOSCALER_CONFIG} || "$ENV{LJHOME}/etc/autoscaler.json";
+my $CLUSTER     = $ENV{DW_ECS_CLUSTER}       || 'dreamwidth';
```
zorkian: Don't hardcode 'dreamwidth', it won't be right for other sites.

bin/worker/web-autoscaler
```diff
@@ -0,0 +1,183 @@
...
+my $REGION      = $ENV{AWS_REGION}           || 'us-east-1';
```
zorkian: Don't hardcode/default region. Force using configs.

bin/worker/web-autoscaler
zorkian: Don't use ENV configs... we use $LJ::FOO configs which go in config-private-prod.pl for things like this.

bin/worker/web-autoscaler
```diff
@@ -0,0 +1,183 @@
...
+
+use JSON ();
+use Time::HiRes ();
+use Paws;
+use DW::Stats;
+use DW::Autoscaler::Counters;
+use DW::Autoscaler::Window;
+use DW::Autoscaler::Decision;
+
+my $CONFIG_PATH = $ENV{DW_AUTOSCALER_CONFIG} || "$ENV{LJHOME}/etc/autoscaler.json";
```
zorkian: This is not how we do configs. We put them in etc/config-private-prod.pl for something like this, and they are Perl objects/etc, don't use JSON.

bin/worker/web-autoscaler
```diff
@@ -0,0 +1,183 @@
...
+    local $/;
+    my $raw = <$fh>;
+    close $fh;
+    my $cfg = eval { JSON::decode_json($raw) };
+    if ( $@ || ref $cfg ne 'HASH' ) {
+
+        # Distinguish a corrupt/bad config push (during live tuning) from an
+        # unreadable file -- both hold, but a parse error must be visible.
+        my $why = $@ || 'not a JSON object';
+        warn "autoscaler: config parse error in $CONFIG_PATH: $why\n";
```
zorkian: General note for logs, look for the $log pattern we use elsewhere using Log4perl, that way we get structured control over logs and log levels... don't just use warn.

cgi-bin/DW/Autoscaler/Counters.pm
```diff
@@ -0,0 +1,40 @@
...
+#
+# DW::Autoscaler::Counters
+#
+# Pure helpers for turning sharded memcached counter snapshots into per-tick
+# deltas, with per-shard reset detection, and into an occupancy ratio.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself.
+
+package DW::Autoscaler::Counters;
```
zorkian: This is like two functions... we should simplify; instead of having 5 files with 2 functions each, having 1 well laid out file is better.

PR #3620, opened 2026-06-27
Make jbackup.pl resilient to throttling and interruptions

afrozenpeach:
CODE TOUR: Allow backing up of larger communities, while being more resilient to being throttled and allowing the user to continue from where they left off if throttled.

Full disclosure. This was mostly written by Claude as I walked through trying to get a backup made, but is tested and functional. I ran into a lot of troubles trying to backup an extremely large community, and this made the job much easier.

# Make `jbackup.pl` resilient to throttling and interruptions on large journals

## Summary

Hardens `src/jbackup/jbackup.pl` for large communities (tens of thousands of
entries, hundreds of thousands of comments). Adds request timeouts, exponential
backoff, polite rate-limiting, adaptive pacing, incremental comment saving, a
tunable comment batch size, durable accounting of skipped windows with an
end-of-run report, a `--backfill` mode to recover them, and safe handling of
metadata-window failures. Together these let a long backup survive server-side
throttling and interruptions without hanging, escalating into a harder block,
silently dropping data, or discarding hours of downloaded comments.

All new behavior is on by default with conservative values; every knob is
optional and overridable via `~/.jbackup`.

## Motivation

On a community with ~31k entries and 300k+ comments, the current script has
several failure modes that compound:

1. **No request timeout.** `do_authed_fetch` and the XML-RPC transport have no
   timeout, so a stalled connection (a throttle that drops the socket rather than
   returning a status) hangs the process indefinitely.
2. **No backoff.** On a timeout, `xmlrpc_call_helper` retries *immediately* and
   unbounded; on an HTTP error, `do_authed_fetch` returns undef and the caller
   `die`s. Neither waits. Sustained immediate retries against a throttling server
   escalate a soft rate-limit into a hard `Denied access to method` block.
3. **No rate-limiting.** There's no delay between calls, and the XML-RPC path
   re-fetches a challenge before every call (two round trips each), so a large
   sync is a high sustained request rate.
4. **All-or-nothing comment save.** Comment bodies accumulate in `%meta` and are
   only written to the database (`save_comment`) after the *entire* body loop
   completes; `comment:lastid` advances only at the end. Any failure during the
   comment phase discards every downloaded body and leaves the checkpoint
   unmoved, forcing a full re-download on the next run.
5. **A skipped window is silent and unrecoverable.** Even once skipping is
   possible, a skip is a transient stderr line with no durable record and no way
   to recover just the gap.
6. **A metadata-window failure is catastrophic, not bounded.** The body fetcher
   can tolerate a skipped window (a localized gap), but the metadata pass cannot:
   a skipped `comment_meta` window leaves `%meta` incomplete and `$server_max_id`
   wrong, which corrupts the saved state of comments in that range and can
   silently truncate the body sweep.

### Tunable knobs (overridable via `~/.jbackup`, e.g. `xmlrpc_delay=3`)

| Knob | Default | Purpose |
|------|---------|---------|
| `xmlrpc_delay` | 2s | delay before each XML-RPC call |
| `fetch_delay` | 1s | delay before each comment HTTP fetch |
| `http_timeout` | 120s | per-request timeout (a stall becomes an error) |
| `max_tries` | 6 | backoff retries for body/XML-RPC before skip/abort |
| `meta_max_tries` | 9 | backoff retries for metadata windows (higher stakes) |
| `max_backoff` | 300s | cap on a single backoff sleep |
| `comment_body_fetch` | 1000 | comment bodies per request (lower to dodge 504s) |
| `adapt_step` / `adapt_max` / `adapt_decay` | 5s / 60s / 10 | adaptive pacing |

### Request timeouts, backoff, and rate-limiting

`do_authed_fetch` (comment HTTP fetcher):

- Sets `$ua->timeout($http_timeout)` so a stalled socket becomes a catchable
  error instead of a hang.
- Sleeps `fetch_delay` (plus the adaptive delay, below) before each request.
- On HTTP error, retries with exponential backoff (15·2ⁿ s, capped at
  `max_backoff`) up to the relevant try cap, then records and skips the window
  (see *Skip accounting*) rather than aborting the whole run.

`xmlrpc_call_helper` / `call_xmlrpc` (XML-RPC path):

- Sleeps `xmlrpc_delay` before each call; sets the transport timeout.
- Threads a `$tries` counter through both subs.
- On a fault (previously an immediate abort), backs off (30·2ⁿ s) and retries up
  to `max_tries` before aborting, with a clearer message that a persistent fault
  likely means a rate/abuse block.
- On no response (previously an immediate, unbounded retry), uses bounded
  exponential backoff.
- The `getchallenge` helper call now receives `$mode`/`$hash`, so a retry after a
  challenge failure restarts the intended call correctly rather than calling
  `call_xmlrpc(undef, undef)`.

### Adaptive pacing (comment fetch)

The per-window backoff resets its try-counter for each new window, so after
surviving a rough patch the fetcher returns to full speed and re-discovers the
throttle from scratch on the next window. On a server returning sustained `504`s
(a slow backend rather than a hard rate-limit) this wastes time and keeps poking
a struggling server. `do_authed_fetch` therefore maintains a small amount of
state that *carries across windows*:

- A persistent extra delay is added to every request, on top of `fetch_delay`.
- On any HTTP error it ramps up multiplicatively (`adapt_step`, then doubling,
  capped at `adapt_max`) and the clean-streak counter resets.
- On success it decays — but only after `adapt_decay` consecutive clean windows,
  and only by halving — so the slowdown is *sustained* through and after a rough
  patch rather than snapping back instantly.

With the defaults, a burst of errors ramps the carried-over delay 5→10→20→40s
and eases off over ~30–40 clean windows. It is self-tuning, bounded, and adds no
delay once a backup is running cleanly.

### Incremental comment saving

- The body loop now calls `save_comment` for the current window's comments,
  advances `comment:lastid`, and `$tied->sync()`s the database **every window**,
  instead of saving everything after the loop.
- A small `@window_ids` list (populated by `body_handler`) lets each iteration
  save only that window.
- **Result:** the comment phase is resumable. An interruption keeps everything
  already pulled, and the next `--sync` resumes from `comment:lastid` instead of
  re-downloading bodies from the start.

The lighter `comment_meta` pass still re-runs from scratch each comment phase, as
before, since it's held in memory (~10k IDs/request, not the bottleneck).

### Tunable comment batch size

`comment_body` requests default to the server maximum of 1000, but a heavy window
(a large thread) can exceed the server's gateway timeout and `504`. The batch
size is now read from `comment_body_fetch` (clamped 1–1000), so it can be lowered
(e.g. 250) to make each request lighter and far less likely to time out, at the
cost of more requests. The checkpoint is in absolute comment-ID units, so the
value can be changed between runs without re-downloading anything.

### Skip accounting and `--backfill`

A skipped window is now recorded durably in the database under a single
`skipped:windows` key as a comma list of `mode:startid:numitems`, so a gap
survives the run ending, carries its exact ID range and an upper-bound comment
count, and can be recovered later. At the end of the comment phase the run prints
a summary of any skipped windows and the upper-bound number of missing comments.

A new `--backfill` mode (run like `--sync`, e.g.
`jbackup.pl --backfill --user=U --journal=J`) skips the entry sweep, rebuilds the
comment metadata, and re-pulls **only** the recorded skipped `comment_body`
windows — splitting each into sub-windows of the current `comment_body_fetch`
size, so a window that `504`d at 1000 can come through at, say, 250. Recovered
comments are saved and their markers cleared; any sub-window that still fails
re-records a *finer* marker, so successive backfill passes (at progressively
smaller batch sizes) narrow the gaps rather than restarting them.

### Metadata-window failure is now safe

A skipped `comment_meta` window is high-stakes — continuing past it would leave
`%meta` incomplete and `$server_max_id` wrong, corrupting and truncating the body
sweep. So:

- Metadata windows get more retry patience than body windows (`meta_max_tries`,
  default 9, vs `max_tries` default 6).
- If a metadata window still fails, the comment phase **stops cleanly before the
  body sweep**, with a clear message, rather than truncating. Entries and any
  previously-saved comments are intact, and a re-run rebuilds the in-memory,
  uncheckpointed metadata pass from scratch — so no `--backfill` is needed for
  metadata.
- When the metadata pass completes in full, any stale `comment_meta` marker from
  a prior aborted run is cleared automatically.

## Compatibility

- One new CLI flag, `--backfill`; all other behavior is on by default with
  conservative values, and every knob is optional.
- Database schema is unchanged. It reuses the existing `comment:lastid`
  checkpoint key (written more frequently) and adds one new bookkeeping key,
  `skipped:windows`, so existing `.jbak` files remain compatible.
- The only behavioral change to a *successful, unthrottled* run is the added
  per-call delays (seconds of overhead) and more frequent DB flushes.

## Testing

- `perl -c` clean under `use strict`.
- Skip accounting (record / parse / remove / report, including finer re-marking
  after a partial backfill and body-only window selection) validated via a
  serialization round-trip harness.
- Adaptive-pacing ramp-up/decay and the capped backoff schedule validated via
  simulation.
- Metadata-marker cleanup validated to clear only `comment_meta` markers and
  leave `comment_body` markers intact.
- The incremental-save path exercises the existing `comment:lastid` checkpoint
  that the script already read but never advanced incrementally.

## Notes for reviewers

These are three logically separable concerns and could be split into separate
PRs if preferred:

1. **Network resilience** — timeouts, backoff, rate-limiting, adaptive pacing,
   tunable batch size. Keeps a long run from escalating a throttle into a block.
2. **Incremental comment saving** — makes the comment phase resumable. The
   highest-impact change for large journals.
3. **Skip accounting, `--backfill`, and metadata-skip safety** — makes data loss
   visible, bounded, and recoverable instead of silent.

### DIFF
--- a/src/jbackup/jbackup.pl
+++ b/src/jbackup/jbackup.pl
@@ -116,7 +116,12 @@ $opts{verbose} = $opts{quiet} ? 0 : 1;
 
 # set some constants that should never need to change.
 my $COMMENTS_FETCH_META = 10000;   # up to 10000 comments, the maximum for comment_meta
-my $COMMENTS_FETCH_BODY = 1000;    # up to 1000 comments, the maximum for comment_body
+# comment_body batch size: the server max is 1000, but heavy windows (big threads) can blow
+# past DW's gateway timeout and 504. Lower this (e.g. comment_body_fetch=250 in ~/.jbackup)
+# to make each request lighter and far less likely to time out, at the cost of more requests.
+my $COMMENTS_FETCH_BODY = defined $opts{comment_body_fetch} ? $opts{comment_body_fetch}+0 : 1000;
+$COMMENTS_FETCH_BODY = 1000 if $COMMENTS_FETCH_BODY > 1000;   # server ceiling
+$COMMENTS_FETCH_BODY = 1    if $COMMENTS_FETCH_BODY < 1;
 
 # --- resilience knobs (override in ~/.jbackup, e.g. a line "xmlrpc_delay=3") ---
 my $XMLRPC_DELAY      = defined $opts{xmlrpc_delay} ? $opts{xmlrpc_delay}+0 : 2;   # sec before each XML-RPC call
--- a/src/jbackup/jbackup.pl
+++ b/src/jbackup/jbackup.pl
@@ -141,7 +141,7 @@ my $ADAPT_DELAY       = 0;   # current carried-over delay (state)
 my $ADAPT_OK          = 0;   # consecutive clean windows (state)
 
 # now figure out what we're doing
-if ($opts{help} || !($opts{sync} || $opts{dumptype} || $opts{alter_security})) {
+if ($opts{help} || !($opts{sync} || $opts{dumptype} || $opts{alter_security} || $opts{backfill})) {
     print <<HELP;
 jbackup.pl -- journal database generator and formatter
 
--- a/src/jbackup/jbackup.pl
+++ b/src/jbackup/jbackup.pl
@@ -124,6 +124,14 @@ my $FETCH_DELAY       = defined $opts{fetch_delay}  ? $opts{fetch_delay}+0  : 1;
 my $HTTP_TIMEOUT      = defined $opts{http_timeout} ? $opts{http_timeout}+0 : 120; # sec per request before it's an error (not a hang)
 my $MAX_BACKOFF_TRIES = defined $opts{max_tries}    ? $opts{max_tries}+0    : 6;   # retries w/ exponential backoff before giving up/skipping
 
+# adaptive pacing: extra delay carried across comment windows, ramps up on errors, decays
+# only after a clean streak -- so a struggling server gets sustained relief, not fresh poking.
+my $ADAPT_STEP        = defined $opts{adapt_step}  ? $opts{adapt_step}+0  : 5;   # first bump (sec) on the first error
+my $ADAPT_MAX         = defined $opts{adapt_max}   ? $opts{adapt_max}+0   : 60;  # ceiling for the carried-over delay
+my $ADAPT_DECAY_AFTER = defined $opts{adapt_decay} ? $opts{adapt_decay}+0 : 10;  # clean windows needed before easing off
+my $ADAPT_DELAY       = 0;   # current carried-over delay (state)
+my $ADAPT_OK          = 0;   # consecutive clean windows (state)
+
 # now figure out what we're doing
 if ($opts{help} || !($opts{sync} || $opts{dumptype} || $opts{alter_security})) {
     print <<HELP;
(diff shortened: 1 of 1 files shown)


afrozenpeach:
I've successfully backed up two full large communities with this updated version.

PR #3627, opened 2026-07-01
Add adversarial XSS corpus test for the HTML cleaner (initial)

zorkian:
An initial improvement to our HTML-cleaner test coverage. Adds `t/cleaner-xss.t`, a corpus of known XSS vectors — classic script/event-handler injection, HTML5 handlers, `javascript:`/`vbscript:` scheme obfuscation (entity, whitespace, case), `data:text/html`, CSS `expression()`/`-moz-binding`/`@import`, and mutation-XSS breakouts (noscript/xmp/style/template) — run through the real `clean_event` / `clean_comment` / `clean_userbio` display paths. Rather than matching exact output, it checks whether any JS-execution vector *survives* the clean, so it stays meaningful as the cleaner's internals change.

It doubles as a baseline: the current cleaner neutralizes the entire standard XSS canon (177 assertions pass). The one gap it records is that `clean_event`'s permissive allow-most mode lets SVG/MathML foreign content and its SMIL/XML-Events children (`<set attributeName=onload>`, `<handler>`, `<animate attributeName=href>`) through with only attribute-level filtering, while `clean_comment`'s strict allowlist strips them. Those are marked `TODO` so the suite stays green while documenting exactly what's missed. Browser-dependent today, but the class a namespace-aware sanitizer would close.

This is deliberately a first slice — more is coming: adversarial coverage for other surfaces (subjects, syndicated/feed content, anonymous comments) and formatting modes (markdown, preformatted), non-XSS viewer harms (CSS `url()` egress, image-proxy policy), and a stronger real-browser test oracle.

CODE TOUR: Dreamwidth lets people write HTML in their entries and comments, and we clean it so that viewing someone's post can't run malicious code in your browser. This change doesn't alter that cleaning at all — it adds a test that throws a big pile of known attack tricks at the cleaner and confirms they get defused, so future changes can't quietly reopen a hole. It also writes down, in the test itself, the few areas we still want to tighten, as a to-do list for follow-up work.

### DIFF
--- a/t/cleaner-resource-loading.t
+++ b/t/cleaner-resource-loading.t
@@ -1,0 +1,98 @@
+# t/cleaner-resource-loading.t
+#
+# Characterization tests for how LJ::CleanHTML handles third-party RESOURCE
+# LOADING in user content (external images and CSS url()). This is a privacy
+# concern, not code execution: fetching an external resource leaks the viewer's
+# IP / User-Agent (and a "this was viewed" signal) to that host.
+#
+# Dreamwidth INTENTIONALLY allows external images -- hotlinking is a long-standing
+# feature, and the DW::Proxy/https_url machinery exists only to fix mixed content
+# (http -> https), not to hide the viewer. Inline CSS url() is treated the same
+# way. These tests pin that accepted behavior so any change to it is deliberate,
+# and -- more importantly -- lock in the controls that stop the *dangerous*
+# variants: <style> blocks are removed (so there is no place to write the
+# attribute selectors that CSS-based data exfiltration needs), and script-y CSS
+# (url(javascript:), expression(), etc.) is stripped.
+#
+# If DW's privacy policy changes (e.g. block or proxy all third-party fetches),
+# update the "accepted policy" expectations below on purpose.
+#
+# Copyright (c) 2026 by Dreamwidth Studios, LLC.
+#
+# This program is free software; you may redistribute it and/or modify it under
+# the same terms as Perl itself. For a copy of the license, please reference
+# 'perldoc perlartistic' or 'perldoc perlgpl'.
+
+use strict;
+use warnings;
+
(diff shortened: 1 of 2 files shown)


zorkian:
Thanks — addressed both review notes in 9249126c9:

1. **Oracle entity decoding.** Replaced the hand-rolled partial decode with a single-pass, browser-accurate `decode_entities_once` before scheme detection. It now decodes named whitespace/colon entities (`&Tab;`, `&NewLine;`, `&colon;`) and numeric entities (hex — any case incl. A–F and uppercase `X` — and decimal), so an entity-obfuscated `javascript:` URL can't slip past the oracle if the cleaner ever regresses. Because it's single-pass, an already-escaped `&amp;Tab;` (which is inert in a browser) stays literal and does **not** false-positive. Added oracle self-check assertions that lock this in (it must see `&Tab;`/hex-entity/raw-handler vectors, and must ignore inert escaped input + attribute-value text).

   Minor note on the review: the hex decoder already matched A–F/uppercase-X via the `/i` flag, but the named-entity point was valid and is the real fix.

2. **Double clean.** Each payload is now cleaned once and the result reused for both the assertion and the failure `diag`.

Suite is green (192 tests, 0 non-TODO failures) and tidy passes.

zorkian:
Addressed the latest round (0f2fe20b3):

- **Header wording (cleaner-xss.t):** corrected — syndicated is an entries-only mode and `clean_comment` has no syndicated content; the header now lists entries' modes, comments' modes, subjects, bios, and the embed path accurately.
- **Brittle url() regexes (cleaner-resource-loading.t):** loosened both to `url(\s*["']?https://example.com/bg.png` (case-insensitive), so harmless CSS serialization changes (quotes/whitespace/casing) won't break them while still verifying the third-party URL is preserved.

Both files green (769 tests), tidy passes.

PR #3633, opened 2026-07-03
Preserve empty form fields in DW::Request arg parsing

### LINKED ISSUE #3628: Customize wizard won't let you unset values
If a value is set for one of the variables, clearing the field and then saving just results in the value you had cleared coming back. If it was initially unset, using 'reset' will clear it, but nothing else will. Notably, this seems to only affect the parts of the wizard that use the S2PropGroups widget - CustomText and LinkList will save clears properly. This is either limited to the Plack route or a recent addition - I can reproduce on prod, but was not able to reproduce on commit 06cce4e839f856d25d38060057b4b875fe3f91d9 via the Apache route. 

Best guess is that something is pruning the empty form fields before it reaches the widget logic, though it's odd that it's only doing it with one widget.

zorkian:
A submitted-but-empty form field (`foo=`) was coming back as `undef` instead of `""`. In `DW::Request::Base::_string_to_multivalue`, `LJ::parse_args` correctly yields `foo => ""`, but `split( /\0/, "" )` returns `()`, so the key produced no pairs and was dropped from the `Hash::MultiValue`. The fix keeps a single `""` when the split is empty.

This is a BML→Plack regression: the old `Apache::BML::parse_inputs` populated `%GET`/`%POST`/`%FORM` directly and kept empty fields, but BML pages now read form data through `$r->get_args`/`$r->post_args` → `_string_to_multivalue`. Added GET/POST cases to `t/request-multi.t` (they fail without the fix, pass with it).

Diagnosis by @momijizukamori.

CODE TOUR: When you submitted a form and left a field blank, the site was quietly throwing that blank away instead of remembering you cleared it — which broke the customize page (and any form that needs to know a field was emptied). This makes an empty field come through as an actual empty value again, the way it did before the move to the new web server. Comes with a test so it stays fixed.

Fixes #3628

### DIFF
--- a/cgi-bin/DW/Request/Base.pm
+++ b/cgi-bin/DW/Request/Base.pm
@@ -237,6 +237,10 @@ sub _string_to_multivalue {
     foreach my $key ( keys %gets ) {
 
         my @parts = defined $gets{$key} ? split( /\0/, $gets{$key} ) : '';
+
+        # split() on '' returns (), dropping the key; keep an empty field as ''
+        @parts = ('') unless @parts;
+
         push @out, map { $opts{lowercase} ? lc $key : $key => $_ } @parts;
     }
 
--- a/cgi-bin/DW/Request/Base.pm
+++ b/cgi-bin/DW/Request/Base.pm
@@ -236,9 +236,9 @@ sub _string_to_multivalue {
     my @out;
     foreach my $key ( keys %gets ) {
 
-        my @parts = defined $gets{$key} ? split( /\0/, $gets{$key} ) : '';
-
-        # split() on '' returns (), dropping the key; keep an empty field as ''
+        # -1 keeps trailing empties (repeated params); split('') returns (), so
+        # guard the all-empty case too. Both would otherwise drop empty fields.
+        my @parts = defined $gets{$key} ? split( /\0/, $gets{$key}, -1 ) : ('');
         @parts = ('') unless @parts;
 
         push @out, map { $opts{lowercase} ? lc $key : $key => $_ } @parts;
--- a/t/request-multi.t
+++ b/t/request-multi.t
@@ -15,7 +15,7 @@
 use strict;
 use warnings;
 
-use Test::More tests => 2;
+use Test::More tests => 4;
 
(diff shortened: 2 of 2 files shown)


momijizukamori:
This fixes #3628

zorkian:
**Risk sweep on the trailing-empty change (`split ... -1`)**

The one behavioral hazard of keeping trailing empties: `Hash::MultiValue` scalar access returns the *last* value, so for `foo=x&foo=` a scalar read of `$args->{foo}` flips from `'x'` to `''`. We swept the codebase for anything that could hit this — it requires (a) the same-named field submitted twice in one form with the last one empty, and (b) a consumer reading the scalar rather than `get_all`. Results:

| Consumer class | Finding | Verdict |
|---|---|---|
| TT views — repeated text/hidden names | 3 candidates (`newcat`, `keyword`, `query`) — all duplicates across *separate* `<form>`s, never co-submitted | safe |
| All repeated-name params (every `get_all` caller: `tags`, `others`, `rel_options`, `delete`, `crosspost`, `custom_bit`, `include`, `feed`, `column_N`, `oldkeywords`…) | All checkbox groups or multi-selects — unchecked/unselected fields aren't submitted at all, and checked values are non-empty constants; a trailing empty can't occur | safe |
| Widget layer (the one scalar-reader, `LJ::Widget` line 203 `$post->{$key}`) | Needs the same name twice in one widget form with the last empty; `html_check` emits no paired hidden, and no hidden-after-checkbox same-name pattern exists in `views/` or `cgi-bin/LJ/Widget/` | safe |
| BML pages | Read the `\0`-joined `%POST`; with `-1` that's byte-identical to the old `Apache::BML` (`join("\0", $apache_r->body($arg))` kept empties) | parity restored |

So: zero callers found that hit the scalar-flip, and the `-1` is what restores exact Apache-era `%POST` for BML pages — without it, `foo=x&foo=` would still come through differently than it did pre-Plack (and `foo=&foo=x` vs `foo=x&foo=` would behave inconsistently, since `split` only drops *trailing* empties).

PR #3634, opened 2026-07-03
Redesign /manage/icons for density and a mobile-friendly layout

zorkian:
The upload form and icon list were two side-by-side columns (`large-4`/`large-8`) that **stacked below 1024px**, pushing the icon list far down under a tall full-width upload box. This reworks the page into a full-width upload row above a **responsive icon grid** (two-up ≥640px, one-up on mobile) and tightens the upload form.

**Upload form**
- Source is now an **"Upload file"** button + an **"or enter URL"** field instead of two radios with separate labels/inputs. Each upload row carries a hidden `src_N` field the JS sets from the chosen source; `DW::Controller::EditIcons` now gates *every* row by its own `src_N` (previously only the first row was gated, by a single `src`), so an unused file/URL input never trips an empty-input error.
- One **"Add another"** button beside the primary action (now **"Upload"**) replaces the separate add-from-file / add-from-URL buttons; `editicons.js` builds a combined-source extra-upload row.
- Fields sit on one aligned grid (right-aligned labels · input · help to the right); **"About Icons"** becomes a **"?"** toggle by the heading; the "acceptable formats" line is dropped (upload already errors on bad types).
- Removed strings retired via `deadphrases.dat`.

Verified end-to-end on the dev server: file upload, URL upload, and multi-row upload all succeed with no empty-input errors, and "Add another" builds working rows in a real browser with no JS errors. `t/02-tidy.t` and `t/00-compile.t` pass.

### Before / after

| | Before | After |
|---|---|---|
| **Desktop (1200px)** | before desktop | after desktop |
| **Tablet (800px)** | before tablet | after tablet |
| **Mobile (430px)** | — | after mobile |

CODE TOUR: The "Manage Icons" page got a cleanup. The box for adding a new icon is now a single tidy row across the top — pick a file or paste a URL, no more duplicated radio buttons — and your existing icons show in two columns on a normal screen (one column on a phone) instead of one long list you had to scroll past a giant upload box to reach. Same features, less scrolling, and it finally works well on mobile.

### DIFF
--- a/bin/upgrading/deadphrases.dat
+++ b/bin/upgrading/deadphrases.dat
@@ -192,6 +192,13 @@ general /create.bml.tos.p1.2
 general /create.bml.username.charsallowed
 general /create.bml.username.text
 
+general /edit/icons.tt.fromfile
+general /edit/icons.tt.fromfile.key
+general /edit/icons.tt.fromurl
+general /edit/icons.tt.fromurl.key
+general /edit/icons.tt.upload.btn.addfile
+general /edit/icons.tt.upload.btn.addurl
+general /edit/icons.tt.upload.formats.desc
 general /editinfo.bml.allowshocontact.email.if_domainaddy
 general /editinfo.bml.allowshowcontact.email.reason
 general /editinfo.bml.allowshowcontact.email.withdomainaddr
--- a/cgi-bin/DW/Controller/EditIcons.pm
+++ b/cgi-bin/DW/Controller/EditIcons.pm
@@ -575,10 +575,11 @@ sub parse_post_uploads {
         }
 
         # uploaded pics
-        if ( $userpic_key =~ /userpic_.*/ ) {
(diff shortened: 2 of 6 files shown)


zorkian:
Thanks — addressed all three in the latest commit:

1. **`Source:` label** — it heads a file-or-URL group rather than a single control, so it's now a non-`label` element (styled identically) instead of a `<label>` with no association.
2. **Hardcoded `src_0`** — now defaults from the re-rendered `formdata` (`url` when `urlpic_0` is populated, else `file`), so a re-submit after a validation error uses the URL field. Verified: submitting a bad URL re-renders with `src_0='url'`.
3. **`bindUploadRow` init** — it now syncs `src_N` from any value already present (form re-render or browser autofill), not just on later focus/input/change.

Verified on the dev server (normal load → `file`, error re-render → `url`) and JS syntax-checked.

PR #3643, opened 2026-07-05
Skip captchas for logged-out browsers that recently held a good session

zorkian:
Adds a signed `ljtrust` cookie issued with the session cookies at login (`LJ::Session::update_trust_cookie`) and deliberately left behind on logout. On logged-out requests, `LJ::Session->trusted_anon_user` validates it (HMAC via the rotating `LJ::get_secret` pool, bound to the browser's `ljuniq` ident, 60-day window) and re-checks the account's standing live. A valid holder then gets the logged-in treatment from both captcha gates: the interstitial gate in `DW::Captcha::should_captcha_view` (checked lazily so trusted browsers never feed the fraud tempban counter) and the anon-only comment checks in `LJ::Talk::Post::require_captcha_test` (journal setting R, `comment_html_anon`); checks that also apply to logged-in users still fire. Skips are counted as `dw.captcha.bypassed` with tags mirroring `dw.captcha.shown`.

Also fixes a #3594 regression (first commit): class-method `DW::Captcha->site_enabled` always returned false, which had silently disabled all comment captchas and hidden the per-journal comment captcha setting.

CODE TOUR: If you browse or comment while logged out, but you'd logged in on that browser recently, Dreamwidth now remembers (via a tamper-proof cookie) that you're an established user in good standing and stops making you solve CAPTCHAs that exist to keep out anonymous bots. Journal owners' stricter settings, rate limits, and bans still apply exactly as they would to a logged-in user, and the free pass vanishes if the account is suspended.

### DIFF
--- a/cgi-bin/LJ/Session.pm
+++ b/cgi-bin/LJ/Session.pm
@@ -19,10 +19,11 @@ use LJ::Utils;
 
 use constant VERSION => 1;
 
-# how long a trust cookie vouches for a browser after it was last signed
-# (matches the "long" session length, so the bypass never outlives the
-# session that could have produced it)
-use constant TRUST_COOKIE_MAX_AGE => 60 * 86400;
+# how long a trust cookie vouches for a browser after it was last signed:
+# the "long" session length, so the bypass never outlives the session that
+# could have produced it (a sub, not a constant, since session_length isn't
+# compiled yet at this point)
+sub TRUST_COOKIE_MAX_AGE { return LJ::Session->session_length('long') }
 
 # NOTES
 #
--- a/cgi-bin/LJ/Session.pm
+++ b/cgi-bin/LJ/Session.pm
@@ -19,6 +19,11 @@ use LJ::Utils;
 
 use constant VERSION => 1;
 
+# how long a trust cookie vouches for a browser after it was last signed
+# (matches the "long" session length, so the bypass never outlives the
+# session that could have produced it)
+use constant TRUST_COOKIE_MAX_AGE => 60 * 86400;
+
 # NOTES
 #
 # * fields in this object:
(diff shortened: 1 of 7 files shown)


cgi-bin/LJ/Session.pm
```diff
@@ -19,6 +19,11 @@ use LJ::Utils;
 
 use constant VERSION => 1;
 
+# how long a trust cookie vouches for a browser after it was last signed
+# (matches the "long" session length, so the bypass never outlives the
+# session that could have produced it)
+use constant TRUST_COOKIE_MAX_AGE => 60 * 86400;
```
zorkian: If it matches another constant, should we just use that constant instead? I worry about these getting out of sync.

zorkian: Done in 25fbfcd4c — it now delegates to `LJ::Session->session_length('long')`. Kept as a sub rather than `use constant` because `session_length` isn't compiled yet at that point in the file.

cgi-bin/LJ/Talk.pm
```diff
@@ -3010,33 +3022,56 @@ sub require_captcha_test {
...
+            $bypassed //= 'comment_html';
+        }
+        else {
+            return 'comment_html';
+        }
+    }
 
-        if ( $body =~ /<[a-z]/i ) {
+    DW::Captcha->record_bypass( 'kind:form', 'page:comment', "reason:$bypassed" )
+        if $bypassed;
```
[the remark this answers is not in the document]
zorkian: Done in 25fbfcd4c — `record_bypass` now tags `type:<impl>` (the implementation that would have been used), matching `dw.captcha.shown`.

PR #3654, opened 2026-07-11
Add on-site multi-account switcher

### LINKED ISSUE #3302: On-site account/session switcher
Copying out notes from discord for capture:

> also, bringing an idea I had back around now that people are online, and with the caveat that this is 100% an 'after conversion are done thing': site-based account switching
> NOT account-linking, to be clear
> but just a multi-session mechanism with some UI for it
> which I thought of because Github just added it
> and it looks like there it's implemented with an additional `saved_user_sessions` cookie, which looks to be of the format `<user id>:<session cookie>|<second id>:<second cookie>`, url-encoded.
> similar behavoir to the google account switcher, if anyone else here also has like five different google accounts.
> Would probably want to put the switcher UI on the site masthead, the entry post page, comment post pages, and the navstrip
> and work out a workflow for expired sessions (google's version lists the account but marks it as being expired and prompts you for the pw when you switch to it)

zorkian:
Lets a browser hold several accounts signed in at once and switch between them without re-entering a password (GitHub/Google-style). The active account stays in `ljmastersession`/`ljloggedin` and is validated exactly as before; a new http_only `ljsessions` cookie carries the other signed-in accounts as `(userid, sessid, auth)` handles, and every switch re-validates against the DB the same way `session_from_master_cookie` does. `DW::AccountSwitcher` owns the cookie and the add/switch/promote/remove logic; wiring lives in `DW::Controller::{Login,Auth,Settings}`, and the caret/fold-out UI is in the Foundation masthead (`schemes/common.tt`), the journal control strip (`views/journal/controlstrip.tt`), and the legacy sitescheme, with an "Account Switcher" management page at `/manage/accounts` linked from the Organize menu.

## Screenshots

A bare caret next to your username opens a switch-only fold-out.

**Masthead (Foundation)**

masthead

**Journal control strip**

control strip

**Legacy (non-Foundation) sitescheme**

legacy

**Manage page (`/manage/accounts`, under the Organize menu)** — one alphabetical list, current account flagged inline, switch/remove per row:

manage

**Adding another account**

add account

CODE TOUR: You can now be logged in to more than one Dreamwidth account in the same browser and flip between them without typing your password each time. Next to your username there's a small arrow that drops down your other accounts; click one to switch. There's also an "Account Switcher" page (under the Organize menu) where you can add, switch, or remove accounts, and changing your password no longer boots you out of the others.

Fixes #3302

### DIFF
--- a/bin/upgrading/en.dat
+++ b/bin/upgrading/en.dat
@@ -2097,6 +2097,8 @@ menunav.organize.customizejournalstyle=Customize Journal Style
 
 menunav.organize.importcontent=Import Content
 
+menunav.organize.manageaccounts=Account Switcher
+
 menunav.organize.managecommunities=Manage Communities
 
 menunav.organize.managefilters=Manage Filters
@@ -4000,6 +4002,12 @@ sitescheme.accountlinks.post=Post
 
 sitescheme.accountlinks.readinglist=Reading Page
 
+sitescheme.accountlinks.switch.go=Switch
+
+sitescheme.accountlinks.switch.label=Switch account
+
+sitescheme.accountlinks.switch.signedout=Signed out — log in
+
 sitescheme.accountlinks.userpic.alt=Upload Icons
 
 sitescheme.footer.info=Copyright &copy; 2009-2011 Dreamwidth Studios, LLC. All rights reserved.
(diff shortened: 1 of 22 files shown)


zorkian:
Both Copilot points addressed in e5d7c26f9:

- **login.tt username field** — switched `value =>` to `default =>` so a POSTed value still wins after a failed login, while the GET `?user=` prefill (used for the expired-session re-login path) keeps working. Verified: add-account GET renders `value="test_friend"`, and `default` is the documented param for exactly this ("does not override the value of a previous form submission").
- **switch_handler returnto** — now strips CR/LF before it can reach a Location header, matching the login controller.

